PR #3, opened 2017-02-15
Add i18n json schema

eyurko:


eyurko:
Whoops -- let's try that again. I think I signed up via the appropriate channels, this time.

schema/i18n.json
```diff
@@ -0,0 +1,112 @@
+"interview_question": {
+  "title": "Internationalization (i18n)",
+  "prompts": [
+    {
+      "instructions": "Welcome to this programming exercise.\n\nYou will complete this exercise by editing the code in the sample.py file\nusing your favorite editor.\nEvery time you save the file, I will evaluate your code and let you know about\nnext steps.\n\nYour first task is to implement the abbreviate function. It takes a string as input\nand return an abbreviation of the string of the form <first character><length of the\nmiddle of the string><last character>. For example, \"internationalization\" should be\nabbreviated as \"i18n\".",
```
seanlip: Do you need the intermediate newlines, e.g. between 'file' and 'using'? I thought we could just trust the UI to handle wrapping, and you'd only need paragraph breaks.

Speaking of paragraph breaks, if you want those too, would it be better to have a list of strings?

schema/i18n.json
```diff
@@ -0,0 +1,112 @@
...
+      "prerequisite_skills": ["Arrays", "Strings", "String Manipulation"],
+      "acquired_skills": ["String Manipulation"],
+      "input_function": null,
+      "output_function": null,
+      "main_function":, sample.abbreviate,
```
seanlip: Might need to make the values strings, otherwise the JSON is invalid.

schema/i18n.json
```diff
@@ -0,0 +1,112 @@
...
+      ],
+      "buggy_outputs": [],
+      "performance_test_data": {
+        "input_data_atom": "o",
+        "transformation_function": createListOfUniqueStrings,
+        "expected_performance": "linear",
+        "evaluation_function": sample.all_are_unique,
+      },
+    },
+  ]
```
seanlip: Is there a missing comma here?

You might want to plug this into an online json linter/validator.

schema/i18n.json
eyurko: I was just copying the format from the previous instance, but I can make these lists of strings if you'd like.

eyurko:
I'm going to submit a PR from a different branch -- want to put this in my experimental directory so I can use cider.

PR #4, opened 2017-02-15
Adding the RFC 4627-compliant version of the i18n json schema.

eyurko:
I updated it according to comments on the previous branch.

### DIFF
--- a/assets/questions/i18n.json
+++ b/assets/questions/i18n.json
@@ -1,37 +1,124 @@
 {
-    "language": "python",
-    "stages": [{
-        "instructions": "Welcome to this programming exercise. Implement the abbreviate function.",
-        "data": [{
+  "interview_question": {
+    "title": "Internationalization (i18n)",
+    "prompts": [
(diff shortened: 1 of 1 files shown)


assets/questions/i18n.json
```diff
@@ -1,37 +1,124 @@
...
+              "Make sure to add the last character of the string back on when you've abbreviated."
+            ]
+          }
+        ],
+        "performance_test_data": {
+        }
+      },
+      {
+        "instructions": [
+          "You need to make sure that your code handles short strings correctly.",
```
apicardgoog: Fine for now but I think we may want to rephrase this in the future to something like:
"It's always important to think about the edge cases for your code. In this case, short strings are not abbreviated by this strategy. Make sure your code doesn't try to abbreviate when it's not necessary."

seanlip: @eyurko Might as well just fix this now, if it's easy? Feel free to merge after that.

PR #5, opened 2017-02-16
Adding the RFC 4627-compliant version of the run-length encoding json schema.

eyurko:
… schema.

### DIFF
--- a/assets/questions/rle.json
+++ b/assets/questions/rle.json
@@ -5,7 +5,6 @@
       {
         "instructions": [
           "Welcome to this programming exercise.",
-          "You will complete this exercise by editing the code in the sample.py file using your favorite editor. Every time you save the file, I will evaluate your code and let you know about next steps.",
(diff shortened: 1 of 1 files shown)


assets/questions/rle.json
```diff
@@ -0,0 +1,135 @@
+{
+  "interview_question": {
+    "title": "Run-Length Encoding",
+    "prompts": [
+      {
+        "instructions": [
+          "Welcome to this programming exercise.",
+          "You will complete this exercise by editing the code in the sample.py file using your favorite editor. Every time you save the file, I will evaluate your code and let you know about next steps.",
```
seanlip: It strikes me that the first sentence of this text no longer applies...

eyurko: I'm about to head out, but what would you recommend?

seanlip: Actually, on re-reading, the second sentence doesn't apply either.

Maybe drop this sentence altogether?

PR #8, opened 2017-02-17
Update schema with starter and extra code

eyurko:


assets/questions/i18n.json
```diff
@@ -1,6 +1,8 @@
 {
   "interview_question": {
     "title": "Internationalization (i18n)",
+    "starter_code": "def abbreviate(word):\n\t\treturn \"\"\n\ndef are_all_unique(words):\n\t\treturn True",
+    "extra_code": "def forgetLastLetter(word):\n\tresult = \"%s%d\" % (word[0], len(word) - 2) if len(word) > 2 else word\n\treturn result\n\ndef useFirstAndLastLetterAndLengthToAbbreviate(word):\n\tif word:\n\t\treturn \"%s%d%s\" % (word[0], len(word) - 2, word[len(word) - 1])\n\treturn \"\"\n\ndef CreateListOfUniqueStrings(atom, size):\n\tresult = []\n\tfor i in range(size):\n\t\tresult.append(atom * (i + 1))\n\treturn result"
     "prompts": [
```
seanlip: Did you validate? JSON linter does not look happy.

assets/questions/i18n.json
```diff
@@ -1,6 +1,8 @@
 {
   "interview_question": {
     "title": "Internationalization (i18n)",
+    "starter_code": "def abbreviate(word):\n\t\treturn \"\"\n\ndef are_all_unique(words):\n\t\treturn True",
```
seanlip: Why two tabs? Ditto for the other schema.

assets/questions/i18n.json
```diff
@@ -1,6 +1,8 @@
...
+    "extra_code": "def forgetLastLetter(word):\n\tresult = \"%s%d\" % (word[0], len(word) - 2) if len(word) > 2 else word\n\treturn result\n\ndef useFirstAndLastLetterAndLengthToAbbreviate(word):\n\tif word:\n\t\treturn \"%s%d%s\" % (word[0], len(word) - 2, word[len(word) - 1])\n\treturn \"\"\n\ndef CreateListOfUniqueStrings(atom, size):\n\tresult = []\n\tfor i in range(size):\n\t\tresult.append(atom * (i + 1))\n\treturn result"
```
seanlip: I lean towards auxiliary_code, since that has a connotation of "supporting". "Extra" has a connotation of ... well, extra :) I.e., sort of optional.

Your call, though. Just a thought.

eyurko:
I wasn't really sure. I'd suggest that we do auxilliary_code / starter_code as a dictionary, mapping from language to the code for that language?

assets/questions/i18n.json
```diff
@@ -1,6 +1,8 @@
 {
   "interview_question": {
     "title": "Internationalization (i18n)",
+    "starter_code": "def abbreviate(word):\n\treturn \"\"\n\ndef are_all_unique(words):\n\treturn True",
+    "auxilliary_code": "def forgetLastLetter(word):\n\tresult = \"%s%d\" % (word[0], len(word) - 2) if len(word) > 2 else word\n\treturn result\n\ndef useFirstAndLastLetterAndLengthToAbbreviate(word):\n\tif word:\n\t\treturn \"%s%d%s\" % (word[0], len(word) - 2, word[len(word) - 1])\n\treturn \"\"\n\ndef CreateListOfUniqueStrings(atom, size):\n\tresult = []\n\tfor i in range(size):\n\t\tresult.append(atom * (i + 1))\n\treturn result",
```
seanlip: NB: "auxiliary" has one "l".

seanlip:
Yep, that sounds fine to me (dict with language --> code).

assets/questions/i18n.json
eyurko: You'd never believe that I was a county spelling bee champion in my youth.

assets/questions/rle.json
```diff
@@ -92,7 +94,7 @@
         "acquired_skills": ["String Manipulation", "Sets", "Arrays", "Maps"],
         "input_function": null,
         "output_function": null,
-        "main_function": "sample.DecodeEncodedString",
+        "main_function": "DecodeEncodedString",
```
seanlip: Need to change this in the other schema too, right?

assets/questions/i18n.json
```diff
@@ -1,6 +1,10 @@
 {
   "interview_question": {
     "title": "Internationalization (i18n)",
+    "starter_code": "def abbreviate(word):\n\treturn \"\"\n\ndef are_all_unique(words):\n\treturn True",
+    "auxiliary_code": {
```
seanlip: Don't forget to add the dict to starter_code, here and in both places in the other file.

eyurko: right right right -- on it.

assets/questions/rle.json
eyurko: Which other schema? This DecodeEncodedString is something that _I'm_ writing, not something the student should be writing.

seanlip: Oh, sorry, now I'm confused. I thought this was the name of the main function in the student's code that is to be executed? Similar to abbreviate() in i18n.

Also, by my previous comment, I meant removing 'sample.' in i18n.json.

eyurko: That's my bad for being confusing :)

Here, the student is expected to write encode. For this test, however, we're asking them to make sure that their encoded string can be properly decoded, and then we're running _our_ Decode method on the result of their encode, and making sure the input == decode(encode(input)). I realized that wouldn't be in sample, so I dropped it. It's provided in aux_code.

seanlip: Sounds like Decode would be an output_function, then? Seems similar to my question a while back on how you'd evaluate whether a set of numbers that are output by a student has a mean of X -- here, you're evaluating whether a student's encoded output can be decoded.

eyurko: I ... hadn't considered that, but it makes perfect sense.

seanlip: OK :)

Also, we should be removing all the `sample`s, right?

eyurko: I wasn't actually sure, but yes, that makes sense -- happy to do that.

eyurko:
Thanks Sean! Have a good weekend. :)

seanlip:
You too! :)

PR #9, opened 2017-02-18
Flesh out the pipeline for submitting and evaluating code.

seanlip:
This PR fleshes out the code submission pipeline. (When reviewing, you might want to try checking out this branch and running it locally.)

/cc @apicardgoog

### DIFF
--- a/client/services/FeedbackGeneratorService.js
+++ b/client/services/FeedbackGeneratorService.js
@@ -14,18 +14,47 @@
 
 /**
  * @fileoverview Service for generating feedback based on the evaluation result
- * of the code submitted by the user.
+ * of the code submitted by the user, preprocessed and augmented with unit
+ * tests. This assumes that the base code already passes "compile-time" checks.
  */
 
 tie.factory('FeedbackGeneratorService', [
   function() {
+    // TODO(sll): Add escaping?
+    var jsToHumanReadable = function(jsVariable) {
+      if (typeof jsVariable === 'string') {
+        return '"' + jsVariable + '"';
+      }
+    };
+
     return {
-      getFeedback: function(question, codeEvalResult) {
+      getFeedback: function(prompt, codeEvalResult) {
         if (codeEvalResult.getErrorMessage()) {
           return (
             'Your code threw an error: ' + codeEvalResult.getErrorMessage());
         } else {
-          return 'Your code gave this output: ' + codeEvalResult.getOutput();
+          var correctnessTests = prompt.getCorrectnessTests();
+          var observedOutputs = codeEvalResult.getTestResults();
+          for (var i = 0; i < correctnessTests.length; i++) {
+            var expectedOutput = correctnessTests[i].getExpectedOutput();
+            var observedOutput = observedOutputs[i];
+
+            if (expectedOutput !== observedOutput) {
+              return [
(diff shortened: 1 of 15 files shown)


client/services/FeedbackGeneratorService.js
```diff
@@ -14,18 +14,47 @@
...
-          return 'Your code gave this output: ' + codeEvalResult.getOutput();
+          var correctnessTests = prompt.getCorrectnessTests();
+          var observedOutputs = codeEvalResult.getTestResults();
+          for (var i = 0; i < correctnessTests.length; i++) {
+            var expectedOutput = correctnessTests[i].getExpectedOutput();
+            var observedOutput = observedOutputs[i];
+
+            if (expectedOutput !== observedOutput) {
+              return [
+                'Your code gave the output ',
```
eyurko: Should we be using the varied messages from intertest/main.py?

client/services/SolutionBrokerService.js
```diff
@@ -18,20 +18,38 @@
  */
 
 tie.factory('SolutionBrokerService', [
```
eyurko: jw - why is it called SolutionBrokerService? Just trying to better understand the codebase.

client/services/code_evaluators/PythonCodeRunnerService.js
```diff
@@ -38,12 +39,17 @@ tie.factory('PythonCodeRunnerService', [
         return Sk.misceval.asyncToPromise(function() {
           return Sk.importMainWithBody('<stdin>', false, code, true);
         }).then(function() {
+          var resultList = [];
+          if (Sk.globals.hasOwnProperty(VARNAME_TEST_RESULTS)) {
+            resultList = Sk.ffi.remapToJs(Sk.globals[VARNAME_TEST_RESULTS]);
```
eyurko: What does this do? Might be worth commenting for posterity, especially for those of us that are less JavaScripty. :)

client/services/FeedbackGeneratorService.js
seanlip: Good point, but let's defer that to a separate incremental PR? I think that's a nice chunk of work on its own, and I don't want this one to get too big.

client/services/SolutionBrokerService.js
seanlip: I had some difficulty coming up with a name for this; let's discuss offline.

The main reason I called it that was because this service "handles" an incoming solution, and sends it through a chain of getting feedback from running the code, correctness checks, etc. -- so this is an overarching, "organizational" service. I'm not entirely sure what to name it, though; perhaps we could go with SolutionHandlerService or something. But we can discuss this offline.

client/services/code_evaluators/PythonCodeRunnerService.js
seanlip: Thanks, good call! It's not a JS thing, actually, it's a Skulpt thing. I've added a comment, PTAL and see if it makes sense.

client/services/SolutionBrokerService.js
seanlip: Changed to SolutionHandlerService per offline discussion.

seanlip:
Thanks, @eyurko! All comments addressed, PTAL.

PR #10, opened 2017-02-22
Add support for multiple prompts within a single question.

seanlip:
/cc @apicardgoog

### DIFF
--- a/client/components/LearnerViewDirective.js
+++ b/client/components/LearnerViewDirective.js
@@ -24,24 +24,41 @@ tie.directive('learnerView', [function() {
       <h3>Exercise: {{title}}</h3>
 
       <div class="tie-learner-view-left-column">
+        <div class="tie-old-instructions">
+          <div ng-repeat="oldInstruction in oldInstructions track by $index">
+            <p ng-repeat="paragraph in oldInstruction track by $index">
+              {{paragraph}}
+            </p>
+            <hr>
+          </div>
+        </div>
         <div class="tie-instructions">
-          <p ng-repeat="instruction in instructions">
-            {{instruction}}
+          <p ng-repeat="paragraph in instructions">
+            {{paragraph}}
           </p>
         </div>
+
+        <button ng-if="nextButtonIsShown" class="tie-action-button active"
+                ng-click="showNextPrompt()">
+          Next
+        </button>
       </div>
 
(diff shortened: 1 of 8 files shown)


client/components/LearnerViewDirective.js
```diff
@@ -24,24 +24,41 @@ tie.directive('learnerView', [function() {
       <h3>Exercise: {{title}}</h3>
 
       <div class="tie-learner-view-left-column">
+        <div class="tie-old-instructions">
```
eyurko: I'm not sure what the difference is between old-instructions and instructions. Could you clarify?

client/components/LearnerViewDirective.js
```diff
@@ -105,25 +133,49 @@ tie.directive('learnerView', [function() {
...
         var question = QuestionDataService.getData();
         var prompts = question.getPrompts();
+        var currentPromptIndex = 0;
 
-        var currentPrompt = prompts[0];
-        $scope.title = question.getTitle();
-        $scope.code = question.getStarterCode(language);
-        $scope.instructions = currentPrompt.getInstructions();
-        $scope.feedback = '';
+        $scope.questionIsCompleted = false;
```
eyurko: So here you're using questionIsCompleted, and elsewhere you use isAnswerCorrect() and isLastPrompt(). Is this just a stylistic difference to differentiate between boolean variables and methods that return booleans?

client/components/LearnerViewDirective.js
eyurko: Oh no wait I got it. Might be worth a comment, or calling them "previous-instructions"?

client/components/LearnerViewDirective.js
seanlip: Yep, that's the intention.

client/components/LearnerViewDirective.js
seanlip: Yep, I can change it to "previous-instructions". I'll do that shortly.

seanlip: Done.

client/components/LearnerViewDirective.js
eyurko: Cool, just wanted to make sure for my own sake.

PR #11, opened 2017-02-22
Add codemirror dependency, which provides line numbers, syntax highlighting and tabs.

seanlip:
Note that this is a small PR. Most of the added stuff is in third-party.

/cc @apicardgoog


client/components/LearnerViewDirective.js
```diff
@@ -151,6 +147,14 @@ tie.directive('learnerView', [function() {
           $scope.$apply();
         };
 
+        $scope.codeMirrorOptions = {
+          autofocus: true,
+          lineNumbers: true,
+          mode: LANGUAGE_PYTHON,
```
eyurko: Might be worth adding a TODO to eventually support more languages, but this is a nit, _at best_.

third_party/codemirror-5.19.0/mode/python/python.js
```diff
@@ -0,0 +1,340 @@
+// CodeMirror, copyright (c) by Marijn Haverbeke and others
```
eyurko: This is kind of neat to read through.

client/components/LearnerViewDirective.js
seanlip: I lean against, since that would be a major product decision which I doubt would slip through the cracks anyway. We'd also have to add a similar TODO in many more places, like the preprocessor and evaluator. So, leaving as-is.

PR #15, opened 2017-02-23
Update tests and fix node installation process.

seanlip:


seanlip:
The main file to review is:
- scripts/run_karma_tests.sh

The other changes are due to tests actually failing on master due to previous modification of the preprocessCode function (mea culpa). We'll need to put in some presubmit checks in order to guard against this in the future.

scripts/run_karma_tests.sh
```diff
@@ -25,18 +25,49 @@
...
+    else
+      NODE_FILE_NAME=node-v6.9.1-linux-x86
+    fi
+  fi
+
+  if [ ! -d "$TOOLS_DIR" ]; then
+    mkdir tools
+  fi
+
+  curl -o node-download.tgz http://nodejs.org/dist/v6.9.1/$NODE_FILE_NAME.tar.gz
```
rabidbit: Why not download over https?

seanlip: My bad (and good catch)! Done.

PR #17, opened 2017-02-23
Updated UI.

rabidbit:
Revised UI to make it prettier and scroll prompts and feedback when they are appended.


client/components/LearnerViewDirective.js
```diff
@@ -21,104 +21,172 @@ tie.directive('learnerView', [function() {
...
+              <div id="tie-feedback" class="tie-feedback">
+                <p ng-repeat="paragraph in feedbackMessages track by $index"
+                    class="tie-feedback-paragraph">
+                  {{paragraph}}
+                </p>
+              </div>
+            </div>
+            <div id="coding-window">
+              <div id="lang-terminal">
+                <!--textarea name="textarea" id="coding-terminal"
```
seanlip: Don't commit any commented-out code.

client/components/LearnerViewDirective.js
```diff
@@ -21,104 +21,172 @@ tie.directive('learnerView', [function() {
...
-             class="tie-feedback-paragraph">
-            {{paragraph}}
-          </p>
         </div>
       </div>
-
       <style>
-        .tie-learner-view-left-column {
+        body {
+          font-family: Roboto, 'Helvetica Neue', 'Lucida Grande', sans-serif;
```
seanlip: Could we, as a general rule, alphabetize CSS (so, background-color, then font-family, then font-size)? It makes it easier to find whether a key is specified, especially in rules that have many keys.

client/components/LearnerViewDirective.js
```diff
@@ -21,104 +21,172 @@ tie.directive('learnerView', [function() {
...
+          font-size: 15px;
+          background-color: rgb(242, 242, 242);
+        }
+        #exercise-ui-outer {
```
seanlip: As a general rule, I recommend using classes rather than IDs, similar to what's already been done in this file. At the moment we have a single HTML file displaying the whole page so it doesn't matter so much, but when we start pulling things out into separate files it may be harder to maintain the requirement of IDs being unique within the entire page. Classes are more easily combined with one another, too.

Also, please prefix tie-specific classes with `.tie-`  -- that helps us distinguish classes for stuff we've written ourselves from classes we need to modify from external libs.

client/components/LearnerViewDirective.js
```diff
@@ -132,6 +200,8 @@ tie.directive('learnerView', [function() {
         var question = QuestionDataService.getData();
         var prompts = question.getPrompts();
         var currentPromptIndex = 0;
+        var feedback_div = document.getElementById('tie-feedback');
```
seanlip: Our naming convention in JS is lowerCamelCase.

client/components/LearnerViewDirective.js
```diff
@@ -164,6 +235,12 @@ tie.directive('learnerView', [function() {
             $scope.previousInstructions.push($scope.instructions);
             $scope.instructions = prompts[currentPromptIndex].getInstructions();
             $scope.nextButtonIsShown = false;
+            // TODO (johnmunoz): DOM not yet updated; is there a better way
+            // than using a timeout?
```
seanlip: I suspect there is indeed a better way to do this, but I need to look into it. Just letting you know I haven't skipped over this!

third_party/codemirror-5.19.0/lib/codemirror.css
```diff
@@ -3,7 +3,8 @@
 .CodeMirror {
   /* Set height, width, borders, and global font properties here */
   font-family: monospace;
-  height: 300px;
+  height: 100%;
```
seanlip: Do not modify files in third-party, otherwise the versioning/pinning we're doing to ensure stability completely fails. We can add library CSS modifications in app.html for now, if needed.

rabidbit:
Addressed comments. PTAL

client/components/LearnerViewDirective.js

client/components/LearnerViewDirective.js

client/components/LearnerViewDirective.js

client/components/LearnerViewDirective.js

third_party/codemirror-5.19.0/lib/codemirror.css
rabidbit: Addressed.

client/components/LearnerViewDirective.js
rabidbit: Leaving as is, until we have a better solution.

client/components/LearnerViewDirective.js
```diff
@@ -21,104 +21,174 @@ tie.directive('learnerView', [function() {
...
-      <h3>Exercise: {{title}}</h3>
-
-      <div class="tie-learner-view-left-column">
-        <div class="tie-previous-instructions">
-          <div ng-repeat="previousInstruction in previousInstructions track by $index">
-            <p ng-repeat="paragraph in previousInstruction track by $index">
-              {{paragraph}}
-            </p>
-            <hr>
+      <div id="tie-exercise-ui-outer" class="tie-exercise-ui-outer">
```
seanlip: Er, I might not have been clear previously. Can we drop the IDs throughout, unless they're really needed?

client/app.html
```diff
@@ -3,50 +3,39 @@
...
     </script>
-  </head>
-
-  <body>
-    <learner-view></learner-view>
-
     <script src="../third_party/angular-1.6.1/angular.min.js"></script>
     <script src="../third_party/codemirror-5.19.0/lib/codemirror.js"></script>
-    <script src="../third_party/codemirror-5.19.0/mode/python/python.js"></script>v
+    <script src="../third_party/codemirror-5.19.0/mode/python/python.js"></script>
```
seanlip: Btw, thanks for catching this!

client/components/LearnerViewDirective.js
```diff
@@ -21,104 +21,174 @@ tie.directive('learnerView', [function() {
...
       </div>
-
       <style>
-        .tie-learner-view-left-column {
-          float: left;
-          width: 400px;
+        body {
+          font-family: Roboto, 'Helvetica Neue', 'Lucida Grande', sans-serif;
+          font-size: 15px;
+          background-color: rgb(242, 242, 242);
```
seanlip: Alphabetize.

client/components/LearnerViewDirective.js
```diff
@@ -21,104 +21,174 @@ tie.directive('learnerView', [function() {
...
+            </div>
+            <div id="tie-coding-window" class="tie-coding-window">
+              <div id="tie-lang-terminal" class="tie-lang-terminal">
+                <div id="tie-coding-terminal" class="tie-coding-terminal">
+                  <ui-codemirror ui-codemirror="codeMirrorOptions" ng-model="code" class="tie-codemirror-container"></ui-codemirror>
+                </div>
+                <select id="tie-lang-select-menu" class="tie-lang-select-menu"
+                    name="lang-select-menu">
+                  <option value="Python" selected>Python</option>
+                  <option value="JavaScript">JavaScript</option>
```
seanlip: Let's remove the dropdown; we aren't supporting multiple languages yet and it'll take a while to do so. (It'll also be problem-specific.)

rabidbit: How far out are we to adding new languages? I can imagine people asking if it will support more languages than Python, and as a prototype, it shows this planned functionality and how users can change the language.

client/components/LearnerViewDirective.js
rabidbit: Addressed

client/components/LearnerViewDirective.js
rabidbit: Addressed

client/components/LearnerViewDirective.js
seanlip: Quite far. From our work with just trying to support Python, it turns out that there's quite a lot of additional tooling and functionality needed in order to support one language, and that this will need to be replicated for other languages. So it will likely not happen in the near future.

Maybe we can have a drop down with a single item for now.

client/components/LearnerViewDirective.js
```diff
@@ -21,104 +21,173 @@ tie.directive('learnerView', [function() {
...
+              <div class="tie-lang-terminal">
+                <div class="tie-coding-terminal">
+                  <ui-codemirror ui-codemirror="codeMirrorOptions" ng-model="code" class="tie-codemirror-container"></ui-codemirror>
+                </div>
+                <select class="tie-lang-select-menu"
+                    name="lang-select-menu">
+                  <option value="Python" selected>Python</option>
+                </select>
+                <button class="tie-run-button"
+                    ng-class="tie-{'active': !nextButtonIsShown}"
```
seanlip: This looks odd. Shouldn't it be `ng-class="{'tie-active': !nextButtonIsShown}`?

rabidbit:
PTAL

client/components/LearnerViewDirective.js
```diff
@@ -132,6 +201,8 @@ tie.directive('learnerView', [function() {
         var question = QuestionDataService.getData();
         var prompts = question.getPrompts();
         var currentPromptIndex = 0;
+        var feedbackDiv = document.getElementById('tie-feedback');
+        var tieInstructionsDiv = document.getElementById('tie-instructions');
```
seanlip: instructionsDiv, to match previous line.

client/components/LearnerViewDirective.js
rabidbit: Addressed

client/components/LearnerViewDirective.js
rabidbit: Arg.. search and replace error. Fixed.

client/components/LearnerViewDirective.js
rabidbit: Addressed

PR #19, opened 2017-02-24
Add initial text to feedback frame; replace previous feedback when new feedback is obtained.

seanlip:
Note: this also switches the left and right columns, otherwise the initial feedback doesn't make sense.

/cc @eyurko @apicardgoog

### DIFF
--- a/client/components/LearnerViewDirective.js
+++ b/client/components/LearnerViewDirective.js
@@ -23,31 +23,10 @@ tie.directive('learnerView', [function() {
     template: `
       <div class="tie-exercise-ui-outer">
         <div class="tie-exercise-ui-inner">
-          <div class="tie-question-ui">
-            <div class="tie-question-window">
-              <h3>{{title}}</h3>
-              <div class="tie-previous-instructions">
-                <div ng-repeat="previousInstruction in previousInstructions track by $index">
-                  <p ng-repeat="paragraph in previousInstruction track by $index">
-                    {{paragraph}}
-                  </p>
-                  <hr>
-                </div>
-              </div>
-              <div id="tie-instructions" class="tie-instructions">
-                <p ng-repeat="paragraph in instructions">
-                  {{paragraph}}
-                </p>
-              </div>
-            </div>
-            <button ng-click="showNextPrompt()" ng-if="nextButtonIsShown"
-                class="tie-next-button">Next</button>
-            </button>
-          </div>
           <div class="tie-coding-ui">
             <div class="tie-feedback-window">
               <div id="tie-feedback" class="tie-feedback">
-                <p ng-repeat="paragraph in feedbackMessages track by $index"
+                <p ng-repeat="paragraph in feedbackParagraphs track by $index"
                     class="tie-feedback-paragraph">
                   {{paragraph}}
                 </p>
@@ -70,6 +49,27 @@ tie.directive('learnerView', [function() {
                 </button>
               </div>
             </div>
+          </div>
+          <div class="tie-question-ui">
+            <div class="tie-question-window">
+              <h3>Exercise {{exerciseNumber}}: {{title}}</h3>
+              <div class="tie-previous-instructions">
+                <div ng-repeat="previousInstruction in previousInstructions track by $index">
+                  <p ng-repeat="paragraph in previousInstruction track by $index">
+                    {{paragraph}}
+                  </p>
+                  <hr>
+                </div>
+              </div>
+              <div id="tie-instructions" class="tie-instructions">
+                <p ng-repeat="paragraph in instructions">
+                  {{paragraph}}
+                </p>
+              </div>
+            </div>
+            <button ng-click="showNextPrompt()" ng-if="nextButtonIsShown"
+                class="tie-next-button">Next</button>
+            </button>
           </div>
         </div>
       </div>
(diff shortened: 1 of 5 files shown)


client/components/LearnerViewDirective.js
```diff
@@ -71,6 +50,27 @@ tie.directive('learnerView', [function() {
               </div>
             </div>
           </div>
+          <div class="tie-question-ui">
+            <div class="tie-question-window">
+              <h3>Exercise {{exerciseNumber}}: {{title}}</h3>
```
rabidbit: I took out "Exercise :" to make the title more flexible.

seanlip: Ah, sorry, I was following the mocks you sent. What do you want the title to be?

client/components/LearnerViewDirective.js
```diff
@@ -249,10 +249,20 @@ tie.directive('learnerView', [function() {
...
           SolutionHandlerService
             .processSolutionAsync(prompts[currentPromptIndex], code, language)
-            .then(appendFeedback);
+            .then(setFeedback);
         };
 
-        clearFeedback();
+        $scope.feedbackParagraphs = [
+          'Greetings!',
+          'This set of exercises focuses on string manipulation.',
```
rabidbit: Avoid including exercise-specific content within the view directive? I.e., to me, it appears this line belongs somewhere else, like i18n.js

client/components/LearnerViewDirective.js
rabidbit: Right, the idea was "Exercise: " would be included in the {{title}}. So, the title can be whatever the content writer wants it to be.

client/components/LearnerViewDirective.js
seanlip: Good catch. I had considered doing that, actually, but in my experience it would be better to do that a bit later, when the language and data schema have stabilized a bit more. Otherwise there's a lot of interim churn/indirection and I don't think that is necessary at this stage. Ultimately, I would like to grab the "focus point" for the exercises from the data file, but we need a data-driven way to represent a set of exercises before that happens.

So, I'd prefer to pull this out when we start supporting these additional use cases (e.g. separate languages), or making plans to do so, but not before. Does that sound OK?

client/components/LearnerViewDirective.js
seanlip: In that case, I disagree. I think that "Exercise 1" is a property of the sequence of exercises, and not a property of the individual exercise. We do not have a data-driven way to represent a sequence of lessons yet, so we'll need to handle that in code until we can factor it out (similar to my other comment about the focus point of the sequence).

seanlip: (Also, the alternative -- which, IIUC, entails putting "Exercise 1: Internationalization" as the title in the question JS file -- seems worse to me, since it implicitly breaks encapsulation.)

rabidbit: Can the internal properties of exercises be separated from the view presented to the user? Not all exercises will be "Exercise 1" (e.g., if they are stand-alone). Maintenance and use of an exerciseNumber for application logic seems fair, but requiring that property to be displayed to the user in a specific format appears to be a UI judgement call.

client/components/LearnerViewDirective.js
rabidbit: My preference would be to not introduce temporary data here, if at all possible, but if it is not possible to move this line into the question/exercise/content file or would require too much work for the Alpha version, then perhaps a TODO comment to remind us to address this later.

seanlip: Ahh -- there are two issues here, actually. On the one hand, yes, if this is question specific, it should go into the data schema ( /cc @eyurko ).  On the other hand, the text that's currently in it is at the set-of-questions level.

So this needs further fleshing out (about what data belongs at which level), and for now I think that a TODO comment is best.

client/components/LearnerViewDirective.js
seanlip: The internal properties of exercises _are_ separate from the view presented to the user --that is the point of the JS data files.

However, the fact that a particular exercise is "Exercise 1" is not an internal property of the exercise. It's a property of where it is in the set-of-exercises, and the latter is where this information should be maintained. Currently, that is done in the application logic, but I think that it would be better for this to be more data driven and I think that in the future we should introduce a set-of-questions data object as well. I will look into this.

With regards to how to display "Exercise 1" (as opposed to, say, "Question 1"?), that's at the level of either a set-of-exercises, or even higher (e.g. if we use the same display for all sets of exercises within the application). I'm not sure whether, by UI judgement call, you mean a customization at the same level as the color of the background of the page (for example), but if that's the case I see no reason why it should be treated any differently for the present (i.e. just included as part of the code).

client/components/LearnerViewDirective.js
seanlip: Added TODO and submitting. Thanks for the review!

client/components/LearnerViewDirective.js
rabidbit: Per offline discussion, I see the value in keeping "Exercise {{exerciseNumber}}:" there and agree.

client/components/LearnerViewDirective.js
rabidbit: Per offline discussion, we ultimately want the intro text to be customizable and unique to each question set. It's fine to leave this as a TODO for this commit and address in a subsequent commit, ideally for Alpha release.

seanlip: Addressed in https://github.com/google/tie/pull/20.

PR #20, opened 2017-02-25
Add QuestionSet domain object. Update Skulpt to fix some bool representation errors.

seanlip:
This PR adds QuestionSets in order to support navigation between multiple questions. Please note that the stepper UI is intentionally minimally functional; @rabidbit plans to replace this with a shinier one next week.

I had to do some yak-shaving to get the new "first" problem to work (parens.js) since Skulpt was representing booleans as 1 and 0. That's why the third-party lib is updated.

@eyurko -- just to check, have I got the questions in the order you intended them (i.e. parens, reverseWords, i18n, rle)? Parens seems a bit hard to me for a question 1, but I can't remember what you mentioned last time.

/cc @apicardgoog


client/components/LearnerViewDirective.js
```diff
@@ -192,26 +205,52 @@ tie.directive('learnerView', [function() {
...
       'LANGUAGE_PYTHON',
       function(
-          $scope, SolutionHandlerService, QuestionDataService,
+          $scope, $timeout, SolutionHandlerService, QuestionDataService,
           LANGUAGE_PYTHON) {
         var language = LANGUAGE_PYTHON;
-        var question = QuestionDataService.getData();
-        var prompts = question.getPrompts();
-        var currentPromptIndex = 0;
+        var questionSetId = 'strings';
```
rabidbit: questionSetId will not always be 'strings', no? If not, add a TODO to assign the appropriate value based on a reaction to some future/as yet undefined user interaction?

seanlip: I think we don't need a TODO for something like this -- in my experience, TODOs tend to rot (since there's no good system for keeping track of them well) and I would lean towards not having them where they aren't necessary. In the future, when we expand the requirements of the app to "handle multiple question sets", this will definitely need to be changed anyway -- it isn't a subtle thing that would otherwise go ignored.

eyurko:
I'd say swap reverseWords and parens.

client/components/LearnerViewDirective.js
rabidbit: It is unclear to me how one would make a distinction between something we know will need to be changed that requires a TODO indicator and something we know will need to be changed but does not require a TODO indicator. It may be obvious to the author what is and isn't subtle, but for anyone else who reads this code and would have the same question, the TODO indicates the author is aware of the issue and they intend to change the code when they are able to. I am not convinced that tracking TODOs is an issue.

seanlip: I think my benchmark was something along the lines of: if it's going to be in the project tracker list, it doesn't need to be in the code. However, I do like the (implicit) point about having the comment in close proximity to the code. I'll make the change.

I think this is also pointing to: let's also have a discussion in the next meeting about how we, as a team, want to keep track of TODOs.

seanlip: Done.

seanlip:
Thanks @eyurko -- done.

Also, merging. Thanks for the reviews!

PR #21, opened 2017-02-28
Add performance tests.

eyurko:


client/services/FeedbackGeneratorService.js
```diff
@@ -52,6 +52,21 @@ tie.factory('FeedbackGeneratorService', [
               ].join(''), false);
             }
           }
+          // TODO(eyurko): Make this fancier. Add in more data points.
```
seanlip: Please elaborate on what "fancier" means. TODOs should, in principle, have enough information to be tackle-able by anyone.

client/services/SolutionHandlerService.js
```diff
@@ -44,7 +44,10 @@ tie.factory('SolutionHandlerService', [
           var preprocessedCode =
             CodePreprocessorDispatcherService.preprocessCode(
               language, code, prompt.getMainFunctionName(),
-              prompt.getCorrectnessTests());
+              prompt.getCorrectnessTests(), prompt.getPerformanceTests());
+          console.log(prompt);
```
seanlip: Drop these before pushing, please.

client/services/code_evaluators/PythonCodeRunnerService.js
```diff
@@ -34,7 +34,13 @@ tie.factory('PythonCodeRunnerService', [
       runCodeAsync: function(code) {
         clearOutput();
         Sk.configure({
-          output: addOutputLine
+         output: addOutputLine,
+         read: function(name) {
```
seanlip: Maybe add brief comment explaining that this snippet enables importing Python stdlib modules, such as `time`.

client/services/FeedbackGeneratorService.js
```diff
@@ -52,6 +52,21 @@ tie.factory('FeedbackGeneratorService', [
...
+          var performanceTests = prompt.getPerformanceTests();
+          for (var i = 0; i < performanceTests.length; i++) {
+            testIndex = i + correctnessTests.length;
+            var expectedPerformance = performanceTests[i].getExpectedPerformance();
+            var observedOutput = observedOutputs[testIndex];
```
seanlip: observedPerformance?

client/services/FeedbackGeneratorService.js
```diff
@@ -52,6 +52,21 @@ tie.factory('FeedbackGeneratorService', [
...
+
+            if (expectedPerformance !== observedOutput) {
+              return FeedbackObjectFactory.create([
+                'Your code is running pretty slowly. Are you sure that it\'s ',
```
seanlip: Well, you haven't really asked them before what they think the performance is, so "Are you sure" is a bit odd.

Maybe: Could you modify it so that it runs in XXX time?

client/services/FeedbackGeneratorServiceSpec.js
```diff
@@ -37,5 +37,5 @@ describe('FeedbackGeneratorService', function() {
           questionMock, codeEvalResult).getMessage()
       ).toEqual(['Your code threw an error: ERROR MESSAGE']);
     });
-  })
+  });
```
seanlip: Oops, thanks! My bad.

client/services/code_preprocessors/CodePreprocessorDispatcherService.js
```diff
@@ -22,13 +22,32 @@ tie.factory('CodePreprocessorDispatcherService', [
...
+          language, code, mainFunctionName, correctnessTests,
+           performanceTests) {
         if (language === LANGUAGE_PYTHON) {
           return PythonCodePreprocessorService.preprocessCode(
+            code, mainFunctionName, correctnessTests, performanceTests);
+        } else {
+          throw Error('Language not supported: ' + language);
+        }
+      },
+      preprocessCorrectnessCode: function(
```
seanlip: I don't understand why we need these extra functions at this level. Shouldn't they be subsidiary to preprocessCode and only added in the PythonCodePreprocessorService? In the grand scheme of things, we are processing the entire code just once and running that preprocessed code just once.

client/services/code_preprocessors/PythonCodePreprocessorService.js
```diff
@@ -22,7 +22,37 @@ tie.factory('PythonCodePreprocessorService', [
...
+          return START_INDENT + line;
+        }
+      });
+      var newCodeLines = [firstLine].concat(systemFunctions);
+      newCodeLines = newCodeLines.concat(subsequentLines);
+      return newCodeLines.join('\n');
+    };
+
     var jsonVariableToPython = function(jsonVariable) {
+      console.log('var = ' + jsonVariable);
```
seanlip: Please remove; ditto elsewhere.

client/services/code_preprocessors/PythonCodePreprocessorService.js
```diff
@@ -54,43 +93,66 @@ tie.factory('PythonCodePreprocessorService', [
...
       });
 
       return testCode;
-    }
+    };
 
-    return {
-      // Wraps a WRAPPER_CLASS_NAME class around the series of functions in a
-      // given code snippet.
-      _wrapCodeIntoClass: function(code) {
```
seanlip: I think you're breaking a bunch of tests by removing this; there are existing tests that test _wrapCodeIntoClass as a standalone thing. Those tests may need to be modified.

client/services/code_preprocessors/PythonCodePreprocessorService.js
```diff
@@ -54,43 +93,66 @@ tie.factory('PythonCodePreprocessorService', [
...
-        var codeLines = code.trim().split('\n');
+    var generatePerformanceTestCode = function(performanceTests) {
+      performanceTests.forEach(function(test, index) {
+        var qualifiedEvaluationFunctionName = (
+          WRAPPER_CLASS_NAME + '().' + test.getEvaluationFunction());
+        var qualifiedTransformationFunctionName = (
+          WRAPPER_CLASS_NAME + '().' + test.getTransformationFunction());
+        // TODO(eyurko): Make this work for other runtimes.
```
seanlip: Can you enumerate what these are?

client/services/code_preprocessors/PythonCodePreprocessorService.js
```diff
@@ -54,43 +93,66 @@ tie.factory('PythonCodePreprocessorService', [
...
+        // TODO(eyurko): 30 seems very scientific. Maybe stop hardcoding?
```
seanlip: Unclear exactly what needs to be done here. Could you rewrite to specify what exactly you want to do (or want someone else to do) instead?

seanlip: Also, for now, at least give the three constants names and pull them out. You could do something like this, for example:

```
    SMALL_INPUT_SIZE = 10
    LARGE_INPUT_SIZE = 100
    UPPER_BOUND_RATIO_IF_LINEAR = (LARGE_INPUT_SIZE / SMALL_INPUT_SIZE) * 3
```

(feel free to pick better descriptions, but at least give some indication of what is going on).

client/services/code_preprocessors/PythonCodePreprocessorService.js
```diff
@@ -54,43 +93,66 @@ tie.factory('PythonCodePreprocessorService', [
...
-        return newCodeLines.join('\n');
+    return {
+      preprocessCode: function(
+      code, mainFunctionName, correctnessTests, performanceTests) {
+        return (
+          wrapCodeIntoClass(code) + '\n' +
+          generateTestCode(
+              mainFunctionName, correctnessTests, performanceTests));
+      },
+      preprocessCorrectnessCode: function(code, mainFunctionName, correctnessTests) {
```
seanlip: Actually I'm not even sure why we need these extra functions at this level. Are they called? (I know you call them in the dispatcher but I don't think anything calls _those_ functions.)

You might want to expose them here with an underscore 'private' indicator for testing purposes (like I did for _wrapCodeIntoClass), but I don't think these should be part of the service's API.

client/services/code_preprocessors/PythonCodePreprocessorServiceSpec.js
```diff
@@ -190,5 +190,25 @@ describe('PythonCodePreprocessorService', function() {
...
+            "transformation_function": "extendString",
+            "expected_performance": "linear",
+            "evaluation_function": "katamariDamashi"
+          }
+        ];
+
+        outputCode = PythonCodePreprocessorService.generatePerformanceTestCode(
+                performanceTests);
+
+        expect(outputCode).toContain('katamariDamashi');
```
seanlip: Should assert equality with expected code, no?

client/services/code_preprocessors/PythonCodePreprocessorService.js
eyurko: I'm not removing _wrapCodeIntoClass, I thought -- isn't it further up in the file?

client/services/code_preprocessors/PythonCodePreprocessorService.js
eyurko: Right -- sorry, only left these in so that we could try to fix the testing thing. Normally before a PR I'd remove these.

client/services/code_preprocessors/CodePreprocessorDispatcherService.js
eyurko: I think this was an accidental holdover. I've dropped them.

client/services/code_evaluators/PythonCodeRunnerService.js
eyurko: Good call.

client/services/SolutionHandlerService.js
eyurko: Done.

client/services/FeedbackGeneratorServiceSpec.js
eyurko: No problem.

client/services/FeedbackGeneratorService.js
eyurko: Got it.

client/services/FeedbackGeneratorService.js
eyurko: Yup!

client/services/FeedbackGeneratorService.js
eyurko: Got it.

client/services/code_preprocessors/PythonCodePreprocessorServiceSpec.js
eyurko: yeah

client/services/code_preprocessors/PythonCodePreprocessorService.js
eyurko: Dropped them -- we don't need them.

client/services/code_preprocessors/PythonCodePreprocessorServiceSpec.js
eyurko: It seems like, long-term, that's going to cause us problems, though, because as we change the performance code we're going to have to update all this hardcoded output code to be consistent with whatever paradigm we choose to enforce going forward. Part of the reason I was only testing the contains was that I didn't see any similar method for correctness, and I'd rather just check for what needs to be in the output code rather than check to ensure that the arbitrary structure is identical.

I'm not suggesting that we NOT do the code equality check, but I think we should be mindful of accruing technical debt. Perhaps it'd be worth writing a helper method to autogenerate the code, test to make sure THAT works, and then check the autogenerated code to make sure it contains the correct parameters for correctness, performance, style, and buggy_output tests.

client/services/code_preprocessors/PythonCodePreprocessorService.js
eyurko: Oops, missed these comments. Will do.

client/services/code_preprocessors/PythonCodePreprocessorService.js
eyurko: got it.

client/services/code_preprocessors/PythonCodePreprocessorService.js
seanlip: Oh, I meant it might need to stay in the return dict so that it can be tested. Do the tests run and pass on your local machine with the new code?

client/services/code_preprocessors/PythonCodePreprocessorServiceSpec.js
seanlip: That's fair, and I agree with the maintainability concern.

For now, can we go mid-way: assert containment, but include a longer portion of the part containing it? That way we know it's in roughly the right place/context.

client/services/code_preprocessors/CodePreprocessorDispatcherService.js
seanlip: They're still here...?

client/services/FeedbackGeneratorService.js
```diff
@@ -52,6 +53,22 @@ tie.factory('FeedbackGeneratorService', [
               ].join(''), false);
             }
           }
+          // TODO(eyurko): Potentially use regression to check runtime complexity, rather than just guessing.
```
seanlip: Why do you need this TODO here? It seems like it's being handled in the preprocessor already, and that this isn't quite the right place for it. (Unless I'm misunderstanding, since I'm not sure what precisely you mean by "use regression to check runtime complexity".)

client/services/code_preprocessors/PythonCodePreprocessorService.js
```diff
@@ -21,8 +21,12 @@ tie.factory('PythonCodePreprocessorService', [
   'WRAPPER_CLASS_NAME', 'VARNAME_TEST_RESULTS',
   function(WRAPPER_CLASS_NAME, VARNAME_TEST_RESULTS) {
     var START_INDENT = '    ';
+    var SMALL_INPUT_SIZE = 10;
+    var LARGE_INPUT_SIZE = 100;
+    var UPPER_BOUND_RATIO_IF_LINEAR = (LARGE_INPUT_SIZE / SMALL_INPUT_SIZE) * 3;
 
     var jsonVariableToPython = function(jsonVariable) {
+      console.log('var = ' + jsonVariable);
```
seanlip: Ahem :)

client/services/code_preprocessors/PythonCodePreprocessorService.js
```diff
@@ -54,20 +67,59 @@ tie.factory('PythonCodePreprocessorService', [
...
+      return testCode;
+    };
+
+    var _generatePerformanceTestCode = function(performanceTests) {
+      performanceTests.forEach(function(test, index) {
+        var qualifiedEvaluationFunctionName = (
+          WRAPPER_CLASS_NAME + '().' + test.getEvaluationFunction());
+        var qualifiedTransformationFunctionName = (
+          WRAPPER_CLASS_NAME + '().' + test.getTransformationFunction());
+        // TODO(eyurko): Make this work for non-linear runtimes, such as quadratic, log(n), and sqrt(n).
```
seanlip: Can you wrap these lines to 80 chars max?

client/services/code_preprocessors/PythonCodePreprocessorService.js
```diff
@@ -54,20 +67,59 @@ tie.factory('PythonCodePreprocessorService', [
...
+          '        output = ' + qualifiedEvaluationFunctionName + '(get_test_input(test_input, input_size))',
+          '        finish = time.time() - start',
+          '        time_array.append(finish)',
+          '    return time_array',
+          '    if time_array[1] > ' + UPPER_BOUND_RATIO_IF_LINEAR + ' * time_array[0]:',
+          '        return "not linear"',
+          '    return "linear"',
+          ''
+        ].join('\n');
+        console.log(test);
```
seanlip: Um.

client/services/code_preprocessors/PythonCodePreprocessorService.js
```diff
@@ -83,14 +135,17 @@ tie.factory('PythonCodePreprocessorService', [
...
-
-        var newCodeLines = [firstLine].concat(subsequentLines);
+        var newCodeLines = [firstLine].concat(systemFunctions);
+        newCodeLines = newCodeLines.concat(subsequentLines);
         return newCodeLines.join('\n');
       },
-      preprocessCode: function(code, mainFunctionName, correctnessTests) {
+      
+      preprocessCode: function(
+      code, mainFunctionName, correctnessTests, performanceTests) {
```
seanlip: Indent by 4 -- would usually be 2, but an additional 2 helps separate this line visually from the next.

client/services/code_preprocessors/PythonCodePreprocessorServiceSpec.js
seanlip: Btw, what you've done is fine too. We can rework this later.

seanlip:
@eyurko -- sorry, you might need to do another merge since I brought the other PR Into master. The main change is that you don't need to copy/paste the question data dict into QuestionDataService.js anymore, which should be a plus :)

(If you need to modify things locally for testing, just change the order of the questions in assets/question_sets/strings.js.)

eyurko:
Should be merged and 80 char lines in the code I've added, now.

PR #23, opened 2017-03-01
Add support for targeted feedback for specific buggy functions.

seanlip:
/cc @rabidbit

### DIFF
--- a/assets/questions/i18n.js
+++ b/assets/questions/i18n.js
@@ -128,7 +128,7 @@ def createListOfUniqueStrings(atom, size):
     }],
     performanceTests: [{
       inputDataAtom: 'm',
-      transformationFunction: 'extendString',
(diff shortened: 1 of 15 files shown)


seanlip:
The main purpose of this PR is to add support for buggy outputs, but along the way I made a few design decisions, which I'll call out specifically here:

- Introduce a "System" namespace for code that is specified by us (and not in question data dicts). So we essentially have three namespaces: System (us), AuxiliaryCode (by the question writer), and StudentCode (by the student).
- Split the generic 'test results' into separate things, since it's conceptually really muddy to use test_results for everything.
- Refactor the preprocessor to improve conceptual clarity of what's going on.

I appreciate any feedback.

seanlip:
Great, thanks! Will merge.

PR #22, opened 2017-02-28
Run Karma presubmit checks automatically.

seanlip:
This replaces #18. Trying to trigger auto-presubmit checks by creating a new branch.

~~Not for review yet. Will send a ping once it's ready.~~

Edit: Now ready for review.

PR #27, opened 2017-03-03
Revised UI with navigation and new next button.

rabidbit:
Error running local tests on my Mac. Away from office; so, cannot test on my workstation. Creating pull request.

### DIFF
--- a/client/components/LearnerViewDirective.js
+++ b/client/components/LearnerViewDirective.js
@@ -25,16 +25,13 @@ tie.directive('learnerView', [function() {
         <div class="tie-exercise-ui-inner">
           <div class="tie-step-container-outer">
             <div class="tie-step-container-inner">
-              <div class="tie-step-item" ng-repeat="questionId in questionIds\
-                  track by $index"
+              <div class="tie-step-item"
+                  ng-repeat="questionId in questionIds track by $index"
                   ng-click="navigateToQuestion($index)">
-                <div class="tie-step-circle" ng-class="{'tie-step-active':\
-                    currentQuestionIndex === $index, 'tie-step-unlocked':\
-                    unlockedQuestions[$index]}">
+                <div class="tie-step-circle" ng-class="{'tie-step-active': currentQuestionIndex === $index, 'tie-step-unlocked': unlockedQuestions[$index]}">
                   <span class="tie-step-text">{{$index + 1}}</span>
                 </div>
-                <div ng-class="{'tie-step-line':\
-                    $index < (questionIds.length - 1)}"></div>
+                <div ng-class="{'tie-step-line': $index < (questionIds.length - 1)}"></div>
               </div>
             </div>
           </div>
@@ -52,10 +49,10 @@ tie.directive('learnerView', [function() {
                 <div class="tie-coding-terminal">
                   <ui-codemirror ui-codemirror="codeMirrorOptions"
                       ng-model="code"
-                      class="tie-codemirror-container"></ui-codemirror>
+                      class="tie-codemirror-container">
+                  </ui-codemirror>
                 </div>
-                <select class="tie-lang-select-menu"
-                    name="lang-select-menu">
+                <select class="tie-lang-select-menu" name="lang-select-menu">
                   <option value="Python" selected>Python</option>
                 </select>
                 <button class="tie-run-button"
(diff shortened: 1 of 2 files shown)


client/components/LearnerViewDirective.js
```diff
@@ -23,12 +23,19 @@ tie.directive('learnerView', [function() {
...
         <div class="tie-exercise-ui-inner">
-          <div>
-            Stepper:
-            <div ng-repeat="questionId in questionIds track by $index">
-              <span ng-class="{'tie-active-question-index': currentQuestionIndex === $index}">
-                {{$index + 1}}
-              </span>
+          <div class="tie-step-container-outer">
+            <div class="tie-step-container-inner">
+              <div class="tie-step-item" ng-repeat="questionId in questionIds\
```
seanlip: Don't end lines with backslash please -- hard to tell what char is being escaped, there could be a space at the end. Prefer long lines to ending with backslash, or alternatively you could move ng-repeat to next line.

client/components/LearnerViewDirective.js
```diff
@@ -23,12 +23,19 @@ tie.directive('learnerView', [function() {
...
+                  track by $index"
+                  ng-click="navigateToQuestion($index)">
+                <div class="tie-step-circle" ng-class="{'tie-step-active':\
+                    currentQuestionIndex === $index, 'tie-step-unlocked':\
+                    unlockedQuestions[$index]}">
+                  <span class="tie-step-text">{{$index + 1}}</span>
+                </div>
+                <div ng-class="{'tie-step-line':\
+                    $index < (questionIds.length - 1)}"></div>
```
seanlip: If you're breaking a tag multi-line, consider moving the closing tag to a separate line. Easier to keep track of indentation that way, since one can eyeball that the `<div>` and `</div>` have the same indentation. Ditto elsewhere.

client/components/LearnerViewDirective.js
```diff
@@ -239,6 +318,7 @@ tie.directive('learnerView', [function() {
           $scope.previousInstructions = [];
           $scope.nextButtonIsShown = false;
           $scope.feedbackParagraphs = introParagraphs;
+          $scope.unlockedQuestions[$scope.currentQuestionIndex] = true;
```
seanlip: My first thought was -- but it's an array!

Then my next thought was -- but JS arrays are objects behind the scenes, so this does kind of work.

Still... what do you think of filling the array with `false` to start with? Then setting it makes more sense conceptually. I do think there is a conceptual difference between false and undefined.

client/components/LearnerViewDirective.js
```diff
@@ -284,6 +368,14 @@ tie.directive('learnerView', [function() {
           }
         };
 
+        $scope.navigateToQuestion = function(index) {
+          if ($scope.unlockedQuestions[index]) {
+            $scope.currentQuestionIndex = index;
+            var questionId = $scope.questionIds[$scope.currentQuestionIndex];
+            loadQuestion(questionId, questionSet.getIntroductionParagraphs());
+          }
+        }
```
seanlip: missing semicolon

client/components/LearnerViewDirective.js
```diff
@@ -284,6 +368,14 @@ tie.directive('learnerView', [function() {
           }
         };
 
+        $scope.navigateToQuestion = function(index) {
+          if ($scope.unlockedQuestions[index]) {
+            $scope.currentQuestionIndex = index;
+            var questionId = $scope.questionIds[$scope.currentQuestionIndex];
+            loadQuestion(questionId, questionSet.getIntroductionParagraphs());
```
seanlip: This is interesting (the fact that we pass intro paragraphs regardless of whether the question is the first one or not). I guess it makes sense, though it doesn't necessarily match history. Happy to go with it for now.

seanlip:
Update: I made the test fix in https://github.com/google/tie/pull/22/commits/ed8294e7edae38248ab896149ca2a5ad88d17a23. Sorry @rabidbit. Hopefully this sorts it out (though that branch still needs to get merged into master before we'll know).

client/components/LearnerViewDirective.js
rabidbit: I could do this...

        $scope.questionIds = questionSet.getQuestionIds();
        $scope.unlockedQuestions = [];
        for (var i = 1; i < $scope.questionIds.length; i++) {
          $scope.unlockedQuestions[i] = false;
        }

Something like that? Any better way to do this?

client/components/LearnerViewDirective.js
rabidbit: We may need an intro for each exercise because when you navigate between questions, the intro text for the first exercise may or may not apply to the subsequent questions. At a minimum, use a generic intro for each exercise. My gut feeling is that a unqiue intro for each exercise will make it seem more conversational. Right now, without an intro for each exercise, navigating between exercises results in neither, which is not a good UX.

client/components/LearnerViewDirective.js

client/components/LearnerViewDirective.js

client/components/LearnerViewDirective.js

client/components/LearnerViewDirective.js
seanlip: Yep -- I'd go with `$scope.unlockedQuestions.push(false)` for the line in the for loop, but otherwise it looks good to me IMO.

client/components/LearnerViewDirective.js
seanlip: Sg. Let's do what you're doing now, we can improve on it later if needed.

client/components/LearnerViewDirective.js

rabidbit:
PTAL

PR #29, opened 2017-03-07
Unbreak performance test

eyurko:
So this isn't a fix and doesn't include linear regression, but I figured while I read up on that I can at least delete the line of code that's causing the issue for y'all. (Currently it looks like skulpt supports numpy but it doesn't say anything about scipy, which is where the linear regression model lives, so I may have to do a bit more work with this bit. Will update.)

I've tested and verified that this allows you to at least proceed through the end of question 2, but it won't work for i18n because that requires a constant-time solution, which is something that needs added.

To that end, I think it's likely better to get constant time working so that we can move through the entire flow, or to remove it altogether.

### DIFF
--- a/client/services/code_preprocessors/PythonCodePreprocessorService.js
+++ b/client/services/code_preprocessors/PythonCodePreprocessorService.js
@@ -168,7 +168,6 @@ tie.factory('PythonCodePreprocessorService', [
             '(get_test_input(test_input, input_size))'),
           '        finish = time.time() - start',
           '        time_array.append(finish)',
-          '    return time_array',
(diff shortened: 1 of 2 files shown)


seanlip:
Hm, this looks like it includes stuff from #28 (see the "Files changed" tab). Is that intentional?

Otherwise -- I see where the bug comes from :) LGTM once you get the above sorted; this PR should contain just those two lines.

Btw, I don't think you need scipy for linear regression -- that's a bit heavyweight. Finding a best-fit 2D line is fairly straightforward math, and calculating residuals is not too bad either; let's . For the question about constant-time, I recommend adding that too; you basically want the slope of the regression line to be near zero.

Thanks!

eyurko:
Sorry about that; all cleaned up, now.

PR #30, opened 2017-03-08
UI fixes: Add timestamp to feedback. Allow jumping between questions.

seanlip:
/cc @rabidbit

### DIFF
--- a/client/components/LearnerViewDirective.js
+++ b/client/components/LearnerViewDirective.js
@@ -28,7 +28,7 @@ tie.directive('learnerView', [function() {
               <div class="tie-step-item"
                   ng-repeat="questionId in questionIds track by $index"
                   ng-click="navigateToQuestion($index)">
-                <div class="tie-step-circle" ng-class="{'tie-step-active': currentQuestionIndex === $index, 'tie-step-unlocked': completedQuestions[$index]}">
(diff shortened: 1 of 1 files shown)


client/components/LearnerViewDirective.js
```diff
@@ -296,9 +296,9 @@ tie.directive('learnerView', [function() {
 
 
         $scope.questionIds = questionSet.getQuestionIds();
-        $scope.unlockedQuestions = [];
+        $scope.completedQuestions = [];
         for (var i = 0; i < $scope.questionIds.length; i++) {
-          $scope.unlockedQuestions.push(false);
+          $scope.completedQuestions.push(false);
```
eyurko: Is this just part of the initial setup, setting every question as "not completed?"

It might not be a bad idea to set a variable or add a comment here, just for future-proofing.

seanlip: Yep, that's right. I'm not sure what's confusing here, though (or what comment you would like to add)? I mean, it's kinda true by default that when you first open the application, you haven't completed any of the questions...

eyurko: I think it's just "Set up all questions as 'not completed' (false). When a question is completed, this value will be changed to true." or something.

seanlip: Per discussion with @eyurko, renaming to `questionsCompletionStatus` instead so the code is more self-documenting.

PR #35, opened 2017-03-08
Add array support and add consistency to question text.

eyurko:
This should remove any reference to "exercises" and replace it with "questions" as well as allow us to use arrays as input for functions, such as are_all_unique.

Autotested and manually tested.


assets/questions/i18n.js
```diff
@@ -70,9 +75,9 @@ def createListOfUniqueStrings(atom, size):
       expectedOutput: 'f8p'
     }],
     buggyOutputTests: [{
-      buggyFunction: 'forgetLastLetter',
+      buggyFunction: 'AuxiliaryCode.forgetLastLetter',
       messages: [
-        "It looks like your output (%s) doesn't match our expected output (%s).",
+        "It looks like your output doesn't match our expected output.",
```
seanlip: I wonder if we could say something like "In your head, try running your code on the following string: ... Does it give the correct answer?"

Otherwise this seems to hang in the air a bit -- it's not clear what the student should do. If you want to be a bit less specific at this point, you could say something like "It looks like your code contains an error. Can you find it?"

In both cases, though, it's clear to the user what they need to do next. With the current message, I was wondering if I needed to be making guesses about the test cases (i.e. reading the "interviewer's" mind).

assets/questions/i18n.js
```diff
@@ -87,8 +92,8 @@ def createListOfUniqueStrings(atom, size):
   }, {
     instructions: [
       [
-        "It's always important to think about the edge cases for your code. ",
-        'In this case, short strings are not abbreviated by this strategy. ',
+        "Next, consider possible edge cases for your code. One potential ",
+        'edge case is usually short strings. Does your code handle them properly? ',
```
seanlip: Does your code handle them properly? --> Check that your code handles this case properly, and submit it when you're convinced it does.

(or something more specific)

assets/questions/i18n.js
```diff
@@ -87,8 +92,8 @@ def createListOfUniqueStrings(atom, size):
...
         "Make sure your code doesn't try to abbreviate when it's not necessary."
```
seanlip: Actually, "necessary" here is part of the requirements and it actually does seem legitimate to me to represent "and" as "a1d" for consistency; I can also imagine another world where I wouldn't bother to abbreviate a 4-letter word.

So it might be worth being a little clearer here -- perhaps something like "... not necessary -- for example, "and" should remain as "and", but "beet" should change to "b2t".

This is just a suggestion, and other rewordings may be possible too.

assets/questions/i18n.js
```diff
@@ -111,9 +116,9 @@ def createListOfUniqueStrings(atom, size):
       expectedOutput: ''
     }],
     buggyOutputTests: [{
-      buggyFunction: 'useFirstAndLastLetterAndLengthToAbbreviate',
+      buggyFunction: 'AuxiliaryCode.useFirstAndLastLetterAndLengthToAbbreviate',
       messages: [
-        "It looks like your output (%s) doesn't match our expected output (%s).",
+        "It looks like your output doesn't match our expected output.",
```
seanlip: Same comment as above.

assets/questions/parens.js
```diff
@@ -68,25 +71,26 @@ def countNumberOfParentheses(s):
...
+        "This doesn't seem quite right. Check your code and try again.",
         [
           'Are you making sure the parentheses are properly balanced? () ',
           'is balanced, but )( is not.'
         ].join(''),
         [
           "It looks like you're counting the number of parentheses, and if ",
-          "they're equal, returning true. That's not quite correct."
+          "you have the same number of each kind, returning true. That's not ",
+          "quite correct."
```
seanlip: "That's not quite correct. See if you can figure out why."

assets/questions/parens.js
```diff
@@ -68,25 +71,26 @@ def countNumberOfParentheses(s):
...
       ]
     }],
     performanceTests: []
   }, {
     instructions: [
       [
-        'Now, make sure your code handles unexpected cases, such as the ',
-        'empty string.'
+        'Now, double-check your code to make sure it handles unexpected cases, ',
+        'such as the empty string.'
```
seanlip: Once you're sure that it does, submit it to see if you're right.

(or similar)

assets/questions/reverseWords.js
```diff
@@ -26,7 +26,9 @@ globalData.questions['reverseWords'] = {
   },
   auxiliaryCode: {
     python:
-`def forgetLastWord(s):
+`class AuxiliaryCode(object):
```
seanlip: Indentation here seems odd, can we keep it consistent with the other files?

assets/questions/reverseWords.js
```diff
@@ -42,10 +44,10 @@ globalData.questions['reverseWords'] = {
   prompts: [{
     instructions: [
       [
-        'Implement the reverseWords function. This function takes a string of ',
-        'words separated by whitespace and reverses the non-whitespace ',
-        'characters in the words, but not their ordering, preserving the ',
-        'original whitespace.'
+        'For this question, you\'ll implement the reverseWords function. ',
+        'This function takes a string of words separated by whitespace and ',
+        'reverses the non-whitespace characters in the words, but not their ',
```
seanlip: What does "their" refer to? The characters, or the words, or the whitespace?

assets/questions/rle.js
```diff
@@ -26,26 +26,33 @@ globalData.questions['rle'] = {
   },
   auxiliaryCode: {
     python:
-`def skipEncodingAtEndOfString(word):
-    pass
+`class AuxiliaryCode(object):
+    @classmethod
+    def skipEncodingAtEndOfString(cls, word):
```
seanlip: It would be nice to actually write these things at some point :)

assets/questions/rle.js
```diff
@@ -116,9 +126,9 @@ def DecodeEncodedString(word):
   }, {
     instructions: [
       [
-        'Good work! But we need to make sure that the string can accurately ',
-        'be decoded. Think about whether the output of your function will ',
-        'correctly be decoded back to the original string.'
+        'Next, make sure that your method\'s output can accurately be decoded. ',
+        'For each <#x> pair, the decode method will repeat the following character ',
```
seanlip: Not sure what "following" means. Perhaps something like:

For each '<#x>' pair, where # is a number and x is a character, the decode method will repeat the x character # times.

client/services/FeedbackGeneratorServiceSpec.js
```diff
@@ -26,6 +26,27 @@ describe('FeedbackGeneratorService', function() {
...
+    it('should return "stringified" (readable) versions of input variables', 
+      function() {
+      expect(
+        FeedbackGeneratorService._jsToHumanReadable('cat')
+      ).toEqual('"cat"');
+      expect(
+        FeedbackGeneratorService._jsToHumanReadable(1)
+      ).toEqual('1');
+      expect(
+        FeedbackGeneratorService._jsToHumanReadable(true)
```
seanlip: Darn, I just realized that the expected value here is language dependent. Can you add a TODO for me (at jsToHumanReadable) to change the human-readable representation based on language? (E.g. true wouldn't be capitalized if the user was programming in JS.)

client/services/FeedbackGeneratorService.js
```diff
@@ -20,13 +20,15 @@
...
-    var jsToHumanReadable = function(jsVariable) {
+    var _jsToHumanReadable = function(jsVariable) {
       if (typeof jsVariable === 'string') {
         return '"' + jsVariable + '"';
       } else if (typeof jsVariable === 'number') {
         return String(jsVariable);
       } else if (typeof jsVariable === 'boolean') {
         return jsVariable ? 'True' : 'False';
+      } else if (Array.isArray(jsVariable)) {
+        return '[' + jsVariable + ']';
```
seanlip: Hmm, you might have to assemble the stringified elements recursively, actually, and then join them with ', ' (note the space). That would be much more future-proof and would correctly handle e.g. lists of bools.

client/services/code_preprocessors/PythonCodePreprocessorService.js
```diff
@@ -57,6 +57,18 @@ tie.factory('PythonCodePreprocessorService', [
         }
 
         return "'" + pythonStringContent + "'";
+      } else if (Array.isArray(jsonVariable)) {
+        // We have to recursively convert the array's elements to Python variables.
+        var pythonArrayContent = ''
+        if (jsonVariable.length > 0) {
```
seanlip: Instead of special-casing the first element, map everything into a JS array and then join using ', '.

client/services/code_preprocessors/PythonCodePreprocessorServiceSpec.js
```diff
@@ -34,6 +34,29 @@ describe('PythonCodePreprocessorService', function() {
...
   }));
 
+  describe('_jsonVariableToPython', function() {
+    it('should correctly convert a json String to a Python string', function() {
+      expect(
+        PythonCodePreprocessorService._jsonVariableToPython('stringify')
+      ).toEqual("'stringify'");
+    });
+
+    it('should correctly convert a json Array to a Python array (or a string version)'
```
seanlip: What do you mean by "or a string version"?

client/services/code_preprocessors/PythonCodePreprocessorServiceSpec.js
```diff
@@ -34,6 +34,29 @@ describe('PythonCodePreprocessorService', function() {
...
+    });
+
+    it('should correctly convert a nested json Array to a similar Python array)'
+      , function() {
+      expect(
+        PythonCodePreprocessorService._jsonVariableToPython(
+          [["1", "2"], ["3", "4"], ["5", "6"]])
+      ).toEqual("[['1', '2'], ['3', '4'], ['5', '6']]");
+    });
+  });
```
seanlip: Maybe add a case for list of bools, since bools have different representation in different languages.

assets/questions/rle.js
eyurko: I'll probably take that + helper functions on, next. Seems like a good idea.

client/services/FeedbackGeneratorServiceSpec.js
eyurko: Done.

assets/questions/rle.js
eyurko: I think I've modified it such that it makes sense to me. What do you think?

client/services/FeedbackGeneratorService.js
eyurko: How's this?

client/services/code_preprocessors/PythonCodePreprocessorService.js
eyurko: Done.

assets/questions/i18n.js
```diff
@@ -29,26 +29,31 @@ def are_all_unique(words):
...
+        for i in range(size):
+            result.append(atom * (i + 1))
+        return result
 `
   },
   prompts: [{
     instructions: [
       [
-        'Implement the abbreviate function. It takes a string as input and ',
+        'This question will have you implement two functions. First, ',
```
eyurko: Done.

assets/questions/i18n.js
eyurko: I tried this a bit -- let me know what your thoughts are.

assets/questions/i18n.js
eyurko: Okay; sounds good.

assets/questions/i18n.js
eyurko: Fair enough -- I've modified it for clarity.

assets/questions/i18n.js
eyurko: Sounds good.

assets/questions/parens.js
eyurko: Sounds good.

assets/questions/parens.js
eyurko: That'll work.

assets/questions/reverseWords.js
eyurko: Sorry, that's my bad. Fixed.

assets/questions/reverseWords.js
eyurko: The words, in this case. Updated it to be more precise.

client/services/code_preprocessors/PythonCodePreprocessorServiceSpec.js
eyurko: Sorry, it's a string representation of a Python array. I've amended the comment to be clearer.

client/services/code_preprocessors/PythonCodePreprocessorServiceSpec.js
eyurko: So we actually don't support booleans as input (or didn't as of your comment), but we do now.

assets/questions/i18n.js
```diff
@@ -70,9 +74,9 @@ def createListOfUniqueStrings(atom, size):
       expectedOutput: 'f8p'
     }],
     buggyOutputTests: [{
-      buggyFunction: 'forgetLastLetter',
+      buggyFunction: 'AuxiliaryCode.forgetLastLetter',
       messages: [
-        "It looks like your output (%s) doesn't match our expected output (%s).",
+        "Try running your code on 'word'. What's the result?.",
```
seanlip: Do you want to say something like "in your head" or "on a piece of paper" here, so that they don't just click the run button?

assets/questions/i18n.js
```diff
@@ -111,9 +116,9 @@ def createListOfUniqueStrings(atom, size):
       expectedOutput: ''
     }],
     buggyOutputTests: [{
-      buggyFunction: 'useFirstAndLastLetterAndLengthToAbbreviate',
+      buggyFunction: 'AuxiliaryCode.useFirstAndLastLetterAndLengthToAbbreviate',
       messages: [
-        "It looks like your output (%s) doesn't match our expected output (%s).",
+        "Try running your code on 'at'. What does it produce?",
```
seanlip: Ditto, here and elsewhere. May not need changing, but just wanted to check what you thought.

assets/questions/rle.js
```diff
@@ -78,7 +85,10 @@ def DecodeEncodedString(word):
     performanceTests: []
   }, {
     instructions: [
-      'You need to make sure that your code handles short strings correctly.'
+      [
+        'Next, double-check your code to make sure it handles short strings. ',
+        'Ideally, these strings should also be as small as possible.',
```
seanlip: I'm not sure I understand this. What's the difference between short and small?

assets/questions/rle.js
```diff
@@ -116,9 +126,9 @@ def DecodeEncodedString(word):
   }, {
     instructions: [
       [
-        'Good work! But we need to make sure that the string can accurately ',
-        'be decoded. Think about whether the output of your function will ',
-        'correctly be decoded back to the original string.'
+        'Next, make sure that your method\'s output can accurately be decoded. ',
+        'For each <#x{c}> pair, the decode method will repeat the character c',
```
seanlip: Did you test this manually? You're missing a space at the end.

assets/questions/rle.js
```diff
@@ -26,26 +26,33 @@ globalData.questions['rle'] = {
...
   prompts: [{
     instructions: [
       [
-        'Implement the encode function. It takes a string as input and ',
-        'returns an encoding of the string where long runs of characters are ',
-        'replaced by <# characters>x<character>. For example, "abcccccd" ',
-        'should be encoded as "ab5xc".'
+        'In this question, you\'ll implement the encode function. It takes a ',
+        'string as input and returns an encoding of the string where long ',
+        'runs of characters are replaced by <# characters>x<character>. For ',
```
seanlip: Btw just wondering -- in this question is the original string composed only of alphabetic characters? I ask because if numbers are part of the original string, it gets tricky.

client/services/FeedbackGeneratorService.js
```diff
@@ -20,13 +20,20 @@
...
+    var _jsToHumanReadable = function(jsVariable) {
       if (typeof jsVariable === 'string') {
         return '"' + jsVariable + '"';
       } else if (typeof jsVariable === 'number') {
         return String(jsVariable);
       } else if (typeof jsVariable === 'boolean') {
+        // TODO(sll): Change human-readable representation based on language.
         return jsVariable ? 'True' : 'False';
+      } else if (Array.isArray(jsVariable)) {
+        var variable_array = [];
```
seanlip: Naming is lowerCamelCase -- i.e. variableArray (or, more descriptively, something like humanReadableElements).

Also, if you like:

    var humanReadableElements = jsVariable.map(function(arrayElement) {
      return _jsToHumanReadable(arrayElement);
    });

client/services/code_preprocessors/PythonCodePreprocessorService.js
```diff
@@ -57,8 +57,17 @@ tie.factory('PythonCodePreprocessorService', [
         }
 
         return "'" + pythonStringContent + "'";
+      } else if (typeof jsonVariable == 'boolean') {
+        return jsonVariable ? 'True' : 'False';
+      } else if (Array.isArray(jsonVariable)) {
+        // We have to recursively convert the array's elements to Python variables.
+        var variable_array = []
```
seanlip: See comments above.

assets/questions/i18n.js
eyurko: Mmm, sure, that's fine.

assets/questions/i18n.js
eyurko: Will fix.

assets/questions/rle.js
eyurko: Our tests in the beginning only focus on alphabetic characters, but we introduce numbers later in the problem to make it pretty tricky.

This is a pretty difficult problem, to be fair.

assets/questions/rle.js
eyurko: I think the also is redundant, there -- it implies a separate condition that's not needed.

assets/questions/rle.js
eyurko: Sorry, missed that in the manual test.

client/services/FeedbackGeneratorService.js
eyurko: Whoops -- that's my bad. I think the map is a bit tighter, so I'll use that.

client/services/code_preprocessors/PythonCodePreprocessorService.js
eyurko: Fixed.

assets/questions/rle.js
seanlip: The use of two different adjectives suggests two different things though. If you want them to mean the same thing, maybe just use "short" throughout?

assets/questions/rle.js
seanlip: Sorry, I was unclear. Take the strings 7xx and xxxxxxx...x (72 times). Encoding them gives the same thing, 72xx. How would decoding distinguish?

eyurko: That's part of the problem that the learner has to figure out. The encoding scheme is intentionally bad, so the student needs to figure out ways to prevent that. I don't want to put a solution to this problem in this comment chain, so I'm happy to discuss this offline, but the burden of making the encodings distinct is on the learner.

Though technically, for your example, the first string should remain 7xx and the second string should be 72xx. :)

assets/questions/rle.js
seanlip: Oh sorry - please ignore this. Your new version actually reads fine since you've clarified the context.

assets/questions/rle.js
eyurko: (Which, as far as 7xx goes, is still a "bad" encoding because a decoding function would decode that as xxxxxxx, rather than 7xx.)

seanlip:
Lgtm. Just one residual comment but maybe it makes sense to address that separately.

PR #39, opened 2017-03-10
Fix #31: always use spaces for initial line indentation.

### LINKED ISSUE #31: Indentation is often inconsistent and incorrect.
For some reason the editor is conflating tabs and spaces and it's causing all sorts of errors. I usually run into this on the i18n question when adding the length check to abbreviate.

Open app.html
Copy/paste solution.
Add lines to solution (hitting "Enter" after a complete line), and then tab inwards.
You see the correct indentation, but it's now mixed tabs with spaces.

seanlip:


eyurko:
Good.

seanlip:
@rabidbit would you also mind taking a quick look and verifying? Thanks!

rabidbit:
Looks like it resolves the issue. Would suggest you either make the initial template code use 2 spaces or make the tab width 4 spaces. Right now, the starter code is indented by 4 spaces and each tab creates 2. I also think 4 spaces sort of look more like a tab.

seanlip:
That's weird! Each tab actually creates 4 spaces for me, and if you look at the changes I've set the [indentation](https://github.com/google/tie/pull/39/files#diff-16eb143ead485efa4f70231c505afab8R356) and [tab size](https://github.com/google/tie/pull/39/files#diff-16eb143ead485efa4f70231c505afab8R360) to 4. I might pop over and take a look.

@eyurko, do you get similar behavior?

rabidbit:
LGTM

PR #43, opened 2017-03-11
Fix #40: parse all possible JS variables human-readably.

### LINKED ISSUE #40: No feedback to user when parsing fails
If you try the following in the reverse word question:
```
def reverseWords(word):
  if word == "moo cow bark dog":
    return "oom woc krab god"
```
Parsing fails and an error is written to console, but no feedback (or error) is given to the user.

Expected behavior: Some indication to user what went wrong.

Observed behavior: Nothing (unless you open a console and see the obscure error message written there)

seanlip:
Note: this also fixes the error @apicardgoog reported where typing "return {}" results in no error and no system feedback.

### DIFF
--- a/client/services/FeedbackGeneratorService.js
+++ b/client/services/FeedbackGeneratorService.js
@@ -46,6 +46,8 @@ tie.factory('FeedbackGeneratorService', [
         }
         return '{' + humanReadableKeyValuePairs.join(', ') + '}';
       } else {
+        console.error(
+          'Could not make the following object human-readable: ', jsVariable);
         return '[UNKNOWN OBJECT]';
       }
     };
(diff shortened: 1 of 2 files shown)


client/services/FeedbackGeneratorService.js
```diff
@@ -20,21 +20,33 @@
...
+        var humanReadableKeyValuePairs = [];
+        for (var key in jsVariable) {
+          humanReadableKeyValuePairs.push(
+            _jsToHumanReadable(key) + ': ' +
+            _jsToHumanReadable(jsVariable[key]));
+        }
+        return '{' + humanReadableKeyValuePairs.join(', ') + '}';
       } else {
-        throw Error('Cannot parse JS variable: ' + jsVariable);
+        return '[UNKNOWN OBJECT]';
```
apicardgoog: Do you not want to make jsVariable part of the output so we can diagnose what it was?

seanlip: That's a good point, thanks for catching it!

I think I'll have to log a console error for this instead, though. The above code covers all the standard JS types, so anything printed out here is going to just end up being something like [object Object]. I'll add the error logging and merge.

Issue #42, opened 2017-03-11
OutputFunction isn't currently supported.


eyurko:
Any output function defined in the auxiliary code and then specified in the data schema (such as rle.js's decodeEncodedString) is not currently used.

Repro steps:

1. Go to the run-length encoding problem, once I've pushed my fixes.
2. Paste in the solution.
3. Run until you get to the "decode" step.
4. Your output should return 1x5xb for one correctness test, rather than 5xb, signifying that `AuxiliaryCode.decodeEncodedString` was not run on the output.

seanlip:
Addressed in #48, I think.

seanlip:
Fixed in #48. Closing.

PR #44, opened 2017-03-11
Add support for student helper functions

eyurko:
Autotested and manually tested as much as possible.

- Output functions don't currently work, so I cannot fully test decodeEncodedString, so I instead ran it offline with a few test cases. (Written up in #42)
- Multiple possible outputs don't work, currently, either, so you have to modify rle.js to only have one output to progress past that point. (Not yet written up; will write up once PR is merged.)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,4 @@
 node_modules/*
 tools/*
+.DS_Store
+assets/.DS_Store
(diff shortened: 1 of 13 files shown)


seanlip:
NB: the multiple possible outputs thing should be fixed in #47.

PR #47, opened 2017-03-11
Support multiple possible correct outputs.

seanlip:


client/services/code_preprocessors/PythonCodePreprocessorServiceSpec.js
```diff
@@ -255,7 +255,7 @@ describe('PythonCodePreprocessorService', function() {
         })];
         var correctnessTests = [CorrectnessTestObjectFactory.create({
           input: 'cat',
-          expectedOutput: 'at'
+          allowedOutputs: ['at']
```
seanlip: Thanks -- PTAL. Added tests for matchesOutput() and getAnyAllowedOutput(). Also added preprocessor test for generating correctness code (and removed an unneeded arg from the buggy outputs generator in the process).

PR #48, opened 2017-03-11
Add support for custom output functions when running correctness tests.

seanlip:


client/domain/PromptObjectFactory.js
```diff
@@ -59,6 +59,10 @@ tie.factory('PromptObjectFactory', [
       return this._mainFunction;
     };
 
+    Prompt.prototype.getOutputFunctionName = function() {
+      return this._outputFunction;
```
seanlip: @eyurko -- I'm wondering, shall we change outputFunction to outputFunctionName in the schema? I think it'd be good to distinguish clearly between the function and the name of the function, but would appreciate your thoughts on this.

eyurko: I'm not bothered by it either way, but for consistency we'd probably need to make them all (Input|Transformation|Evaluation|Output)FunctionName in the schema.

seanlip: OK, thanks. I'll take care of that in a separate PR, then.

PR #52, opened 2017-03-15
Add nullcheck for errorMessage.

eyurko:


client/services/FeedbackGeneratorService.js
```diff
@@ -58,7 +58,8 @@ tie.factory('FeedbackGeneratorService', [
         var errorMessage = codeEvalResult.getErrorMessage();
         // We want to catch and handle a timeout error uniquely, rather than
         // integrate it into the existing feedback pipeline.
-        if (errorMessage.toString().startsWith('TimeLimitError')) {
+        if (errorMessage !== null &&
```
seanlip: Ah, I missed this. Looks good, but I suggest making the following "if" an "else if" so that it's clear that at most one of these branches will be executed.

eyurko: Sure, that's fine. Yeah, sorry, I missed it too -- I didn't realize errorMessage was potentially nullable.

client/services/FeedbackGeneratorService.js
```diff
@@ -67,7 +68,7 @@ tie.factory('FeedbackGeneratorService', [
             "more efficiently?"].join(''));
           return feedback;
         }
-        if (errorMessage) {
+        else if (errorMessage) {
```
seanlip: One small style nit: can you move this to the previous line? So:

`} else if (errorMessage) {`

That way, it's clearer it's part of the same block. Eventually this will be in a lint check too, but thought I'd mention it here so that the alternative pattern doesn't propagate.

PR #53, opened 2017-03-16
Create Transcript and Snapshot ObjectFactories.

eyurko:
Seemed like a good idea to make the initial PR to get some design feedback. There are a few sticky spots, here.

### DIFF
--- a/client/domain/SnapshotObjectFactory.js
+++ b/client/domain/SnapshotObjectFactory.js
@@ -19,9 +19,9 @@
 
 tie.factory('SnapshotObjectFactory', [
   function() {
-    var Snapshot = function(codeEvalResult, feedbackObject) {
+    var Snapshot = function(codeEvalResult, feedback) {
       this._codeEvalResult = codeEvalResult;
-      this._feedback = feedbackObject;
+      this._feedback = feedback;
       this._timestamp = '';
     };
 
@@ -38,13 +38,13 @@ tie.factory('SnapshotObjectFactory', [
       return this._feedback;
     };
 
-    Snapshot.prototype.setFeedback = function(feedbackObject) {
+    Snapshot.prototype.setFeedback = function(feedback) {
       this._feedback = feedback;
     };
 
     // Static class methods.
-    Snapshot.create = function(codeEvalResult, feedbackObject) {
-      return new Snapshot(codeEvalResult, feedbackObject);
+    Snapshot.create = function(codeEvalResult, feedback) {
+      return new Snapshot(codeEvalResult, feedback);
     };
 
     return Snapshot;
--- a/client/domain/SnapshotObjectFactory.js
+++ b/client/domain/SnapshotObjectFactory.js
@@ -19,7 +19,6 @@
 
 tie.factory('SnapshotObjectFactory', [
   function() {
-    // Add support for a TieFeedback object.
     var Snapshot = function(codeEvalResult, feedbackObject) {
       this._codeEvalResult = codeEvalResult;
       this._feedback = feedbackObject;
--- a/client/domain/SnapshotObjectFactory.js
+++ b/client/domain/SnapshotObjectFactory.js
@@ -1,0 +1,53 @@
+// Copyright 2017 The TIE Authors. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS-IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+/**
+ * @fileoverview Factory for creating a snapshot of a student's code
+ * and performance at a given time.
+ */
+
+tie.factory('SnapshotObjectFactory', [
+  function() {
+    // Add support for a TieFeedback object.
+    var Snapshot = function(codeEvalResult, feedbackObject) {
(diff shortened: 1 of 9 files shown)


client/services/FeedbackGeneratorService.js
```diff
@@ -53,94 +55,102 @@ tie.factory('FeedbackGeneratorService', [
...
+          'You\'ve completed all the tasks for this question! Click the ',
+          '"Next" button to move on to the next question.',
+        ].join(''));
+        return feedback;
+      }
+    };
+
+    return {
+      _prepareFeedback: _prepareFeedback,
+      getFeedback: function(prompt, codeEvalResult) {
```
eyurko: So I think it makes more sense to do the Transcript processing stuff here, but that's not possible for getSyntaxErrorFeedback since CodeEvalResult isn't passed in, so I have to handle it at the Solution level. Does it make more sense to try to pull CodeEvalResult into getSyntaxErrorFeedback, to push all the TranscriptService stuff up to the SolutionLevel, or to handle each individually? Happy to fix it to what's best.

eyurko:
I'm intentionally not doing much in this PR to just get the skeleton in and working -- I'll add the rest of the functionality in a subsequent PR.

client/services/FeedbackGeneratorService.js
seanlip: Let's do the transcript stuff in SolutionHandlerService -- that's the service that just does a bunch of "general management" of things and pushes stuff into various pipelines. I think that's the right "seam".

eyurko: SGTM

client/services/TranscriptService.js
```diff
@@ -18,15 +18,18 @@
...
+  'EventObjectFactory', 'TranscriptObjectFactory', function(EventObjectFactory, TranscriptObjectFactory) {
+    var transcript = TranscriptObjectFactory.create();
 
     return {
+      recordSnapshot: function(snapshot) {
+        transcript.recordSnapshot(snapshot);
+      },
       recordSolution: function(code) {
-        transcript.push(EventObjectFactory.createCodeSubmittedEvent(code));
+        transcript.recordEvent(EventObjectFactory.createCodeSubmittedEvent(code));
```
seanlip: Really, I think it's fine to drop these old functions. We can add them later if needed but let's not have dead code, especially now that you have an actual use for the transcript.

client/domain/TranscriptObjectFactory.js
```diff
@@ -0,0 +1,50 @@
...
+/**
+ * @fileoverview Factory for creating a transcript of a student's TIE 
+ * session.
+ */
+
+tie.factory('TranscriptObjectFactory', [
+  function() {
+    var Transcript = function() {
+      this._snapshots = [];
+      this._events = [];
```
seanlip: Let's drop this.

client/domain/TranscriptObjectFactory.js
```diff
@@ -0,0 +1,50 @@
...
+    Transcript.prototype.getPreviousSnapshot = function() {
+      return this._snapshots[this._snapshots.length - 1];
+    };
+
+    Transcript.prototype.recordSnapshot = function(snapshot) {
+      this._snapshots.push(snapshot);
+      return this._snapshots.length;
+    };
+
+    // TODO(eyurko): Probably not needed. Remove in a subsequent PR.
```
seanlip: Might as well remove now IMO.

client/domain/TranscriptObjectFactory.js
```diff
@@ -0,0 +1,50 @@
...
+
+tie.factory('TranscriptObjectFactory', [
+  function() {
+    var Transcript = function() {
+      this._snapshots = [];
+      this._events = [];
+    };
+
+    // Instance methods.
+    Transcript.prototype.getPreviousSnapshot = function() {
```
seanlip: What if the transcript is empty?

client/domain/SnapshotObjectFactory.js
```diff
@@ -0,0 +1,53 @@
...
+// limitations under the License.
+
+/**
+ * @fileoverview Factory for creating a snapshot of a student's code
+ * and performance at a given time.
+ */
+
+tie.factory('SnapshotObjectFactory', [
+  function() {
+    // Add support for a TieFeedback object.
```
seanlip: Comment should have TODO(eyurko): 

But I don't understand it. It looks like we do have a feedbackObject here already?

seanlip:
One general question, too: you mentioned that you'll add the rest of the functionality in a subsequent PR. Just to make sure we're on the same page, could you explain what "the rest of the functionality" means? Thanks!

eyurko:
Yes -- I updated your TODO (and gave it to myself) to actually use the Transcript / Snapshot objects to verify that the student has submitted _new_ code and then use that to give new hints for buggy output.

This may also require a bit more functionality added to Transcript than just "getPreviousSnapshot", since we'll actually want to return buggyHint[X], where X is the number of times they've submitted _new_ code while consecutively getting this error. I'll think more about it, but getting the base functionality down seems like a useful thing for this PR.

client/domain/SnapshotObjectFactory.js
eyurko: We do! Sorry, I forgot to remove that comment.

client/domain/TranscriptObjectFactory.js
eyurko: Will do.

client/domain/TranscriptObjectFactory.js
eyurko: ooh, great catch. Thank you! I should be being more responsible, here. Added tests.

client/domain/TranscriptObjectFactory.js
eyurko: Done.

client/services/TranscriptService.js
eyurko: Sounds good.

PR #56, opened 2017-03-18
Add multi-hint functionality for buggy outputs, provided the student has changed their code.

eyurko:
Linted and three (count 'em) new tests added.

We should probably add a testing infrastructure, at some point ...


client/services/FeedbackGeneratorService.js
```diff
@@ -55,6 +56,9 @@ tie.factory('FeedbackGeneratorService', [
 
     return {
       getFeedback: function(prompt, codeEvalResult) {
+        // TODO(eyurko): This is getting long enough that we should
```
eyurko: I'd prefer to do this in a separate PR, and am happy to split it after this PR.

client/services/FeedbackGeneratorServiceSpec.js
```diff
@@ -91,5 +98,159 @@ describe('FeedbackGeneratorService', function() {
         "3 seconds) we've set. Can you try to make it run ",
         "more efficiently?"].join(''));
     });
+
+    it(['should return the next hint in sequence for buggy outputs, ',
```
eyurko: There's a fair bit of redundancy, here, and in subsequent tests -- would it be better to make this a global buggyOutputTestDict?

client/services/TranscriptService.js
```diff
@@ -22,6 +22,9 @@ tie.factory('TranscriptService', [
     var transcript = TranscriptObjectFactory.create();
 
     return {
+   	  getTranscript: function(){
```
seanlip: Use spaces not tabs.

eyurko: That's so weird -- it LOOKS fine on mine. Fixing.

client/domain/FeedbackObjectFactory.js
```diff
@@ -22,6 +22,10 @@ tie.factory('FeedbackObjectFactory', [
     var Feedback = function(answerIsCorrect) {
       this._paragraphs = [];
       this._answerIsCorrect = answerIsCorrect;
+      // This records what message was displayed with this feedback.
+      // If no message (or no message with additional potential messages)
+      // was displayed, this will remain at -1.
```
seanlip: Why not null? Some feedback isn't associated with hints anyway.

client/services/FeedbackGeneratorService.js
seanlip: Mmm. How about pulling just the buggy function stuff out into a separate function now? Otherwise you have a giant block of code here -- and then the next PR will contain two giant blocks of code that are largely the same but hard to compare.

client/services/FeedbackGeneratorServiceSpec.js
seanlip: Yes, I think so.

client/services/TranscriptService.js
```diff
@@ -22,6 +22,9 @@ tie.factory('TranscriptService', [
     var transcript = TranscriptObjectFactory.create();
 
     return {
+      getTranscript: function(){
```
seanlip: Add space before {

client/domain/FeedbackObjectFactory.js
eyurko: It lets me +1 it to 0, which is handy instead of having to have it as a separate case. The name suggests an int to me, anyways.

eyurko: Although I don't actually do any work with it as an int, so I can change it to null if you'd prefer -- I usually just use -1 here because the name suggests an int, rather than an object to me.

client/services/TranscriptService.js
eyurko: Sure, done.

client/domain/FeedbackObjectFactory.js
seanlip: In general, null is like None in python, which is a fine value for ints/floats/strings/etc. properties that indicates the absence of a value. So it's totally fine to use it for fields of any given type. (Otherwise you wouldn't ever be able to use it...)

So I think null is fine; -1 seems a bit opaque to me and doesn't necessarily convey "none".

client/services/FeedbackGeneratorService.js
```diff
@@ -53,8 +54,52 @@ tie.factory('FeedbackGeneratorService', [
...
+      if (lastSnapshot !== null) {
+        // This section makes sure to provide a new hint
+        // if the student gets stuck on the same bug by checking
+        // that they've submitted new code with the same error.
+        var previousFeedback = lastSnapshot.getFeedback();
+        var previousHintIndex = previousFeedback.getHintIndex();
+        if (previousHintIndex !== -1 && 
+          previousHintIndex < buggyMessages.length) {
+          var previousMessages = previousFeedback.getParagraphs();
+          // This could cause a problem if two different buggy outputs
```
seanlip: Can you make a note on the data schema doc that this is a requirement, so that we can include a check for it later in our schema validation code?

client/services/FeedbackGeneratorService.js
```diff
@@ -53,8 +54,52 @@ tie.factory('FeedbackGeneratorService', [
...
+          // have the exact same hint, but that shouldn't be allowed.
+          if (previousMessages[0].getContent() == 
+            buggyMessages[previousHintIndex]) {
+            var previousCode = (
+              lastSnapshot.getCodeEvalResult().getCode());
+            if (previousCode === codeEvalResult.getCode()) {
```
seanlip: Can we just set the variable hintIndex here, and leave the construction of the feedback to outside the 'if' branch? That is: if the previous code is the same, set hintIndex to previousHintIndex, otherwise set it to 0 or previousHintIndex + 1. Construct the feedback object after that.

(Aim is to reduce some of the branching. It feels to me like a lot of this logic is just to figure out the correct hint index, and the actual construction of the feedback is orthogonal to that. Ditto for the "0" case later below.)

client/services/FeedbackGeneratorServiceSpec.js
```diff
@@ -92,4 +99,149 @@ describe('FeedbackGeneratorService', function() {
...
+            "It looks like you're counting the number of parentheses, and if ",
+            "you have the same # of each kind, returning true. That's not ",
+            "quite correct. See if you can figure out why."
+          ].join('')
+        ]
+      };
+      var buggyOutputTest = BuggyOutputTestObjectFactory.create(
+        buggyOutputTestDict);
+      var codeEvalResult = CodeEvalResultObjectFactory.create(
+        'some code', 'same output', [], [true], [], false,
```
seanlip: Should the sixth arg be null, and the seventh arg be null, if there is no runtime error? Ditto below.

client/services/FeedbackGeneratorService.js
eyurko: Done.

client/services/FeedbackGeneratorService.js
eyurko: Sure. Fixed it.

client/services/FeedbackGeneratorServiceSpec.js
eyurko: I think either way it doesn't trigger, but I'm fine modifying it.

seanlip: Yep, I agree it doesn't trigger (since we're not testing that part of the code in this test) but I think it's a good idea to make these calls reflect reality. Thanks!

client/services/FeedbackGeneratorService.js
```diff
@@ -53,8 +54,46 @@ tie.factory('FeedbackGeneratorService', [
...
+      var lastSnapshot = (
+        TranscriptService.getTranscript().getPreviousSnapshot());
+      if (lastSnapshot !== null) {
+        // This section makes sure to provide a new hint
+        // if the student gets stuck on the same bug by checking
+        // that they've submitted new code with the same error.
+        var previousFeedback = lastSnapshot.getFeedback();
+        var previousHintIndex = previousFeedback.getHintIndex();
+        if (previousHintIndex !== null && 
+          previousHintIndex < buggyMessages.length) {
```
seanlip: Interesting -- what is supposed to happen if you're at the last hint of a set, and you change the code and still trigger the same error?

client/services/FeedbackGeneratorService.js
```diff
@@ -53,8 +54,46 @@ tie.factory('FeedbackGeneratorService', [
...
+          // This could cause a problem if two different buggy outputs
+          // have the exact same hint, but that shouldn't be allowed.
+          if (previousMessages[0].getContent() == 
+            buggyMessages[previousHintIndex]) {
+            var previousCode = (
+              lastSnapshot.getCodeEvalResult().getCode());
+            if (previousCode === codeEvalResult.getCode()) {
+              hintIndex = previousHintIndex;
+            }
+            hintIndex = previousHintIndex + 1 % buggyMessages.length;
```
seanlip: Shouldn't there be an 'else' here? Otherwise you're just overwriting the previous value in all cases.

Also, not sure I understand the significance of the "%"?

client/services/FeedbackGeneratorService.js
```diff
@@ -53,8 +54,46 @@ tie.factory('FeedbackGeneratorService', [
       }
     };
 
+    var _getBuggyOutputTestFeedback = function (
+      failingTest, codeEvalResult) {
+      var hintIndex = 0;
+      var buggyMessages = failingTest.getMessages();
+      var feedback = FeedbackObjectFactory.create(false);
```
seanlip: Minor: consider moving this below, near where you do all the feedback-related stuff.

client/services/FeedbackGeneratorServiceSpec.js
```diff
@@ -92,4 +99,149 @@ describe('FeedbackGeneratorService', function() {
         "more efficiently?"].join(''));
     });
   });
+
+  describe('_getBuggyOutputTestFeedback', function() {
+    it(['should return the next hint in sequence for buggy outputs, ',
+        'provided the code has been changed'].join(''), function() {
+      var buggyOutputTestDict = {
```
seanlip: It might be worth declaring this more globally (in the 'describe', but outside the 'it') if you're going to reuse it a lot. Up to you, though.

client/services/FeedbackGeneratorServiceSpec.js
```diff
@@ -92,4 +99,149 @@ describe('FeedbackGeneratorService', function() {
...
+        codeEvalResultWithNewError, unusedRuntimeErrorFeedback));
+
+      var feedback = FeedbackGeneratorService._getBuggyOutputTestFeedback(
+        buggyOutputTest, codeEvalResultWithSameBug).getParagraphs();
+
+      expect(paragraphs.length).toEqual(1);
+      expect(paragraphs[0].isTextParagraph()).toBe(true);
+      expect(paragraphs[0].getContent()).toBe(buggyOutputTestDict.messages[0]);
+    });
+  });
```
seanlip: Should we also have tests for the behavior near the end of the list of hints?

eyurko: I added a test for making sure that the message repeats at the end of a list, but other than that it seems like it should behave the same throughout?

client/services/FeedbackGeneratorServiceSpec.js
eyurko: It seems that way, so I will.

client/services/FeedbackGeneratorService.js
eyurko: Ah, yes, sorry, forgot to change that when I rearchitected the method. The modulo was originally to wrap it back around to the start, but I think it's better for the student to see the same "most helpful" hint until they finish, so I've tweaked this.

client/services/FeedbackGeneratorService.js
eyurko: At this point it'd loop around, hence the %, I believe. But it might be better to just stay on the last hint. I'll switch to that, but happy to change back if you prefer the other.

client/services/FeedbackGeneratorService.js
eyurko: Sure, that works.

client/domain/FeedbackObjectFactory.js
```diff
@@ -22,6 +22,10 @@ tie.factory('FeedbackObjectFactory', [
     var Feedback = function(answerIsCorrect) {
       this._paragraphs = [];
       this._answerIsCorrect = answerIsCorrect;
+      // This records what message was displayed with this feedback.
+      // If no message (or no message with additional potential messages)
```
eyurko: Yup, I think this is from an old understanding of how this works.

client/services/FeedbackGeneratorService.js
seanlip: OK, then do you need the previousHintIndex < buggyMessages.length check? Would it not always be true?

client/services/FeedbackGeneratorService.js
```diff
@@ -53,8 +54,48 @@ tie.factory('FeedbackGeneratorService', [
...
+          // have the exact same hint, but that shouldn't be allowed.
+          if (previousMessages[0].getContent() == 
+            buggyMessages[previousHintIndex]) {
+            var previousCode = (
+              lastSnapshot.getCodeEvalResult().getCode());
+            if (previousCode === codeEvalResult.getCode() ||
+                previousHintIndex == buggyMessages.length - 1) {
+              hintIndex = previousHintIndex;
+            } else {
+              hintIndex = previousHintIndex + 1 % buggyMessages.length;
```
seanlip: Why the `% buggyMessages.length`, if we are not looping around?

Also for future reference, % takes precedence over +.

client/services/FeedbackGeneratorServiceSpec.js
seanlip: Yep, sgtm. Thanks!

client/services/FeedbackGeneratorService.js
eyurko: oop, that's fixed on my side. Must not have saved before committing. Sorry about that.

seanlip:
Oh, @eyurko, sorry -- the previous PR merge caused this to have a merge conflict. Would you mind fixing it, and merging after that? Thanks!

PR #57, opened 2017-03-18
Fix #24: fix incorrect line numbers in error messages.

### LINKED ISSUE #24: Runtime errors have the wrong line numbers.
Repro instructions: Go to the reverseWords problem, and type the code seen in the screenshot below.

Observed behaviour: Runtime errors have the wrong line numbers. For example, in the screenshot below, the error is shown as being on line 15, whereas the actual error is on line 3.

Expected behaviour: The error message should show the correct line number.

screenshot from 2017-03-01 15 40 38

seanlip:


seanlip:
Sorry, this PR is a bit large -- it went through a few iterations. The current version represents the most elegant method I can think of to do this (which is to add a new domain object containing the raw and preprocessed code and cleanly maintain the mapping between the two), but I welcome any suggestions for improvement.

client/services/FeedbackGeneratorService.js
```diff
@@ -75,7 +76,28 @@ tie.factory('FeedbackGeneratorService', [
...
+              var preprocessedCodeLineIndex = (
+                Number(humanReadableLineNumber) - 1);
+              if (preprocessedCodeLineIndex < 0 ||
+                  preprocessedCodeLineIndex >= rawCodeLineIndexes.length) {
+                throw Error(
+                  'Line number index out of range: ' + preprocessedCodeLineIndex);
+              }
+
+              if (rawCodeLineIndexes[preprocessedCodeLineIndex] === null) {
+                return 'a line in the test code';
```
eyurko: Interesting. I like this.

client/domain/CodeSubmissionObjectFactorySpec.js
```diff
@@ -0,0 +1,102 @@
...
+      var codeSubmission = CodeSubmissionObjectFactory.create(originalCode);
+
+      expect(codeSubmission.getRawCode()).toEqual(originalCode);
+      expect(codeSubmission.getPreprocessedCode()).toEqual(
+        'line 1\nline 2\nline 3');
+      expect(codeSubmission.getRawCodeLineIndexes()).toEqual([0, 1, 2]);
+
+      expect(function() {
+        codeSubmission.replace('this code has too few lines');
+      }).toThrow();
```
eyurko: Should this check for a specific error being thrown? Ditto below. Not certain if that's not a thing in JS.

seanlip:
Thanks! For your last comment, could you explain what the benefits are exactly? Just to help make a more informed decision.

eyurko:
Oh, sure, this would be primarily to allow for session restoration from a transcript. If we save the non-processed code, then we can easily get it back.

seanlip:
Ah OK -- I see what you mean re storing it in the transcript, and that seems reasonable to me. But SolutionHandlerService has access to the CodeSubmission and CodeEvalResult -- so, instead of bundling them together, and making a giant amorphous object containing all the stuff, how about keeping them as separate things that form part of a snapshot?

eyurko:
That seems totally fine as well.

--
Eric Yurko

client/domain/CodeSubmissionObjectFactorySpec.js
seanlip: Per offline discussion, leaving as-is for now.

seanlip:
Per offline discussion, will file a separate issue for the transcript stuff. Merging; thanks @eyurko!

client/services/FeedbackGeneratorService.js
```diff
@@ -54,11 +54,12 @@ tie.factory('FeedbackGeneratorService', [
     };
 
     return {
-      getFeedback: function(prompt, codeEvalResult) {
+      getFeedback: function(
+          prompt, codeEvalResult, rawCodeLineIndexes) {
```
eyurko: I missed this, but you didn't update the spec with rawCodeLineIndexes. Did you verify that this doesn't break any tests?

seanlip: I don't understand, sorry. Why is this part of the spec?

eyurko: Or, rather, I should have noticed this earlier, but I'd like to see some tests (or updates to tests) that verify that this works as intended in a future PR.

eyurko: Sorry, I mean FeedbackGeneratorServiceSpec.js -- all the calls to getFeedback don't use rawCodeLineIndexes.

seanlip: Ah, I see. Yep, this didn't trigger in the tests because that code path wasn't executed, oops. I'll make a note to fix it in a follow-up. Thanks for catching!

PR #61, opened 2017-03-21
UI: Fix next button prompt and style. Add simple processing animation.

rabidbit:
Not for merge. Needs code fix.

### DIFF
--- a/client/components/LearnerViewDirective.js
+++ b/client/components/LearnerViewDirective.js
@@ -291,9 +291,9 @@ tie.directive('learnerView', [function() {
           border: 1px solid rgb(42, 112, 232);
         }
         .tie-run-button:active {
-          box-shadow: inset 0 1px 2px rgba(0,0,0.3);
           background-color: rgb(42, 112, 232);
           border: 1px solid rgb(32, 100, 200);
+          box-shadow: inset 0 1px 2px rgba(0,0,0.3);
         }
         .tie-step-container-inner {
           display: flex;
--- a/client/components/LearnerViewDirective.js
+++ b/client/components/LearnerViewDirective.js
@@ -36,6 +36,8 @@ tie.directive('learnerView', [function() {
             </div>
           </div>
           <div class="tie-coding-ui">
+<!-- LEFT OFF -->
+<p>hideDotDiv = {{ hideDotDiv }} </p>
             <div class="tie-feedback-window">
               <div class="tie-feedback">
                 <p ng-repeat="paragraph in feedbackParagraphs track by $index"
@@ -44,6 +46,12 @@ tie.directive('learnerView', [function() {
                   <span ng-if="$first">{{feedbackTimestamp}}</span>
                   {{paragraph.getContent()}}
                 </p>
+              </div>
+              <div class="tie-dot-container"
+                   ng-class="{'tie-dot-hidden': hideDotDiv}">
+                <div class="tie-dot tie-dot-1"></div>
+                <div class="tie-dot tie-dot-2"></div>
+                <div class="tie-dot tie-dot-3"></div>
               </div>
             </div>
             <div class="tie-coding-window">
@@ -66,6 +74,7 @@ tie.directive('learnerView', [function() {
                 <div class="tie-next-curtain-container"
                     ng-if="nextButtonIsShown">
                   <div class="tie-next-curtain"></div>
+                  <div class="tie-arrow-highlighter"></div>
                   <div ng-click="showNextPrompt()" class="tie-next-arrow">
                     <span class="tie-next-button-text">Next</span>
                   </div>
@@ -97,6 +106,16 @@ tie.directive('learnerView', [function() {
         }
         .CodeMirror-scroll > .CodeMirror-gutters {
           z-index: 1;
+        }
+        .tie-arrow-highlighter {
+          background-color: white;
+          border-radius: 100px;
+          box-shadow: 0px 0px 42px 67px white;
+          height: 50px;
+          left: calc(50% - 25px);
+          position: absolute;
+          top: calc(50% - 25px);
+          width: 50px;
         }
         .tie-coding-terminal .CodeMirror {
           /* Overwriting codemirror defaults */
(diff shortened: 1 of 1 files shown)


client/components/LearnerViewDirective.js
```diff
@@ -411,6 +470,8 @@ tie.directive('learnerView', [function() {
         };
 
         $scope.submitCode = function(code) {
+// LEFT OFF
+$scope.hideDotDiv = false;
           SolutionHandlerService.processSolutionAsync(
```
seanlip: I think it's due to this: http://stackoverflow.com/questions/38507733/angular-promise-hangs-ui

Basically, wrap SolutionHandlerService.processSolutionAsync in a $timeout with 0 delay so the execution of the $scope.hideDotDiv can happen first. I think things are hanging on the promise.

client/components/LearnerViewDirective.js
```diff
@@ -45,6 +45,12 @@ tie.directive('learnerView', [function() {
                   {{paragraph.getContent()}}
                 </p>
               </div>
+              <div class="tie-dot-container"
+                   ng-class="{'tie-dot-hidden': hideDotDiv}">
```
seanlip: If the only point of this is to hide/show, then how about using ng-if="loadingIndicatorIsShown" instead? Using ng-if seems simpler since it's already built-in. The reason for the name change is twofold:

- the "concept" behind the doc is a loading indicator. Its visual appearance may change in the future.
- hideDotDiv starts with a verb, and therefore sounds like a method/function.

client/components/LearnerViewDirective.js
```diff
@@ -410,11 +469,21 @@ tie.directive('learnerView', [function() {
...
           SolutionHandlerService.processSolutionAsync(
-            prompts[currentPromptIndex], code,
-            question.getAuxiliaryCode(language), language
-          ).then(setFeedback);
+              prompts[currentPromptIndex], code,
+              question.getAuxiliaryCode(language), language
+              ).then(setFeedback);
+        };
+
+        $scope.submitCode = function(code) {
```
seanlip: Not sure you need to break the inner thing out into a separate function -- it just adds indirection. Could you just do something like this?

        $scope.submitCode = function(code) {
          $scope.hideDotDiv = false;
          feedbackDiv.scrollTop = feedbackDiv.scrollHeight + 17;
          $timeout(function() {
            SolutionHandlerService.processSolutionAsync(
              prompts[currentPromptIndex], code,
              question.getAuxiliaryCode(language), language
            ).then(setFeedback);
          }, 0);
        };

client/components/LearnerViewDirective.js
```diff
@@ -410,11 +469,21 @@ tie.directive('learnerView', [function() {
...
+          $scope.hideDotDiv = false;
+          $timeout(
+              function(){feedbackDiv.scrollTop = feedbackDiv.scrollHeight + 17},
+              0);
+          var serviceCall = callSolutionHandlerService.bind({'code': code});
+          $timeout(serviceCall, 1000);
```
seanlip: What is the rationale for delaying the call for 1000ms? JS is single-threaded so the main point of the $timeout is just to force the angular $digest to happen (and the $scope values to be updated) before the code starts being processed. But I don't think we need an actual additional delay (at least from an infrastructure perspective), and 0 should work fine.

client/components/LearnerViewDirective.js
```diff
@@ -410,11 +469,21 @@ tie.directive('learnerView', [function() {
...
-          ).then(setFeedback);
+              prompts[currentPromptIndex], code,
+              question.getAuxiliaryCode(language), language
+              ).then(setFeedback);
+        };
+
+        $scope.submitCode = function(code) {
+          $scope.hideDotDiv = false;
+          $timeout(
+              function(){feedbackDiv.scrollTop = feedbackDiv.scrollHeight + 17},
```
seanlip: What is the significance of 17? Should probably be a named constant.

client/components/LearnerViewDirective.js
rabidbit: The reason was in case we add additional css animation. Since we don't have any right now, I changed it to ng-if, as suggested. I changed hideDotDiv to loadingIndicatorIsShown, as suggested.

client/components/LearnerViewDirective.js
rabidbit: Hmm... This is what I had originally, but I ran into some bug earlier that made me define a separate function and didn't revisit it. Thanks for the catch!

client/components/LearnerViewDirective.js

client/components/LearnerViewDirective.js
rabidbit: Ah yes... as it's written, a minimum 15-20ms (at least on my laptop) is needed for the scrolling to work (and 0ms for the scrollTop change). I don't like having any timeouts > 0 either, but.. any suggestions?

seanlip: Oh, I see, that's odd, but I think it's because some DOM work is going on here.

I think 15-20 is fine -- that's much less noticeable than 1000ms. You could perhaps do something like this:

```
        $scope.submitCode = function(code) {
          $scope.hideDotDiv = false;
          $timeout(function() {
            feedbackDiv.scrollTop = feedbackDiv.scrollHeight + 17;
            $timeout(function() {
              SolutionHandlerService.processSolutionAsync(
                prompts[currentPromptIndex], code,
                question.getAuxiliaryCode(language), language
              ).then(setFeedback);
            }, 10);
          }, 0);
        };
```

client/components/LearnerViewDirective.js
```diff
@@ -469,21 +465,17 @@ tie.directive('learnerView', [function() {
...
         $scope.submitCode = function(code) {
-          $scope.hideDotDiv = false;
-          $timeout(
-              function(){feedbackDiv.scrollTop = feedbackDiv.scrollHeight + 17},
-              0);
-          var serviceCall = callSolutionHandlerService.bind({'code': code});
-          $timeout(serviceCall, 1000);
+          $scope.loadingIndicatorIsShown = true;
+          var additionalHeightForLoadingIndicator = 17;
+          $timeout(function(){
```
seanlip: Nit: add space before {

client/components/LearnerViewDirective.js
```diff
@@ -469,21 +465,17 @@ tie.directive('learnerView', [function() {
...
+              feedbackDiv.scrollTop = feedbackDiv.scrollHeight +
```
seanlip: Very minor nit: I think we're indenting by 2 throughout the codebase, not 4. (Sorry, this will be automatically handled once we get a linter in, but thought it worth mentioning so that the pattern doesn't propagate.)

client/components/LearnerViewDirective.js

client/components/LearnerViewDirective.js
```diff
@@ -469,21 +465,17 @@ tie.directive('learnerView', [function() {
...
+              additionalHeightForLoadingIndicator;
+              $timeout(function() {SolutionHandlerService.processSolutionAsync(
```
seanlip: Start "SolutionHandlerService" on a new line.

client/components/LearnerViewDirective.js
```diff
@@ -469,21 +465,17 @@ tie.directive('learnerView', [function() {
...
+                  prompts[currentPromptIndex], code,
+                  question.getAuxiliaryCode(language), language
+                  ).then(setFeedback);}, 20);
```
seanlip: Add line break before }

rabidbit:
@seanlip PTAL

PR #71, opened 2017-03-27
Merge two tasks in i18n into a single task

apicardgoog:


client/services/FeedbackGeneratorService.js
```diff
@@ -125,6 +125,7 @@ tie.factory('FeedbackGeneratorService', [
       feedback.appendTextParagraph(
         "Looks like your code had a runtime error" + inputClause +
         ". Here's the trace:")
+      debugger;
```
seanlip: ?

PR #78, opened 2017-03-30
Add code coverage indicator.

seanlip:


seanlip:
(NB: This PR adds a coverage indicator to show us how much code is covered by tests, so that we can see areas which need more testing.)

PR #79, opened 2017-03-31
Add validation tests for questions.

seanlip:
@eyurko -- pretty sure I've missed some additional stuff you mentioned, but my thinking was, let's just get this in and then we'll have a baseline.

### DIFF
--- a/assets/tests/QuestionSchemaValidationService.js
+++ b/assets/tests/QuestionSchemaValidationService.js
@@ -17,8 +17,11 @@
  */
 
 tie.factory('QuestionSchemaValidationService', [
-  'ALL_SUPPORTED_LANGUAGES',
-  function(ALL_SUPPORTED_LANGUAGES) {
+  'ALL_SUPPORTED_LANGUAGES', 'CLASS_NAME_AUXILIARY_CODE',
+  'CodeCheckerService',
+  function(
+      ALL_SUPPORTED_LANGUAGES, CLASS_NAME_AUXILIARY_CODE,
+      CodeCheckerService) {
     return {
       verifyTitleIsString: function(question) {
         return angular.isString(question.getTitle());
@@ -52,6 +55,30 @@ tie.factory('QuestionSchemaValidationService', [
         return (
           angular.isArray(question.getTasks()) &&
           question.getTasks().length > 0);
+      },
+      verifyStyleTestsAreArray: function(question) {
+        return angular.isArray(question.getStyleTests());
+      },
+      // TODO(sll): Implement verifyEvaluationFunctionNameAppearsInAuxiliaryCode
+      // for style tests.
+      verifyStyleTestsEvaluationFunctionNameIsString: function(question) {
+        return question.getStyleTests().every(function(test) {
+          return (
+            angular.isString(test.getEvaluationFunctionName()) &&
+            test.getEvaluationFunctionName().length > 0);
+        });
+      },
+      verifyExpectedOutputIsNotUndefined: function(question) {
+        return question.getStyleTests().every(function(test) {
+          return (test.getExpectedOutput() !== undefined);
+        });
+      },
+      verifyStyleTestMessageIsNonemptyString: function(question) {
+        return question.getStyleTests().every(function(test) {
+          return (
+            angular.isString(test.getMessage()) &&
+            test.getMessage().length > 0);
+        });
       }
     };
   }
--- a/assets/tests/QuestionSchemaValidationService.js
+++ b/assets/tests/QuestionSchemaValidationService.js
@@ -1,0 +1,58 @@
+// Copyright 2017 The TIE Authors. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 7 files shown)


assets/tests/QuestionSchemaValidationServiceSpec.js
```diff
@@ -0,0 +1,74 @@
...
+            verifierFunctionName];
+          expect(verifierFunction(question)).toBe(true, [
+            verifierFunctionName,
+            ' failed for the question: "' + title + '"'
+          ].join(''));
+        });
+
+        expect(functionCount).toEqual(EXPECTED_VERIFIER_FUNCTION_COUNT, [
+          'Only ' + functionCount + ' functions were called for ',
+          question.getTitle()
```
eyurko: Any reason you use .getTitle() here rather than just title?

assets/tests/TaskSchemaValidationServiceSpec.js
```diff
@@ -18,25 +18,28 @@
...
   beforeEach(inject(function($injector) {
     globalData.questionSets.all = {};
     globalData.questionSets.all.questionIds = QUESTION_IDS;
     globalData.questionSets.all.introductionParagraphs = [];
+
     // Used for testing the validator. Values will be inserted during the tests
     // so that we don't have to redefine the dict every time.
     QuestionDataService = $injector.get('QuestionDataService');
-    Tsvs = $injector.get('TaskSchemaValidationService');
+    TaskSchemaValidationService = $injector.get('TaskSchemaValidationService');
```
eyurko: Oh, right. Good call.

eyurko:
Couple linter errors, but other than that + my one comment it seems reasonable. I'll do a second pass to see if we need more tests and we can just add those as TODOs.

assets/tests/QuestionSchemaValidationService.js
```diff
@@ -0,0 +1,85 @@
...
+ * @fileoverview Service for validating non-task data put into a question.
+ */
+
+tie.factory('QuestionSchemaValidationService', [
+  'ALL_SUPPORTED_LANGUAGES', 'CLASS_NAME_AUXILIARY_CODE',
+  'CodeCheckerService',
+  function(
+      ALL_SUPPORTED_LANGUAGES, CLASS_NAME_AUXILIARY_CODE,
+      CodeCheckerService) {
+    return {
```
eyurko: Need tests to verify that the BuggyOutputTest messages are unique across all tasks.
Should verify that AuxiliaryCode contains the AuxiliaryCode class definition.
Should probably verify that starter code contains Python function definitions (and we have the regex to validate ... somewhere).
It's probably not a BAD idea to verify that the starter code matches a language pattern for the language specified, but that seems hard so that seems like a "later" thing.

Other than that, this should be fine.

assets/tests/QuestionSchemaValidationServiceSpec.js
seanlip: Whoops, oversight. Fixed, thanks!

assets/tests/QuestionSchemaValidationService.js
seanlip: Added TODOs. Thanks!

seanlip:
PTAL -- everything is now fixed, I think. Sorry about the lint errors!

assets/tests/QuestionSchemaValidationServiceSpec.js
```diff
@@ -0,0 +1,73 @@
...
+
+    QuestionDataService.initCurrentQuestionSet('all');
+    questions = QUESTION_IDS.map(function(questionId) {
+      return QuestionDataService.getQuestion(questionId);
+    });
+  }));
+
+  describe('validateAllQuestions', function() {
+    it('should validate the structure of all sample questions', function() {
+      var functions = Object.keys(QuestionSchemaValidationService);
```
eyurko: It seems like it might be worth consolidating these into SchemaValidationService and then have Question and Task inherit from that in some way.

At best, that's a TODO, though.

client/app.js
```diff
@@ -30,6 +30,7 @@ tie.constant('CODE_EXECUTION_TIMEOUT_SECONDS', 3);
 
 // Supported languages.
 tie.constant('LANGUAGE_PYTHON', 'python');
+tie.constant('ALL_SUPPORTED_LANGUAGES', ['python']);
```
eyurko: Should this be [LANGUAGE_PYTHON], or is that not how it works?

seanlip: Nope -- I felt sad about this one too. Need to hardcode, otherwise it becomes too complicated for a constant. Can probably find a better way when we actually support more languages, and all the constants are in the same location anyway so I was comfortable punting it (since anyone adding a new language is going to see this).

assets/tests/QuestionSchemaValidationServiceSpec.js
seanlip: I thought about that, but I think that's a level of indirection we don't really need yet. Let's address it when we have more schemas and a more concrete philosophy for how we handle nested testing, so that it's more obvious what code is actually duplicated across all cases.

(I had a bit of trouble figuring out the best way to combine the Question / Task validation, too, and still keep the function count; I was also thinking about breaking out individual structures like PerformanceTest into its own thing. We can discuss the best way to do this at some point, but it's not urgent.)

seanlip:
Thanks @eyurko  -- treating these as no-ops and submitting, but let me know if you want me to do anything as a follow-up!

PR #114, opened 2017-04-04
fix #88: i18n description confusion

### LINKED ISSUE #88: Description of question i18n not clear
The description/examples of this question could be more detailed, as `dog->d1g` is a valid case according to the description but failed here:

screen shot 2017-04-03 at 1 43 08 pm

Maybe like below:
An abbreviation of a word follows the form `<first character><length of the middle of the string><last character>`. Below are some examples of word abbreviations:
a) it                   --> it    (no abbreviation)
b) d|o|g            --> dog (no abbreviation)
c) i|nternationalizatio|n  --> i18n

jianjinjian:
Modified the original question description from 
" _returns an abbreviation of the string of the form `<first character><length of the middle of the string><last character>`._
 to
" _returns an abbreviation of the string of the form `<first character><length of the middle of the string><last character>` if it shortens the input._ "

### DIFF
--- a/assets/questions/i18n.js
+++ b/assets/questions/i18n.js
@@ -55,7 +55,8 @@ def are_all_unique(words):
       [
         'First, implement the abbreviate function. It takes a string as input ',
         'and returns an abbreviation of the string of the form ',
-        '<first character><length of the middle of the string><last character>. ',
+        '<first character><length of the middle of the string><last character> ',
+        'but only if it shortens the input. ',
         'For example, "internationalization" should be abbreviated as "i18n".'
       ].join('')
     ],


assets/questions/i18n.js
```diff
@@ -55,7 +55,8 @@ def are_all_unique(words):
       [
         'First, implement the abbreviate function. It takes a string as input ',
         'and returns an abbreviation of the string of the form ',
-        '<first character><length of the middle of the string><last character>. ',
+        '<first character><length of the middle of the string><last character> ',
+        'if it shortens the input. ',
```
seanlip: What do you think about saying: "but only if it shortens the input"?

jianjinjian: Yeah, sounds good, thx for the suggestion, this was added

Issue #108, opened 2017-04-04
Horizontal scrolling in editor leads to UI issue


abinash-behera:
**Reproduction Steps:**

1. Open the app.html file in chrome
2. Go to Sample Question 1 (Any sample question would do)
3. Add a long line of code that extends beyond the visible area
4. Scroll to the extreme right using the horizontal scroll bar

**Expected Output:** The line numbers should not overlap with the code 

**Actual Output:** The code contents overlap with the line numbers on the left
![screenshot from 2017-04-04 11 44 53](https://cloud.githubusercontent.com/assets/15383817/24673172/33f80e7e-192c-11e7-9344-78fda2fbb107.png)

eyurko:
Might be worth playing around with some of the information here: https://codemirror.net/doc/manual.html

abinash-behera:
Fixed by PR [132](https://github.com/google/tie/pull/132).

PR #115, opened 2017-04-04
Fix the ui issue caused by horizontal scrolling in editor. (#108)

luofei11:
Now the code doesn't overlap with line number. However, the line number will be hidden if the line is too long. I'm not sure if this introduces a new issue.

screen shot 2017-04-04 at 2 45 07 pm


seanlip:
I think @abinash-behera might have a better solution to this; /cc-ing him here.

PR #142, opened 2017-04-05
Added three extra validation tests, two to check buggyOutputTests, and one to check uniqueness of titles.

eyurko:
Necessary.

### DIFF
--- a/assets/tests/QuestionSchemaValidationService.js
+++ b/assets/tests/QuestionSchemaValidationService.js
@@ -60,6 +60,23 @@ tie.factory('QuestionSchemaValidationService', [
           angular.isArray(question.getTasks()) &&
           question.getTasks().length > 0);
       },
+      verifyAtLeastOneBuggyOutputTestExists: function(question) {
+        return question.getTasks().some(function(task) {
+          return task.getBuggyOutputTests().length > 0;
+        });
+      },
+      verifyAllBuggyOutputTestMessagesAreUnique: function(question) {
+        var messages = new Set();
+        return question.getTasks().every(function(task) {
+          return task.getBuggyOutputTests().every(function(test) {
+            return test.getMessages().every(function(message) {
+              var isUnique = !(message in messages);
+              messages.add(message);
+              return isUnique;
+            });
+          });
+        });
+      },
       verifyStyleTestsAreArray: function(question) {
         return angular.isArray(question.getStyleTests());
       },
--- a/assets/tests/QuestionSchemaValidationService.js
+++ b/assets/tests/QuestionSchemaValidationService.js
@@ -65,18 +65,6 @@ tie.factory('QuestionSchemaValidationService', [
           return task.getBuggyOutputTests().length > 0;
         });
       },
-      verifyAllBuggyOutputTestMessagesAreUnique: function(question) {
-        var messages = new Set();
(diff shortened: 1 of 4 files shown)


assets/tests/QuestionSchemaValidationService.js
```diff
@@ -60,6 +60,23 @@ tie.factory('QuestionSchemaValidationService', [
...
       },
+      verifyAtLeastOneBuggyOutputTestExists: function(question) {
+        return question.getTasks().some(function(task) {
+          return task.getBuggyOutputTests().length > 0;
+        });
+      },
+      verifyAllBuggyOutputTestMessagesAreUnique: function(question) {
+        var messages = new Set();
+        return question.getTasks().every(function(task) {
+          return task.getBuggyOutputTests().every(function(test) {
```
seanlip: Shouldn't the `messages` set just be within the context of a single task? If so, it might also be simpler to just dump all the messages into a set and verify that the set's size equals the length of the list of messages.

assets/tests/QuestionSchemaValidationService.js
```diff
@@ -60,6 +60,23 @@ tie.factory('QuestionSchemaValidationService', [
           angular.isArray(question.getTasks()) &&
           question.getTasks().length > 0);
       },
+      verifyAtLeastOneBuggyOutputTestExists: function(question) {
```
seanlip: I'm not 100% sure that this is an absolute requirement, but happy for you to make the call.

assets/tests/QuestionSchemaValidationServiceSpec.js
```diff
@@ -47,6 +47,18 @@ describe('QuestionSchemaValidationService', function() {
...
 
+  describe('validateTitlesAreUnique', function() {
+    it('should verify that all questions have a unique title', function() {
+      var titles = new Set();
+      expect(questions.every(function(question) {
+        var title = question.getTitle();
+        var isUnique = !(title in titles);
+        titles.add(title);
+        return isUnique;
+      })).toBe(true);
```
seanlip: I mildly think it's simpler/cleaner to dump all titles into a set and verify the set size matches the list length, but happy to defer to you on this.

assets/tests/QuestionSchemaValidationService.js
eyurko: uh, I'm not 100% sure. I think you're correct, though, so I'll fix it.

assets/tests/QuestionSchemaValidationServiceSpec.js
eyurko: This way IS slightly more fun.

assets/tests/QuestionSchemaValidationService.js
eyurko: As discussed, we can leave it until we find a counterexample.

assets/tests/QuestionSchemaValidationServiceSpec.js
seanlip: Well, yeah, but then every reader does have to do a bit of thinking to make sure it works right, and I kinda think you want to avoid that where possible :)

eyurko: fine fine fine fine fine :P

Issue #92, opened 2017-04-04
Strange text when undo is used in editor


shaman-rajan:
Pressing undo in the editor from the default text makes strange text appear in the editor.

![screen shot 2017-04-03 at 7 52 46 pm](https://cloud.githubusercontent.com/assets/1597618/24639806/2732e388-18a7-11e7-92aa-0f0836504c1b.png)

Repro steps:
1. Open app.html
2. Click on editor and press undo (command-Z used here on a mac)

seanlip:
Fixed in #143. Thanks @shaman-rajan!

PR #143, opened 2017-04-05
Fixed random text appearing when undo from default code

shaman-rajan:
Resolving #92 

Bug on line 62 - look [here](https://github.com/angular-ui/ui-codemirror#options) for reference
The blank line and space characters just before closing the ui-codemirror tag are preserved, and after undo is used they appear in the code instead of a blank editor.

### DIFF
--- a/client/components/LearnerViewDirective.js
+++ b/client/components/LearnerViewDirective.js
@@ -59,10 +59,9 @@ tie.directive('learnerView', [function() {
             <div class="tie-coding-window">
               <div class="tie-lang-terminal">
                 <div class="tie-coding-terminal">
-                  <ui-codemirror ui-codemirror="codeMirrorOptions"
+                  <ui-codemirror ui-codemirror-opts="codeMirrorOptions"
                       ng-model="code"
-                      class="tie-codemirror-container">
-                  </ui-codemirror>
+                      class="tie-codemirror-container"></ui-codemirror>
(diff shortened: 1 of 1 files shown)


client/components/LearnerViewDirective.js
```diff
@@ -59,10 +59,9 @@ tie.directive('learnerView', [function() {
             <div class="tie-coding-window">
               <div class="tie-lang-terminal">
                 <div class="tie-coding-terminal">
-                  <ui-codemirror ui-codemirror="codeMirrorOptions"
+                  <ui-codemirror ui-codemirror-opts="codeMirrorOptions"
                       ng-model="code"
-                      class="tie-codemirror-container">
-                  </ui-codemirror>
+                      class="tie-codemirror-container"></ui-codemirror>
```
eyurko: Why was </ui-codemirror> moved to another line?

shaman-rajan: Take a look at the difference between these 2 images. The first one is with the original version with the closing tag in a new line. The second one is with the change and the tag moved up.
image
image

As mentioned in the comment above, the blank line and space characters just before </ui-codemirror> are preserved in the editor. After undo is used, they appear in the editor.

eyurko: Ah, fun. Okay!

PR #153, opened 2017-04-05
Fix #126: Add starter code/script for e2e test

### LINKED ISSUE #126: Add functionality for e2e tests to ensure that the application does not break.
We should probably add an end-to-end test that loads the page and clicks some of the buttons / submits some code, in order to prevent breakages like the one that was fixed in #80. (Note that even an end-to-end test which just loads the app.html page in a browser and checks that there are no console errors would have helped here.)

Protractor is likely to be a good library for this, and is worth investigating.

luofei11:
Screenshot of passed e2e test:

screen shot 2017-04-05 at 4 49 10 pm

This is just a starter test, which just loads the app.html page and ensures there are no console errors.

Update: Travis check passed.


luofei11:
@seanlip Hi Sean, could you help me figure out why the Travis CI build failed, since you are more experienced with it? The tests run perfectly with no errors on my machine.

seanlip:
Sorry, in meeting. Can you talk to @eyurko?

seanlip:
OK -- firstly, thank you for the PR! It's great that you were able to write this from scratch, and I'm glad that we are bringing it into the codebase.

I just took a look at the Travis error. I don't know what is going wrong, but here are some tips for what you can do:

- You can view the Travis error log by clicking on the "Details" link next to the "some checks were not successful" message: https://travis-ci.org/google/tie/builds/219094165 . 
- Doing so should give you an error message which you can search online for. I managed to find [this one](http://stackoverflow.com/questions/20055593/unable-to-run-protractor-econnrefused-connect-econnrefused), for example.

It's OK to try various attempts at making fixes and pushing them, then waiting for the Travis checks to run and repeating the above process until it does. I know that the debugging cycle is slower (sorry about that), but this is the sort of thing that happens when you're integrating multiple systems.

Hope this helps! Feel free to ask if any of the above is unclear.

luofei11:
Thanks, @seanlip ! Now the Travis check pass.

testing/page-loading-spec.js
```diff
@@ -0,0 +1,28 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+/**
+ * @fileoverview End-to-end tests for the app.
+ */
+
+describe("Loading the homepage", function() {
+  it('should load the homepage without errors', function() {
+  browser.get('/client/app.html');
```
seanlip: This line and the following ones should be indented by 2, and truncated to 80 chars.

Could you also please make sure this file is referenced in scripts/run_linter.sh?

testing/page-loading-spec.js
```diff
@@ -0,0 +1,28 @@
...
+
+  var feedbackParagraphs = element.all(by.repeater('paragraph in feedbackParagraphs track by $index'));
+  expect(feedbackParagraphs.count()).toEqual(3);
+  expect(feedbackParagraphs.get(0).getText()).toEqual('Greetings!');
+
```
seanlip: There should also be an explicit check to ensure that there are no console errors. Here is a [reference](https://github.com/oppia/oppia/blob/develop/core/tests/protractor_utils/general.js#L44) that might be useful for this.

Also, please make sure that you try to induce a console error, and that this test detects it and actually fails. It's fine to push a commit that does induce such an error to check whether it shows up as a Travis failure as expected; once you've confirmed that, you can push a second commit that removes the cause of the failure.

testing/page-loading-spec.js
```diff
@@ -0,0 +1,28 @@
+// Copyright 2017 The TIE Authors. All Rights Reserved.
```
seanlip: Let's put this in a folder callled tests/ and call the file allPages.js. I'm not entirely sure what our naming convention going forward would be but I think we might have e2e tests describing different facets of the site, so the directory might contain allPages.js, learnerView.js, etc., with allPages.js just checking whether each page loads OK.

protractor.conf.js
```diff
@@ -0,0 +1,17 @@
+// Protractor configuration for TIE.
+
+exports.config = {
+  framework: 'jasmine',
+  capabilities: {
+    browserName: 'chrome'
+  },
+  specs: ['testing/page-loading-spec.js'],
```
seanlip: Maybe `['tests/*.js']` (to make sure that future files also get run)?

testing/page-loading-spec.js
luofei11: "allPages.js" or "allPages-spec.js"?

seanlip: I would say allPages.js; anything in tests/ is an e2e test. I would leave *Spec.js for the karma unit tests.

testing/page-loading-spec.js
luofei11: ESlint shows some undefined errors(e.g., "browser", "element", etc.), but those modules have already been imported. I saw Oppia also uses ESlint. Any thought?

seanlip: This might help: http://eslint.org/docs/user-guide/configuring#specifying-environments

There's a "protractor" option.

luofei11: Thx a lot!

testing/page-loading-spec.js
luofei11: Will push another commit soon that induces console error and breaks Travis build.

seanlip:
Thanks, @luofei11 -- great job! I just wanted to say that the code looks great to me, so if the Travis error checking works as expected then it is fine to merge this.

luofei11:
@seanlip I think it works as expected. Thanks!

seanlip:
Awesome. Thank you very much @luofei11! It's great to have these tests.

PR #161, opened 2017-04-06
Add schema q15

skywalker-lili:
Add the Q15: sort scrambled itinerary. In abstraction, it's a path finding problem in graph. It contains 3 levels of difficulties:

lvl1: graph doesn't contain loop and there is a unique path to connect all nodes. Can be solved using basic topological sort or union-find.

lvl2: graph contains loop and there is unique path. Topological sort and union-find doesn't work. Need to use dynamic programming with backtracking to brutal force. NP problem but the test data is small so that's fine.

lvl3: graph contains loop and there are two paths to connect all nodes. Learner needs to read the context to choose which path to return. Still use DP and backtracking but need to impose an alphabetic order when doing DP.

### DIFF
--- a/assets/question_sets/strings.js
+++ b/assets/question_sets/strings.js
@@ -26,5 +26,5 @@ globalData.questionSets['strings'] = {  // eslint-disable-line dot-notation
       "I will provide you with feedback."
     ].join('')
   ],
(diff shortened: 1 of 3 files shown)


skywalker-lili:
Didn't read the email this morning. Need to improve.

PR #165, opened 2017-04-06
Added Q5 Find best meeting point in grid-like city

cwcai633:
Added Q5 "find best meeting point in grid-like city".

First, the candidate is asked to give any one answer from all the correct answers.

Second, the candidate is asked to give all possible answers.


eyurko:
Please run the linter before submitting a PR.

Issue #96, opened 2017-04-04
[Question Schema] No Checks on Previous Tasks


xchenan:
The Judge does not test my code for tasks that I've passed before. 

For example, on Question 2, my code went through Task One smoothly but had bugs on Task Two. Then I made some changes so as to pass the tests for Task Two. Even through now my code failed on the tests of Task One, the Online Judge would not find out, but instead moved me forward to the third task. Such design seemed quite inappropriate to me, because the next round task, based on the previous one, encourages challengers to generalize their codes, rather than allowing them to pass the tasks individually.

eyurko:
Thanks -- could you provide repro steps from a clean copy and a screenshot of the error in progress? We're aware of this issue and working to address it, but it's good to have it captured in the tracker.

xchenan:
Repo steps:

1. Open "app.html".

2. Navigate to Question 2, which has three tasks in total.

3. Do code as follows then click "Run", which passes all tests for Task One:
![image](https://cloud.githubusercontent.com/assets/6829866/24670444/e3185062-1922-11e7-930a-44a1f9781fe3.png)

4. The Judge now leads me to the second task.
![image](https://cloud.githubusercontent.com/assets/6829866/24670519/3908e036-1923-11e7-870e-4a63c4af5c85.png)

5. Likewise, do code as follows then click "Run", which should pass Task Two but not Task One:
![image](https://cloud.githubusercontent.com/assets/6829866/24670592/7eee1a4e-1923-11e7-9ee9-aaa2d00c997c.png)

6. However, the Judge still leads to the next task, instead of reporting test failures on Task One:
![image](https://cloud.githubusercontent.com/assets/6829866/24670638/aed1bcac-1923-11e7-99fa-102713e85fab.png)

Expected behavior:
The Judge shall report test failure on tests of all current and previous tasks, instead of merely the current.

seanlip:
Fixed in #173. Thanks!

PR #173, opened 2017-04-06
Fixed issue #96: No checks on previous tasks

xchenan:
Issue reported at #96

Run all test cases for previous and current tasks through recursion. The code passes current task only if it passes all previous and current test cases.


spwahaha:
@eyurko @seanlip  Would you please review this PR?

.gitignore
```diff
@@ -3,3 +3,4 @@ tools/*
 .DS_Store
 assets/.DS_Store
 karma_coverage_reports/*
+assets/questions/*.js~
```
seanlip: Instead of this, find a way to ignore all files with a trailing ~.

client/components/LearnerViewDirective.js
```diff
@@ -477,10 +477,22 @@ tie.directive('learnerView', [function() {
             feedbackDiv.scrollTop = feedbackDiv.scrollHeight +
               additionalHeightForLoadingIndicator;
             $timeout(function() {
-              SolutionHandlerService.processSolutionAsync(
-                tasks[currentTaskIndex], code,
+              var index = 0;
+              var fn = function() {
+                SolutionHandlerService.processSolutionAsync(
+                tasks[index], code,
```
seanlip: Indent this and the following line by 2.

client/components/LearnerViewDirective.js
```diff
@@ -477,10 +477,22 @@ tie.directive('learnerView', [function() {
             feedbackDiv.scrollTop = feedbackDiv.scrollHeight +
               additionalHeightForLoadingIndicator;
             $timeout(function() {
-              SolutionHandlerService.processSolutionAsync(
-                tasks[currentTaskIndex], code,
+              var index = 0;
```
seanlip: Interesting -- so you're using sequential execution instead of assembling all the test code and doing a single exec run.

I wonder how the speed of this looks like in practice and how well it scales. If you had a question with 3-4 tasks, how much of a delay do you see in practice when running this on the last task in that question?

client/components/LearnerViewDirective.js
```diff
@@ -477,10 +477,22 @@ tie.directive('learnerView', [function() {
...
+              var fn = function() {
+                SolutionHandlerService.processSolutionAsync(
+                tasks[index], code,
                 question.getAuxiliaryCode(language), language
-                ).then(setFeedback);
+                ).then(function(feedback) {
+                  if (index === currentTaskIndex ||
+                      !feedback.isAnswerCorrect()) {
+                    setFeedback(feedback);
```
seanlip: The functionality of providing feedback to an earlier part of the question is new, so I wanted to check whether there's the potential for confusion if the user is solving task 3 and gets feedback for task 2. Can you post some screenshots showing what such a flow would look like?

client/components/LearnerViewDirective.js
xchenan: Do you mean we have to run the tests of different tasks simultaneously? What do you mean by assembling?

Now, with sequential execution, the delay is rather subtle, since existing test cases are simple, plus the fact that total runtime is just the sum of the time for running test cases of first task to current.

client/components/LearnerViewDirective.js
xchenan: 1. The following code fails on the first task:
screenshot from 2017-04-10 15 11 57

2. Now it passes
screenshot from 2017-04-10 15 10 25

3. Such code passes the second task, but fails on the first one. Originally the system still approves it, whereas with our code, the feedback from the first test is displayed:
screenshot from 2017-04-10 15 11 23

4. If we write down a code that passes the first task but fails on the second one, the system will notify us again:
screenshot from 2017-04-10 15 10 56

client/components/LearnerViewDirective.js
seanlip: Right. How many secs/millisecs is it in practice? Just wanted to get a feel for that.

Re your first paragraph, yes, that was my thought -- find some way to batch all the test code together and send it to Skulpt for a single exec run. Open to debate on the pros/cons of that alternative vs this one, just want to make sure you've considered it.

client/components/LearnerViewDirective.js
seanlip: That seems reasonable. Thanks!

client/components/LearnerViewDirective.js
spwahaha: We proposed a design for solving this problem
https://docs.google.com/a/google.com/document/d/1HmmorRKmaV6IVqUJYDSBBEYRAzn-MK1y3HLctkOSTJ8/edit?usp=sharing
Please review that.

xchenan: @seanlip Please view our design doc. We are ready to discuss the impl once you've finished reading.

.gitignore
xchenan: Done. Now we ignore all files whose name end with tilde.

xchenan:
@seanlip As we discussed right before we left from work yesterday, now we wrap up all test cases from first task to current, and send them over to the server side in one single request.

seanlip:
(Update: this was discussed offline. Some refactoring is still needed.)

xchenan:
@seanlip We modified the testing logic so that all types of tests (buggy, correctness and performance) of one certain task will be run before the next task. 
Instead of changing the script formatting in codeSubmission (making a 2-D array), we actually changed the logic of comparing answers. Please refer to client/services/FeedbackGeneratorService.js for details.

seanlip:
Discussed offline. Still needs rework so that the code that is sent for async testing is conceptually clear.

xchenan:
@seanlip We modified the data structure that stores the test cases in webpage's python script (1-D arrays to 2-D arrays). Now the test goes as task 1's BuggyOutputTest -> task 1's CorrectnessTest -> task 1's PerformanceTest -> task 2's BuggyOutputTest -> task 2's CorrectnessTest -> task 2's PerformanceTest when the student is running its code on task 2. 

@eyurko Sean told me you would take care of this while he was away. Please review and feel free to give suggestions.

client/components/LearnerViewDirective.js
```diff
@@ -492,10 +492,15 @@ tie.directive('learnerView', [function() {
             feedbackDiv.scrollTop = feedbackDiv.scrollHeight +
               additionalHeightForLoadingIndicator;
             $timeout(function() {
+              // Tasks from the first to current.
+              var tasksToCurrent = [];
```
eyurko: tasksToCurrent is a bit confusing. Maybe orderedTasks or something? What do you think?

client/services/code_preprocessors/PythonCodePreprocessorService.js
```diff
@@ -155,32 +155,50 @@ tie.factory('PythonCodePreprocessorService', [
...
-        );
-        var testRunCode = (
-          'System.runTest(' + qualifiedMainFunctionName + ', '
-          + testInputCode + ')'
-        );
-        var testOutputCode = (
-          outputFunctionName ?
-          outputFunctionName + '(' + testRunCode + ')' : testRunCode);
 
+      var VARNAME_ONE_TASK_CORRECTNESS_TEST_RESULTS = 'one_task_test_results';
```
eyurko: Make this a TIE constant?

client/services/code_preprocessors/PythonCodePreprocessorService.js
```diff
@@ -155,32 +155,50 @@ tie.factory('PythonCodePreprocessorService', [
...
+          var taskTestOutputCode = '[' + testOutputCode +']'
+
+          testCode += (
+            '\n' +
+            VARNAME_ONE_TASK_CORRECTNESS_TEST_RESULTS + '.append(' + testOutputCode + ')');
+        }
+
+          testCode += (
+            '\n' +
+            VARNAME_CORRECTNESS_TEST_RESULTS + '.append(' + VARNAME_ONE_TASK_CORRECTNESS_TEST_RESULTS + ')');
```
eyurko: Line's too long.

client/services/code_preprocessors/PythonCodePreprocessorService.js
```diff
@@ -155,32 +155,50 @@ tie.factory('PythonCodePreprocessorService', [
...
+            + testInputCode + ')'
+          );
+          var testOutputCode = (
+            outputFunctionName ?
+            outputFunctionName + '(' + testRunCode + ')' : testRunCode);
+          var taskTestOutputCode = '[' + testOutputCode +']'
+
+          testCode += (
+            '\n' +
+            VARNAME_ONE_TASK_CORRECTNESS_TEST_RESULTS + '.append(' + testOutputCode + ')');
```
eyurko: Line's too long.

client/services/code_preprocessors/PythonCodePreprocessorService.js
```diff
@@ -205,63 +223,90 @@ tie.factory('PythonCodePreprocessorService', [
...
         '    return buggy_results == ' + VARNAME_CORRECTNESS_TEST_RESULTS,
         '',
         VARNAME_BUGGY_OUTPUT_TEST_RESULTS + ' = []',
         ''
       ].join('\n');
-      buggyOutputTests.forEach(function(buggyOutputTest) {
-        var qualifiedBuggyFunctionName = (
-          buggyOutputTest.getBuggyFunctionName());
+
+      var VARNAME_ONE_TASK_BUGGY_OUTPUT_TEST_RESULTS = 'one_buggy_output_test_results';
```
eyurko: TIE constant

client/services/code_preprocessors/PythonCodePreprocessorService.js
```diff
@@ -205,63 +223,90 @@ tie.factory('PythonCodePreprocessorService', [
...
+          '.append('+ VARNAME_ONE_TASK_BUGGY_OUTPUT_TEST_RESULTS +')\n');
       });
 
       return fullTestCode;
     };
 
     var _generatePerformanceTestCode = function(performanceTests) {
-      var testCode = '';
+      var testCode = (VARNAME_PERFORMANCE_TEST_RESULTS + ' = []\n')
+      var VARNAME_ONE_TASK_PERFORMANCE_TEST_RESULTS = 'one_task_performance_test_results';
```
eyurko: Tie Constant.

client/services/code_preprocessors/PythonCodePreprocessorService.js
```diff
@@ -205,63 +223,90 @@ tie.factory('PythonCodePreprocessorService', [
...
+            'def run_performance_test_from_'+
+                 qualifiedTransformationFunctionName.replace('.', '_') + '_' + index +
+                 '(test_input):',
+            '    time_array = []',
+            '    for input_size in [' + SMALL_INPUT_SIZE + (
+              ', ' + LARGE_INPUT_SIZE + ']:'),
+            '        start = time.time()',
+            '        output = ' + qualifiedEvaluationFunctionName + (
+              '(get_test_input_from_' +
+                qualifiedTransformationFunctionName.replace('.', '_') + '_' + index + '(test_input, input_size))'),
```
eyurko: line's too long.

client/services/code_preprocessors/PythonCodePreprocessorServiceSpec.js
```diff
@@ -437,16 +437,18 @@ describe('PythonCodePreprocessorService', function() {
...
         var expectedGeneratedCode = [
-          'test_inputs = [\'cat\']',
+          'test_inputs = [[\'cat\']]',
           '',
           'correctness_test_results = []',
-          'correctness_test_results.append(outputFnName(' +
-            'System.runTest(StudentCode().mainFnName, inputFnName(test_inputs[0]))))'
+          'one_task_test_results = []',
+          'one_task_test_results.append(outputFnName(' +
+            'System.runTest(StudentCode().mainFnName, inputFnName(test_inputs[0][0]))))',
```
eyurko: Line's too long.

client/services/code_preprocessors/PythonCodePreprocessorServiceSpec.js
```diff
@@ -460,23 +462,28 @@ describe('PythonCodePreprocessorService', function() {
   describe('_generateBuggyOutputTestCode', function() {
     it('should add correct buggy output test code to skeleton code',
       function() {
-        var buggyOutputTests = [BuggyOutputTestObjectFactory.create({
+        var buggyOutputTests = [[BuggyOutputTestObjectFactory.create({
```
eyurko: Might want to pull out the [ ] a bit so that it reads a bit cleaner. This is a nit though.

client/services/code_preprocessors/PythonCodePreprocessorServiceSpec.js
```diff
@@ -572,26 +597,29 @@ describe('PythonCodePreprocessorService', function() {
...
           evaluationFunctionName: 'katamariDamashi'
         });
-        var performanceTests = [performanceTest];
+        var performanceTests = [[performanceTest]];
         var expectedGeneratedCode = [
+          'performance_test_results = []',
+          'one_task_performance_test_results = []',
           '',
-          'def get_test_input(atom, input_size):',
+          'def get_test_input_from_System_extendString_0(atom, input_size):',
```
eyurko: Why is this _0?

client/components/LearnerViewDirective.js
xchenan: Done.

client/services/code_preprocessors/PythonCodePreprocessorService.js
xchenan: Done.

client/services/code_preprocessors/PythonCodePreprocessorService.js
xchenan: Done.

client/services/code_preprocessors/PythonCodePreprocessorService.js
xchenan: Done.

client/services/code_preprocessors/PythonCodePreprocessorService.js
xchenan: Done.

client/services/code_preprocessors/PythonCodePreprocessorService.js
xchenan: Done.

client/services/code_preprocessors/PythonCodePreprocessorServiceSpec.js
xchenan: Previously, we want to avoid re-defining if multiple tasks use the same helper function for performance test. That's why we added taskIndex at the end of helper function name. But now we use a set which stores names of all those helper functions. In this way, code redundancy can be prevented.

client/services/code_preprocessors/PythonCodePreprocessorService.js
xchenan: Done.

client/services/code_preprocessors/PythonCodePreprocessorService.js
xchenan: Done.

xchenan: Fixed.

client/services/code_preprocessors/PythonCodePreprocessorServiceSpec.js
xchenan: Fixed.

client/services/code_preprocessors/PythonCodePreprocessorServiceSpec.js
xchenan: Shifted it to a new line so as to make it look nicer.

client/services/code_preprocessors/PythonCodePreprocessorServiceSpec.js
```diff
@@ -572,26 +599,29 @@ describe('PythonCodePreprocessorService', function() {
...
+          'def get_test_input_from_System_extendString(atom, input_size):',
           '    return System.extendString(atom, input_size)',
           '',
-          'def run_performance_test(test_input):',
+          'def run_performance_test_from_System_extendString(test_input):',
           '    time_array = []',
           '    for input_size in [10, 100]:',
           '        start = time.time()',
-          '        output = StudentCode().katamariDamashi(get_test_input(test_input, input_size))',
+          '        output = StudentCode().katamariDamashi(get_test_input_from_System_extendString(test_input, input_size))',
```
eyurko: This line is really long.

xchenan:
@seanlip @eyurko We added unit test for ./client/service/SolutionHandlerService.js so as to test the logic of student code checks. Please review.

client/services/SolutionHandlerServiceSpec.js
```diff
@@ -0,0 +1,227 @@
...
+      '        return True',
+      '    @classmethod',
+      '    def mockAuxiliaryCodeTwo(cls, input):',
+      '        return False'
+    ].join('\n');
+  });
+
+  describe('processSolutionAsync', function() {
+    describe('correctnessTests', function() {
+      it('should be passed on both task1 and task2', function() {
```
eyurko: should check both task1 and task2 to verify that the learner has the correct answer

client/services/SolutionHandlerServiceSpec.js
```diff
@@ -0,0 +1,227 @@
...
+          SolutionHandlerService.processSolutionAsync(
+            orderedTasks, studentCode,
+            auxiliaryCode, 'python'
+          ).then(function(feedback) {
+            expect(feedback.isAnswerCorrect()).toEqual(true);
+          });
+        }, 0);
+      });
+
+      it('should be passed on task2 and failed on task1', function() {
```
eyurko: similar to previous comment

client/services/SolutionHandlerServiceSpec.js
```diff
@@ -0,0 +1,227 @@
...
+            auxiliaryCode, 'python'
+          ).then(function(feedback) {
+            expect(feedback.isAnswerCorrect()).toEqual(false);
+            expect(feedback.getParagraphs()[1].getContent()).toEqual(
+                "Input: \"testcase_1\"\nOutput: False");
+          });
+        }, 0);
+      });
+
+      it('should be passed on task1 and failed on task2', function() {
```
eyurko: again, similar to previous comment

client/services/SolutionHandlerServiceSpec.js
```diff
@@ -0,0 +1,227 @@
...
+            auxiliaryCode, 'python'
+          ).then(function(feedback) {
+            expect(feedback.isAnswerCorrect()).toEqual(false);
+            expect(feedback.getParagraphs()[1].getContent()).toEqual(
+               "Input: \"testcase_2\"\nOutput: True");
+          });
+        }, 0);
+      });
+
+      it('should be failed on both task1 and task2' +
```
eyurko: ditto

client/services/SolutionHandlerServiceSpec.js
```diff
@@ -0,0 +1,227 @@
...
+            expect(feedback.isAnswerCorrect()).toEqual(false);
+            expect(feedback.getParagraphs()[1].getContent()).toEqual(
+                "Input: \"testcase_0\"\nOutput: False");
+          });
+        }, 0);
+      });
+    });
+
+    describe("buggyOutputTests", function() {
+      it('should be failed on task1', function() {
```
eyurko: ditto

client/services/SolutionHandlerServiceSpec.js
```diff
@@ -0,0 +1,227 @@
...
+            auxiliaryCode, 'python'
+          ).then(function(feedback) {
+            expect(feedback.isAnswerCorrect()).toEqual(false);
+            expect(feedback.getParagraphs()[0].getContent()).toEqual(
+               "Mock BuggyOutputTest Message One for task1");
+          });
+        }, 0);
+      });
+
+      it('should be failed on task2,' +
```
eyurko: ditto

client/services/code_preprocessors/PythonCodePreprocessorService.js
```diff
@@ -207,63 +234,92 @@ tie.factory('PythonCodePreprocessorService', [
       var fullTestCode = [
         'def matches_buggy_function(func):',
         '    buggy_results = []',
-        '    for test_input in test_inputs:',
-        '        buggy_results.append(' + testOutputCode + ')',
+        '    for tests in test_inputs:',
```
eyurko: should this be task in tasks?

client/services/code_preprocessors/PythonCodePreprocessorService.js
```diff
@@ -207,63 +234,92 @@ tie.factory('PythonCodePreprocessorService', [
...
+        '        one_task_results = []',
```
eyurko: then this would be task_results

client/services/code_preprocessors/PythonCodePreprocessorService.js
```diff
@@ -207,63 +234,92 @@ tie.factory('PythonCodePreprocessorService', [
...
+        '        for test in tests:',
```
eyurko: and then this could be test in task

client/services/code_preprocessors/PythonCodePreprocessorServiceSpec.js
```diff
@@ -437,16 +437,18 @@ describe('PythonCodePreprocessorService', function() {
...
         var expectedGeneratedCode = [
-          'test_inputs = [\'cat\']',
+          'test_inputs = [[\'cat\']]',
           '',
           'correctness_test_results = []',
-          'correctness_test_results.append(outputFnName(' +
-            'System.runTest(StudentCode().mainFnName, inputFnName(test_inputs[0]))))'
+          'one_task_correctness_test_results = []',
+          'one_task_correctness_test_results.append(outputFnName(' +
+            'System.runTest(StudentCode().mainFnName, inputFnName(test_inputs[0][0]))))',
```
eyurko: Line's too long. Keep line lengths to 80 characters, if you can.

client/services/code_preprocessors/PythonCodePreprocessorServiceSpec.js
```diff
@@ -460,23 +462,30 @@ describe('PythonCodePreprocessorService', function() {
...
+          '        for test in tests:',
+          '            one_task_results.append(System.runTest(func, test))',
+          '        buggy_results.append(one_task_results)',
           '    return buggy_results == correctness_test_results',
           '',
           'buggy_output_test_results = []',
-          'buggy_output_test_results.append(matches_buggy_function(buggyFunc))',
-          'buggy_output_test_results.append(matches_buggy_function(buggyFunc2))',
+          'one_task_buggy_output_test_results = []',
+          'one_task_buggy_output_test_results.append(matches_buggy_function(buggyFunc))',
```
eyurko: Line's too long.

client/services/code_preprocessors/PythonCodePreprocessorServiceSpec.js
```diff
@@ -460,23 +462,30 @@ describe('PythonCodePreprocessorService', function() {
...
+          '        one_task_results = []',
+          '        for test in tests:',
+          '            one_task_results.append(System.runTest(func, test))',
+          '        buggy_results.append(one_task_results)',
           '    return buggy_results == correctness_test_results',
           '',
           'buggy_output_test_results = []',
-          'buggy_output_test_results.append(matches_buggy_function(buggyFunc))',
-          'buggy_output_test_results.append(matches_buggy_function(buggyFunc2))',
+          'one_task_buggy_output_test_results = []',
```
eyurko: I'd drop the one and just call this task_buggy_ etc.

xchenan: The array stores the test results of **one single task only**, therefore in our opinion it's important to emphasize 'one task' in variable names.

eyurko: I'd be more inclined to call it all_tasks if it were all tasks and just call it "task" for one task, rather than one_task. That's my preference.

xchenan:
@seanlip @eyurko Done with Specs and all tests have been passed. Conflicts have been resolved. Please feel free to review. Thanks a lot.

client/services/SolutionHandlerServiceSpec.js
```diff
@@ -0,0 +1,232 @@
...
+            expect(feedback.getParagraphs()[1].getContent()).toEqual(
+                "Input: \"testcase_0\"\nOutput: False");
+          });
+        }, 0);
+      });
+    });
+
+    describe("buggyOutputTests", function() {
+      it('should check both task1 and task2 to ' +
+          'verify that the leaner fails on task1', function() {
```
xchenan: Fixed

client/services/SolutionHandlerServiceSpec.js
```diff
@@ -0,0 +1,232 @@
...
+          ).then(function(feedback) {
+            expect(feedback.isAnswerCorrect()).toEqual(false);
+            expect(feedback.getParagraphs()[0].getContent()).toEqual(
+               "Mock BuggyOutputTest Message One for task1");
+          });
+        }, 0);
+      });
+
+      it('should check both task1 and task2, ' +
+          'though leaner fails on task2 buggy tests, ' +
```
xchenan: Fixed

seanlip:
Just a quick note -- I know I need to review this, it's on my TODO list. Will try and get to it tomorrow!

client/components/LearnerViewDirective.js
```diff
@@ -629,10 +629,15 @@ tie.directive('learnerView', [function() {
             feedbackDiv.scrollTop = feedbackDiv.scrollHeight +
               additionalHeightForLoadingIndicator;
             $timeout(function() {
+              // Tasks from the first to current.
+              var orderedTasks = [];
```
seanlip: You might want to look at Array.prototype.slice(): https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/slice?v=control

client/services/SolutionHandlerService.js
```diff
@@ -45,19 +45,28 @@ tie.factory('SolutionHandlerService', [
...
+          }
+          var lastTask = tasks[tasks.length - 1];
           var codeSubmission = CodeSubmissionObjectFactory.create(
             studentCode.trim());
           CodePreprocessorDispatcherService.preprocess(
             language, codeSubmission, auxiliaryCode,
-            task.getInputFunctionName(), task.getMainFunctionName(),
-            task.getOutputFunctionName(), task.getCorrectnessTests(),
-            task.getBuggyOutputTests(), task.getPerformanceTests());
+            lastTask.getInputFunctionName(), lastTask.getMainFunctionName(),
```
seanlip: For questions like i18n, different tasks may have different function names. Does your code still work in this case?

client/services/SolutionHandlerServiceSpec.js
```diff
@@ -0,0 +1,232 @@
...
+    var taskDict = [
+      {
+        instructions: [''],
+        prerequisiteSkills: [''],
+        acquiredSkills: [''],
+        inputFunctionName: null,
+        outputFunctionName: null,
+        mainFunctionName: 'mockMainFunction',
+        correctnessTests: [{
+          input: 'testcase_0',
```
seanlip: It might be better to name this 'task_1_correctness_test_1' or similar, and check explicitly for the substring 'task_1' in the student code. That would make it easier to read the test cases -- the idea here is to make obvious, to someone who reads only the student code, which test cases are expected to fail. Otherwise lots of careful cross-referencing ends up being needed.

Ditto for the other input strings.

client/services/code_preprocessors/PythonCodePreprocessorService.js
```diff
@@ -156,32 +164,51 @@ tie.factory('PythonCodePreprocessorService', [
...
+            VARNAME_TEST_INPUTS + '[' + i + '][' + j + ']'
+          );
+          var testRunCode = (
+            'System.runTest(' + qualifiedMainFunctionName + ', '
+            + testInputCode + ')'
+          );
+          var testOutputCode = (
+            outputFunctionName ?
+            outputFunctionName + '(' + testRunCode + ')' : testRunCode);
+          var taskTestOutputCode = '[' + testOutputCode +']'
```
seanlip: Where is this variable used?

client/services/code_preprocessors/PythonCodePreprocessorService.js
```diff
@@ -156,32 +164,51 @@ tie.factory('PythonCodePreprocessorService', [
...
+
+          testCode += (
+            '\n' +
+            VARNAME_TASK_CORRECTNESS_TEST_RESULTS +
+            '.append(' + testOutputCode + ')');
+        }
+
+          testCode += (
```
seanlip: Deindent this and the following 3 lines by 2.

client/services/code_preprocessors/PythonCodePreprocessorService.js
```diff
@@ -20,14 +20,22 @@
...
-      VARNAME_PERFORMANCE_TEST_RESULTS, VARNAME_MOST_RECENT_INPUT) {
+      VARNAME_PERFORMANCE_TEST_RESULTS,
+      VARNAME_TASK_CORRECTNESS_TEST_RESULTS,
+      VARNAME_TASK_BUGGY_OUTPUT_TEST_RESULTS,
+      VARNAME_TASK_PERFORMANCE_TEST_RESULTS,
+      VARNAME_MOST_RECENT_INPUT) {
     var PYTHON_FUNCTION_DEF_REGEX = new RegExp(
       'def\\s+([A-Za-z_][A-Za-z_0-9]*)\\s*\\(', 'g');
-    var VARNAME_TEST_INPUTS = 'test_inputs';
+    var VARNAME_TEST_INPUTS = 'all_tasks_test_inputs';
```
seanlip: VARNAME_ALL_TASKS_TEST_INPUTS perhaps

client/services/code_preprocessors/PythonCodePreprocessorService.js
```diff
@@ -206,63 +233,92 @@ tie.factory('PythonCodePreprocessorService', [
...
         '    return buggy_results == ' + VARNAME_CORRECTNESS_TEST_RESULTS,
         '',
         VARNAME_BUGGY_OUTPUT_TEST_RESULTS + ' = []',
         ''
       ].join('\n');
-      buggyOutputTests.forEach(function(buggyOutputTest) {
-        var qualifiedBuggyFunctionName = (
-          buggyOutputTest.getBuggyFunctionName());
+
+      buggyOutputTests.forEach(function(oneTaskBuggyOutputTest) {
```
seanlip: Perhaps this ought to be: `allTasksBuggyOutputTests.forEach(function(oneTaskBuggyOutputTests) {`. Note the pluralization of Tests at the end.

client/services/code_preprocessors/PythonCodePreprocessorService.js
```diff
@@ -206,63 +233,92 @@ tie.factory('PythonCodePreprocessorService', [
...
-          '        finish = time.time() - start',
-          '        time_array.append(finish)',
-          '    if time_array[1] > ' + UPPER_BOUND_RATIO_IF_LINEAR + (
-            ' * time_array[0]:'),
-          '        return "not linear"',
-          '    return "linear"',
-          '',
-          VARNAME_PERFORMANCE_TEST_RESULTS + ' = []'
-        ].join('\n');
+      var testCode = (VARNAME_PERFORMANCE_TEST_RESULTS + ' = []\n')
```
seanlip: Missing semicolon.

client/services/SolutionHandlerService.js
xchenan: It still works. Our impl just uses the code that the student submits for the current task and test it on all test cases.

client/services/SolutionHandlerServiceSpec.js
xchenan: Testcase inputs changed.

client/services/code_preprocessors/PythonCodePreprocessorService.js
xchenan: Done.

client/services/code_preprocessors/PythonCodePreprocessorService.js
xchenan: Deleted. We don't think it shall be surrounded by brackets. That was a mistake.

client/services/code_preprocessors/PythonCodePreprocessorService.js
xchenan: Done.

client/services/code_preprocessors/PythonCodePreprocessorService.js
xchenan: Done. Also done with allTasksCorrectnessTests and allTasksPerformanceTests.

client/services/code_preprocessors/PythonCodePreprocessorService.js
xchenan: Added.

client/services/SolutionHandlerService.js
seanlip: Oh, I see. But then what is the point of passing in the mainFunctionName at all?

To be clear about what I'm asking, the mainFunctionName for https://github.com/google/tie/blob/master/assets/questions/internationalization.js#L76 starts off as 'abbreviate' but then changes to 'are_all_unique'. In your implementation, I believe there is supposed to be nothing special about the last task. So either the mainFunctionName etc. matters (in which case you'll need the data from all the tasks) or it doesn't matter (in which case you wouldn't need to pass it in at all). Am I missing something here?

seanlip:
Thanks! Just one residual comment. Also could you please fix the merge conflicts?

client/services/SolutionHandlerService.js
xchenan: I am not sure if I understand your question.

So a python script is generated for running. Inside the script the function that student implements should be invoked while testing. Hence the function name is a must. Even through the function name is changed from task to task, we just care about the current one that the student is writing and want to test it on test cases of beginning from the first task to the current.

I don't think we are able to invoke the function that the student implements without the name. 

Do I answer your question?

seanlip: In the internationalization question, there are two tasks, which involve the student writing two separate functions. The first task contains one function name, and the second task contains a different function name. (The second task's function will make a call to the first task's function.)

When doing the test run that involves both task 1 and 2, the first function name should still be used when evaluating correctness for the first task. It sounds like you're not doing that?

client/services/code_preprocessors/PythonCodePreprocessorService.js
```diff
@@ -152,117 +160,183 @@ tie.factory('PythonCodePreprocessorService', [
...
-          qualifiedBuggyFunctionName + '))\n');
-      });
+      var matchesBuggyFunctionSet = new Set();
+
+      for (var i = 0; i < allTasksInputFunctionNames.length; i++) {
+        var inputFunctionName = allTasksInputFunctionNames[i];
+        var outputFunctionName = allTasksOutputFunctionNames[i];
+
+        var matchesBuggyFunctionName = 'matches_buggy_function' +
+            (inputFunctionName == null? '' : '_' + inputFunctionName) +
```
seanlip: This is an interesting approach, but what if the inputFunctionName or outputFunctionName contains periods? Did you test with any of the existing files that contain non-null names? I think it's going to break.

That said, another approach that might be better would be to add additional arguments input_func and output_func to a single common matches_buggy_function function. Functions are first-class objects in Python, and can be passed around. I think that that would be simpler and less error-prone.

client/services/code_preprocessors/PythonCodePreprocessorService.js
```diff
@@ -152,117 +160,183 @@ tie.factory('PythonCodePreprocessorService', [
...
+            outputFunctionName + '(' + testRunCode + ')' : testRunCode);
+
+          var fullTestCode = [
+            'def ' + matchesBuggyFunctionName + '(func):',
+            '    buggy_results = []',
+            '    for task_tests in all_tasks_test_inputs:',
+            '        task_results = []',
+            '        for test_input in task_tests:',
+            '            task_results' +
+            '.append(' + testOutputCode + ')',
```
seanlip: Go ahead and put this on the previous line, even if it exceeds the line limit. It is much easier to read that way.

client/services/code_preprocessors/PythonCodePreprocessorService.js
```diff
@@ -152,117 +160,183 @@ tie.factory('PythonCodePreprocessorService', [
...
+            test.getTransformationFunctionName());
+          // TODO(eyurko): Make this work for non-linear runtimes, such as log(n).
+          // TODO(eyurko): Use linear regression to determine if the data points
+          // "look" linear, quadratic, etc, and then provide feedback accordingly.
+          if (!performanceTestFunctionSet.has(qualifiedEvaluationFunctionName)) {
+            performanceTestFunctionSet.add(qualifiedEvaluationFunctionName);
+            testCode += [
+              '',
+              'def get_test_input_from_'+
+                  qualifiedTransformationFunctionName.replace('.', '_') +
```
seanlip: Same thing -- this sort of munging of the function names is a bit hacky. Can we treat the functions as first-class objects and pass them in as args instead?

xchenan:
Note: Almost finish re-writing generateBuggyOutputTestCode and generatePerformanceTestCode. Handling indentation and test right now. 
Hopefully will have an update later today.

xchenan:
@seanlip @eyurko Just finished the changes. Please feel free to review.
Let me know if you have comments. Thanks.

client/services/SolutionHandlerServiceSpec.js
```diff
@@ -0,0 +1,240 @@
...
+      }
+    ];
+
+    orderedTasks = taskDict.map(function(task) {
+      return TaskObjectFactory.create(task);
+    });
+  }));
+
+  // Mock auxiliary code for processing.
+  beforeEach(function() {
```
seanlip: Can you include the stuff in this and the next beforeEach clause into the above one (i.e. after defining orderedTasks in line 88)? We don't need separate beforeEach clauses for each variable assignment.

client/services/code_preprocessors/PythonCodePreprocessorService.js
```diff
@@ -152,117 +160,181 @@ tie.factory('PythonCodePreprocessorService', [
...
-          '.append(matches_buggy_function(' +
-          qualifiedBuggyFunctionName + '))\n');
-      });
+      var matchesBuggyFunctionSet = new Set();
+
+      for (var i = 0; i < allTasksInputFunctionNames.length; i++) {
+        var inputFunctionName = allTasksInputFunctionNames[i];
+        var outputFunctionName = allTasksOutputFunctionNames[i];
+
+          var testInputCode = (
```
seanlip: The indentation here and below seems wrong -- it should be on the same level as the previous line of code.

client/services/code_preprocessors/PythonCodePreprocessorService.js
```diff
@@ -152,117 +160,181 @@ tie.factory('PythonCodePreprocessorService', [
...
+            outputFunctionName ?
+            outputFunctionName + '(' + testRunCode + ')' : testRunCode);
+
+          var fullTestCode = [
+            'def matches_buggy_function(func, inputFunctionName, outputFunctionName):',
+            '    buggy_results = []',
+            '    for task_tests in all_tasks_test_inputs:',
+            '        task_results = []',
+            '        for test_input in task_tests:',
+            '            if inputFunctionName == None and outputFunctionName == None:',
```
seanlip: Use `... is None` rather than `... == None`, here and below.

seanlip:
Also, please could you fix the merge conflict?

seanlip:
OK -- everything LGTM except the merge conflict. Thanks!

seanlip:
NB: there are some test failures.

PR #180, opened 2017-04-07
added stdout

cwcai633:
Write stdout to the page

### DIFF
--- a/client/services/FeedbackGeneratorService.js
+++ b/client/services/FeedbackGeneratorService.js
@@ -99,7 +99,7 @@ tie.factory('FeedbackGeneratorService', [
       feedback.appendCodeParagraph(
         'Input: ' + _jsToHumanReadable(correctnessTest.getInput()) + '\n' +
         'Output: ' + _jsToHumanReadable(observedOutput) + '\n' +
-        'stdout: ' + stdOutput);
+        'Stdout: ' + stdOutput);
       feedback.appendTextParagraph('However, the expected output is:');
       feedback.appendCodeParagraph(
         _jsToHumanReadable(allowedOutputExample));
--- a/client/services/FeedbackGeneratorService.js
+++ b/client/services/FeedbackGeneratorService.js
@@ -187,8 +187,9 @@ tie.factory('FeedbackGeneratorService', [
           var stdOutputs = codeEvalResult.getOutput();
           for (i = 0; i < correctnessTests.length; i++) {
             var observedOutput = observedOutputs[i];
-            var outputCount =  stdOutputs.length /correctnessTests.length;
-            var stdOutput = stdOutputs.slice(i * outputCount, (i + 1) * outputCount);
+            var outputCount = stdOutputs.length / correctnessTests.length;
+            var stdOutput = stdOutputs.slice(i * outputCount,
+                (i + 1) * outputCount);
 
             // TODO(eyurko): Add varied statements for when code is incorrect.
             if (!correctnessTests[i].matchesOutput(observedOutput)) {
--- a/client/services/FeedbackGeneratorService.js
+++ b/client/services/FeedbackGeneratorService.js
@@ -92,14 +92,14 @@ tie.factory('FeedbackGeneratorService', [
     };
 
     var _getCorrectnessTestFeedback = function(
-      correctnessTest, observedOutput, stdOutput) {
+      correctnessTest, observedOutput, stdout) {
       var allowedOutputExample = correctnessTest.getAnyAllowedOutput();
       var feedback = FeedbackObjectFactory.create(false);
       feedback.appendTextParagraph('Your code produced the following result:');
       feedback.appendCodeParagraph(
         'Input: ' + _jsToHumanReadable(correctnessTest.getInput()) + '\n' +
         'Output: ' + _jsToHumanReadable(observedOutput) + '\n' +
-        'Stdout: ' + stdOutput);
+        'Stdout: ' + stdout);
       feedback.appendTextParagraph('However, the expected output is:');
       feedback.appendCodeParagraph(
         _jsToHumanReadable(allowedOutputExample));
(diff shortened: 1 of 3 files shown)


client/services/code_evaluators/PythonCodeRunnerService.js
```diff
@@ -76,8 +76,9 @@ tie.factory('PythonCodeRunnerService', [
           }
 
           // The run was successful.
+          //outputLines = outputLines.slice(0, outputLines.length / correctnessTestResults.length);
```
eyurko: Don't leave commented-out code in the codebase.

client/services/FeedbackGeneratorService.js
```diff
@@ -92,13 +92,14 @@ tie.factory('FeedbackGeneratorService', [
...
-      correctnessTest, observedOutput) {
+      correctnessTest, observedOutput, stdOutput) {
       var allowedOutputExample = correctnessTest.getAnyAllowedOutput();
       var feedback = FeedbackObjectFactory.create(false);
       feedback.appendTextParagraph('Your code produced the following result:');
       feedback.appendCodeParagraph(
         'Input: ' + _jsToHumanReadable(correctnessTest.getInput()) + '\n' +
-        'Output: ' + _jsToHumanReadable(observedOutput));
+        'Output: ' + _jsToHumanReadable(observedOutput) + '\n' +
+        'stdout: ' + stdOutput);
```
eyurko: Why is this lowercase when Input and Output are capitalized?

seanlip:
A reminder: please could you reply to all comments, following the instructions in the playbook? In particular, for comments that you've addressed, please reply "Done." -- this helps keep track of which comments have been addressed and which haven't.

Also, please don't forget to read and respond to @eyurko's comment about the screenshot.

client/services/code_evaluators/PythonCodeRunnerService.js
cwcai633: Done.

client/services/FeedbackGeneratorService.js
cwcai633: Done.

cwcai633:
It works like this.
screenshot 2017-04-08 at 6 05 26 pm
screenshot 2017-04-08 at 6 04 38 pm
screenshot 2017-04-08 at 6 09 16 pm

However, it does not support to print out when there are errors in the code.
screenshot 2017-04-08 at 6 07 27 pm

seanlip:
Thank you for the screenshots! I am passing this to @rabidbit for a UI review. I can give you feedback on the code once the UI is agreed on.

rabidbit:
Please remove the trailing newline after stdout; otherwise, looks fine from a UI perspective Thanks!

client/services/FeedbackGeneratorService.js
```diff
@@ -92,13 +92,14 @@ tie.factory('FeedbackGeneratorService', [
     };
 
     var _getCorrectnessTestFeedback = function(
-      correctnessTest, observedOutput) {
+      correctnessTest, observedOutput, stdOutput) {
```
seanlip: Let's just call it stdout. There are two types of "output" here and I think it makes sense to think of the second one as just a stdout feed.

client/services/FeedbackGeneratorService.js
```diff
@@ -183,13 +184,17 @@ tie.factory('FeedbackGeneratorService', [
 
           var correctnessTests = task.getCorrectnessTests();
           var observedOutputs = codeEvalResult.getCorrectnessTestResults();
+          var stdOutputs = codeEvalResult.getOutput();
```
seanlip: Why plural?

client/services/FeedbackGeneratorService.js
```diff
@@ -183,13 +184,17 @@ tie.factory('FeedbackGeneratorService', [
...
           for (i = 0; i < correctnessTests.length; i++) {
             var observedOutput = observedOutputs[i];
+            var outputCount = stdOutputs.length / correctnessTests.length;
```
seanlip: It's not always going to be the case that the output for each test case is of equal length. You might need to find a way to catch each of them somewhere.

client/services/code_evaluators/PythonCodeRunnerService.js
```diff
@@ -77,7 +77,7 @@ tie.factory('PythonCodeRunnerService', [
 
           // The run was successful.
           return CodeEvalResultObjectFactory.create(
-            code, outputLines.join('\n'), correctnessTestResults,
+            code, outputLines, correctnessTestResults,
```
seanlip: If we are doing this then perhaps we might need to change the output property in CodeEvalResult to outputLines.

rabidbit:
Re: question (1) above, you could make the same argument for syntax highlighting, errors, and whether or not the code passes actual tests. I'd say not being able to print to console could be part of the "whiteboard" mode "where you are not provided with syntax highlighting and errors, no ability to print to console, and there is a timer". We ask if users would find this useful in the user study. We can also encourage people to try in whiteboard mode if/when we implement that. So, I think it's probably OK to let users print in the current UI (non-whiteboard mode).

client/services/FeedbackGeneratorService.js
cwcai633: TODO: support variable length stdout.

client/services/FeedbackGeneratorService.js
cwcai633: Done.

client/services/FeedbackGeneratorService.js
cwcai633: Done.

client/services/code_evaluators/PythonCodeRunnerService.js
cwcai633: Done.

rabidbit:
Per offline discussion, this feature will be placed on hold and not merged into TIE. We will revisit this if we decide to incorporate it later on.

seanlip:
Closing, since we are not going to merge this in right now (per previous discussion).

PR #183, opened 2017-04-07
Add feature to correctness test: support comparing list of strings

skywalker-lili:
Now, learner's code can output list of strings and correctness test will be able to return `True` if any of the allowedOutputs has the same strings (and in same order) as list of strings that learner's code output. 

The approach is to stringify list of strings so that they can be compared.

### DIFF
--- a/client/domain/CorrectnessTestObjectFactory.js
+++ b/client/domain/CorrectnessTestObjectFactory.js
@@ -31,18 +31,9 @@ tie.factory('CorrectnessTestObjectFactory', [
     };
 
     CorrectnessTest.prototype.matchesOutput = function(output) {
-    // If allowOutputs should be arrays and learner's code returns an array
-    // Then the if statement will check if learner's output matches
-    // any of arrays in the allowedOutputs
-      if (this._allowedOutputs.length > 0 && 
-          angular.isArray(this._allowedOutputs[0]) &&
-          angular.isArray(output)) {
-        var stringifiedOutput = JSON.stringify(output);
-        return this._allowedOutputs.some(function(allowedOutput) {
-          return JSON.stringify(allowedOutput) === stringifiedOutput;
-        });
-      }
-      return this._allowedOutputs.indexOf(output) !== -1;
+      return this._allowedOutputs.some(function(allowedOutput) {
+        return angular.equals(allowedOutput, output);
+      });
     };
 
     CorrectnessTest.prototype.getAnyAllowedOutput = function() {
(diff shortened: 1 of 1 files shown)


client/domain/CorrectnessTestObjectFactory.js
```diff
@@ -31,6 +31,22 @@ tie.factory('CorrectnessTestObjectFactory', [
     };
 
     CorrectnessTest.prototype.matchesOutput = function(output) {
+    // If allowOutputs should be arrays and learner's code returns an array
+    // Then the if statement will check if learner's output matches
+    // any of arrays in the allowedOutputs
+      if (this._allowedOutputs.length > 0 && 
+        this._allowedOutputs[0] instanceof Array &&
```
eyurko: angular.isArray

client/domain/CorrectnessTestObjectFactory.js
```diff
@@ -31,6 +31,22 @@ tie.factory('CorrectnessTestObjectFactory', [
...
+        output instanceof Array) { 
+      // Stringify the arrays and then compare
+        var stringifiedOutput = JSON.stringify(output);
+        for (var i = 0, l = this._allowedOutputs.length; i < l; i++) {
```
eyurko:
```
return this._allowedOutputs.every(function(output) {
  return JSON.stringify(output) === stringifiedOutput;
});
```

skywalker-lili: shouldn't this be this._allowedOutputs.some() ? since `this._allowedOutputs` can have multiple allowed outputs and learner should be able to pass if any of outputs matches?

client/domain/CorrectnessTestObjectFactory.js
```diff
@@ -31,6 +31,17 @@ tie.factory('CorrectnessTestObjectFactory', [
     };
 
     CorrectnessTest.prototype.matchesOutput = function(output) {
+    // If allowOutputs should be arrays and learner's code returns an array
```
seanlip: I'm a little confused by all this. Would the following work just as well?

```
    return this._allowedOutputs.some(function(allowedOutput) {
      return angular.equals(allowedOutput, output);
    });
```

or is there some case where it fails and you need the more complicated version?

skywalker-lili: Modified as your suggested. It's cleaner and more efficient. No traps here, just I didn't know JS/angular before and wasn't ware that angular had this beauty...

skywalker-lili:
Hi @eyurko , I have changed the function to use `angular.equals()` nested in `array.some()` to re implement this function. Now this unction would return `True` if learner's output matches any of `this._allowedOutputs`, and the output can be an primitive type as  before, or be an array of primitives. 

Please PTAL.

client/domain/CorrectnessTestObjectFactory.js
skywalker-lili: Done.

seanlip:
@eyurko could you approve?

Issue #155, opened 2017-04-06
Add a pre-push git hook to catch silly mistakes prior to submit.


seanlip:
It would be nice to have a script that runs the lint checks and karma tests just prior to a push -- see e.g. the "pre-push" hook in [this link](https://git-scm.com/book/gr/v2/Customizing-Git-Git-Hooks). This script would then be run automatically after "git push" is done on the command line.

We probably do not need to include e2e tests in this script, since they generally take a while to run and the developer is expected to do some manual testing anyway.

seanlip:
Fixed in #184. Thanks @shaman-rajan!

PR #184, opened 2017-04-07
Solving #155 Hook to run tests on current code before pushing

shaman-rajan:
For #155 

When configured to be used, runs linter tests before pushing and checks if the tests passed (1 line in output only in this case). If tests fail, option to push anyway.

Karma test not set up yet.


scripts/setup_hooks.sh
```diff
@@ -0,0 +1,3 @@
+#!/bin/sh
+
+git config core.hooksPath "./hooks"
```
seanlip: Quick check: let's say I set up a different git project on the same machine. Does this hook run?

(If so, that's a problem.)

shaman-rajan: If this script is run from a directory that belongs to that repository, then it will run the hook from that repo. Otherwise, the config is set only locally for this repo (since it's missing the global flag).

shaman-rajan:
@seanlip PTAL

scripts/setup_hooks.sh
seanlip: OK. What instructions does a dev need to follow in order to set up the hook? Would it make sense to include this in the main setup script so they do this automatically?

hooks/pre-push
```diff
@@ -0,0 +1,23 @@
+#!/bin/bash
+
```
seanlip: Add copyright etc. boilerplate at top (see other scripts).

hooks/pre-push
```diff
@@ -0,0 +1,23 @@
...
+echo "Running linter tests"
+lines=$(sh scripts/run_linter.sh | wc -l)
+if [ $lines -eq 1 ]; then
```
seanlip: I think looking at the exit codes would be a more direct signal. See how travis does it.

hooks/pre-push
```diff
@@ -0,0 +1,23 @@
...
+    case $yn in
+        [Yy]* ) ;;
+        [Nn]* ) echo "Not pushing changes"; exit 1;;
+        * ) echo "Invalid input";;
+    esac
+fi
+
+echo "Running karma tests"
+#TODO
+#sh scripts/run_karma_tests.sh
```
seanlip: This needs to work before this can be checked in.

hooks/pre-push
```diff
@@ -0,0 +1,23 @@
+#!/bin/bash
+
+echo "Running linter tests"
+lines=$(sh scripts/run_linter.sh | wc -l)
+if [ $lines -eq 1 ]; then
+    echo "Linter tests passed"
+else
+    # Assign standard input to keyboard
+    exec < /dev/tty
```
seanlip: Just to check, is this necessary? If it was omitted, what would happen?

shaman-rajan: If omitted, the script does not wait for an input ([reference](http://stackoverflow.com/questions/3417896/how-do-i-prompt-the-user-from-within-a-commit-msg-hook))

seanlip: OK, thanks! Maybe add the ref (shortlink) to the comment.

scripts/setup_hooks.sh
shaman-rajan: Yes, this could be added to the main setup script. In that case, we would also need to make the developer aware of the existence of the hook script and include an option (a flag) to disable this (in case he is using custom hooks of his own, or for any other reason).

seanlip: OK. I think I want to get a second opinion on this -- @rabidbit @eyurko do you have any thoughts/preferences?

seanlip: (To clarify, the question is what the default behavior for this hook, which does linting and runs tests before pushing to github, should be. Should it be enabled automatically in the default workflow, or be more of an optional add-on?)

seanlip: Hi @shaman-rajan -- I checked with @rabidbit, and I think it's fine to just add this to the main setup script. So then we would tell devs: run the main setup script on your first push, and then thereafter you just need to "git push" and not run anything else (incl. tests). We can also add a bypass --disable-presubmit-checks flag too. How does that sound?

shaman-rajan: Sounds good. Will add this functionality.

shaman-rajan:
@seanlip PTAL

Both linter and karma tests run before the push command.

hooks/pre-push
```diff
@@ -0,0 +1,57 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# Usage:
+#     Not meant for manual use
+#     Runs automatically before a git push command is executed
+#     Behaviour can be set up using "bash scripts/setup_hooks.sh"
+
+# Ask user if push should continue anyway if a test fails
```
seanlip: Add full stop at end. Also remove blank line below (I assume this line is meant to be connected to the function below.)

hooks/pre-push
```diff
@@ -0,0 +1,57 @@
...
+    exec < /dev/tty
+
+    read -p "$1 tests failed. Do you want to continue anyway? (y/n) " yn
+    case $yn in
+        [Yy]* ) ;;
+        [Nn]* ) echo "Not pushing changes"; exit 1;;
+        * ) echo "Invalid input"; exit 1;;
+    esac
+
+    # Close standard input
```
seanlip: Add periods at ends of sentences, here and elsewhere.

hooks/pre-push
```diff
@@ -0,0 +1,57 @@
...
+
+# Usage:
+#     Not meant for manual use
+#     Runs automatically before a git push command is executed
+#     Behaviour can be set up using "bash scripts/setup_hooks.sh"
+
+# Ask user if push should continue anyway if a test fails
+
+proceed_when_test_fail () {
+    # Assign standard input to keyboard
```
seanlip: I think we use 2-space indentation for all our scripts, let's just do that in general to keep things consistent.

hooks/pre-push
```diff
@@ -0,0 +1,57 @@
...
+    if [ $? -eq 0 ]
+    then
+        echo "$1 tests passed"
+    else
+        proceed_when_test_fail "$1"
+    fi
+}
+
+echo "Running linter tests"
+sh scripts/run_linter.sh
```
seanlip: sh --> bash ? Ditto elsewhere.

karma.auto_test.conf.js
```diff
@@ -0,0 +1,82 @@
+// Karma configuration for TIE.
```
seanlip: Is it possible to just replace the existing Karma config with this one? I don't really see the benefit of having two separate test configs.

scripts/setup.sh
```diff
@@ -83,3 +92,6 @@ fi
 
 # Generate a list of already-installed modules.
 NPM_INSTALLED_MODULES="$(npm list)"
+
+# Set up hooks if not disabled
```
seanlip: Is this comment supposed to refer to something?

shaman-rajan:
@seanlip I've made the changes based on comments. The last commit replaces the old karma config with the new one. The only noticeable difference between them is "singleRun: true", which closes the browser and exits the script once it completes the tests. If this is not desirable, I will revert that commit and use the 2 separate config files.

scripts/setup.sh
```diff
@@ -83,3 +93,5 @@ fi
 
 # Generate a list of already-installed modules.
 NPM_INSTALLED_MODULES="$(npm list)"
+
```
seanlip: Don't need all the extra lines here.

seanlip:
(Just fix one minor thing please, but after that feel free to merge without further review.)

shaman-rajan:
@seanlip please merge (I do not have the permissions). Requested changes made.

PR #192, opened 2017-04-10
get question IDs from keys of question dict.

lk-chen:
A new key-value pair is added to globalData.questions in each &lt;question>.js file. Thus we can get the list of question IDs from keys of globalData.questions, I guess?


lk-chen:
I just found out that there is a similar TODO for the same issue, in TaskSchemaValidationServiceSpec.js .

seanlip:
Ah yep, @eyurko just noticed that too. Would you mind fixing that one too?

lk-chen:
I have fixed it. Shall I put that commit in this same PR, or shall I open another PR?

seanlip:
Another PR (and separate branch), please -- this one's already been merged. Thanks!

PR #198, opened 2017-04-11
Changed Task Schema Validation to use globalData.questions.

eyurko:


seanlip:
Ah, it auto-recognizes and omits it. Interesting. https://github.com/google/tie/commit/19188d051944815be678de8fe8a80c990086a458

So diffs shown in github aren't actually against master, they are against the previous state of the branch. Good to know :)

PR #199, opened 2017-04-11
Fix #119: Create question findClosestValue in BST

### LINKED ISSUE #119: Create "Find closest value in BST" question

Eumenides2009:
fix #119


assets/questions/bstClosestValue.js
```diff
@@ -0,0 +1,173 @@
...
+            return cls.failToCompareAll([kid,target])
+`
+  },
+  tasks: [{
+    instructions: [
+      [
+        'In this question, you need to find the closest value in a non-empty binary search tree ',
+        'given the target value. The target value is given in float format. It\'s guaranteed ',
+        'that there\'s only one closest value. You will have the root TreeNode. Each TreeNode ',
+        'has attributes val, left and right.'
```
rabidbit: Recommend ending this paragraph with a specific task or question along with sample input and output.

eprym: Add a simple example in the instruction. And the explanation of the serialized format is already in the comment part of the main function

rabidbit: The user will not see the comments in the code (we don't expect users to open the code GitHub or in the console or anything); so, it should be explained in the instructions. Also, please keep in mind that we currently do not have a way for the user to ask clarifying questions if they don't understand the question; so, you may need to be more explicit than you normally would if you were running a live interview.

eprym: screenshot from 2017-04-12 16 51 29

The comments are not in the code, it's in webpage which users will see it. The reason why we don't put it in the instructions is that it's hard to put multiline strings in the instruction struct. It seems to strip the whitespace automatically.

rabidbit: Hmm... seems we should call it a bug for the instructions struct. I would prefer to have the instructions in the instructions window (or the question window in this case) rather than taking up space in the coding window and causing confusion from a UX perspective. @seanlip @eyurko, what do you think?

seanlip: Yes, let's file a bug for the instructions struct. If the current format is not sufficiently expressive then we should try and figure out one that is.

Eumenides2009: Filed #205 bug

Eumenides2009:
Hi @eyurko   and @rabidbit  could you please take a look and let me know what I need to do here? 
Thanks

rabidbit:
Thanks for filing the bug. That bug should get fixed so that this question can be formatted correctly and placed in the correct window. From a UX perspective, this PR is blocked by that bug.

Eumenides2009:
Thank you @rabidbit , I will move on to another issue and wait for this bug to be fixed

PR #201, opened 2017-04-11
Fix #189: Question Title UI Modifications

### LINKED ISSUE #189: Question Title UI Modifications for better space utilization and prominence
**Description:** Every question title has a prefix "Question #:" in it. This is redundant as the question index is already being displayed in the navigation bar on the top. The highlighted index number displays the current question number. So, we can remove the "Question #:" prefix from the question title. 
Also, the CSS of the question title can be modified to make it more prominent.

**Current state:** 
screen shot 2017-04-10 at 10 37 54 am

**Suggested Modification:** The following screenshot is just a mock-up of the suggested modifications.We can have a discussion about the best way to approach the same. 
screen shot 2017-04-07 at 10 25 59 am

abinash-behera:
a) Removed the 'Question #:' prefix from question title
b) Changed the text color

screenshot from 2017-04-11 14 46 41

### DIFF
--- a/client/components/LearnerViewDirective.js
+++ b/client/components/LearnerViewDirective.js
@@ -258,6 +258,9 @@ tie.directive('learnerView', [function() {
         .tie-previous-instructions {
           opacity: 0.5;
         }
+        .tie-question-title {
+          color: rgb(66, 133, 244);
+        }
         .tie-question-ui {
           vertical-align: top;
         }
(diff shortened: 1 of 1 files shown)


client/components/LearnerViewDirective.js
```diff
@@ -336,6 +336,9 @@ tie.directive('learnerView', [function() {
         .tie-step-unlocked {
           background-color: rgb(0, 128, 0);
         }
+        .tie-question-title {
```
rabidbit: Please insert this alphabetically (i.e., before .tie-question-ui).

abinash-behera: Done!

PR #203, opened 2017-04-13
Fix #139: Add code reset button to page

### LINKED ISSUE #139: No reset button to reset the current code
@sniffsky and I are currently working on #98 , however, we realize that if we are able to save codes, then it's probably good to have a reset button once #98 is resolved.

wmcalyj:
This change contains both UI change (new button added) and backend change (code are removed from localStorage)


seanlip:
OK -- I'm not sure what's going on here, but I'm pretty certain it shouldn't include 118 commits. Could you close this, sync to latest version of master, checkout a brand new branch, make your changes on that branch, and push it?

wmcalyj:
Hmm.. That's weird. I didn't even notice it. OK. I will do that.

PR #207, opened 2017-04-13
Fix #202 User can resize coding terminal window

### LINKED ISSUE #202: Allow user to change size of code editing area
From feedback received in #112 , users would like to be able to change the size of the code editing area in a way that is similar to the other boxes on the page.

shaman-rajan:
Fixing #202 

image

The icon on the bottom right can be used to change the coding terminal window size similar to other text boxes on the screen.


seanlip:
So, there's a benefit here in terms of increased functionality, but a cost due to requiring additional third-party imports. What is the size of the HTML/JS transferred over the wire with and without the new dependencies?

rabidbit:
Thanks for taking this on! Can this be done purely with CSS? I would prefer to avoid importing additional libraries and image files to keep this app lightweight and simple.

shaman-rajan:
Size of new libraries: 382kB
Total size of third party libraries: 2MB

There may be a way of doing this with CSS using "resize: both". But with that, the resize icon is hidden behind the coding UI inside the div - Only when I hide the coding UI it allows me to resize the div.

shaman-rajan:
I've done some more work on this, and this is what I found:
- Jqueryui adds DOM elements on top of the existing ones and sets them at a high z-index so they are visible over the current elements in the div, and uses drag events on those elements to resize the div
- The CSS resize option does not do this - the resize icon is attached to the div itself. so in order for the resize icon on the bottom right to be visible, we would need to increase the z-index of the div. But, this would then hide the coding UI so is not an option.
- One fix I see is to make the parent div a little bigger than the coding UI so that the resize icon does not overlap with it, but this would make it look pretty ugly (I tried).

I don't think we can do this with plain CSS, and should check whether adding 382kB of libraries is worth adding the feature. (libraries may be useful later too)

shaman-rajan:
Discussed this with @seanlip and we decided that since 
- It is not possible to do this with just CSS,
- jquery-ui is a standard library and only ~300kB (20% of the total network load)
we will use it to implement the feature.
@rabidbit PTAL

rabidbit:
SGTM (using jquery-ui). The resizing controls should be consistent across windows; so, I assume you will use jquery-ui for all three windows?

shaman-rajan:
Was currently using jquery-ui for just 1 window, will update code to use it for all 3.

shaman-rajan:
@rabidbit changes made. PTAL

rabidbit:
Since overflow is no longer set to auto, you cannot scroll the feedback or question windows when content height exceeds window height. Also, the content text would overflow past the window a bit when that happens.

shaman-rajan:
@rabidbit Fixed. Everything should be as according to requirement now.

rabidbit:
Scrollbars are misaligned. Please see screenshots below.

misaligned-scrollbars-1
misaligned-scrollbars-2

seanlip:
Hi @shaman-rajan -- there seem to be some merge conflicts. Could you please fix them?

client/app.html
```diff
@@ -4,7 +4,10 @@
     <meta charset="UTF-8">
     <meta name="viewport" content="width=device-width, initial-scale=1.0, user-scalable=yes">
     <link rel="stylesheet" href="../third_party/codemirror-5.19.0/lib/codemirror.css">
+    <link rel="stylesheet" href="../third_party/jquery-ui-1.12.1.custom/jquery-ui.min.css">
```
seanlip: We can't be using a custom build here. Can we just use standard jquery ui instead?

Rationale: we may need to upgrade the libraries in the future, and in doing so, would need to know exactly what to include. So, probably better to just use the default vanilla jquery UI.

seanlip:
@rabidbit @shaman-rajan I propose we get this PR ready for merge, but hold off for a little bit on actually merging it until we've done the experiment that @theresa-lee is working on in order to use a larger window for coding. I'd like to see how the two compare (in terms of user feedback) before committing to a full alternative UI or a larger third-party dependency. Would that be OK?

shaman-rajan:
Addressed comments. PR should be ready to merge.

seanlip:
Closing this for now since we closed #202, but we'll reopen and use it if user feedback suggests it again (with the new UI). Thanks @shaman-rajan!

PR #209, opened 2017-04-13
Fix #208: Add code auto-save feature

### LINKED ISSUE #208: Add Auto-save feature to save codes to localStorage
Add Auto-save feature to save codes to localStorage. A follow up feature for #98

wmcalyj:
With the new feature, the code will be auto-saved every minute.
This pull request contains UI change (Display/Hide 'Saving code now...')

### DIFF
--- a/client/components/LearnerViewDirective.js
+++ b/client/components/LearnerViewDirective.js
@@ -384,7 +384,7 @@ tie.directive('learnerView', [function() {
         var SECONDS_TO_MILLISECONDS = 1000;
         // Default time interval, in seconds, after which code will
         // be auto-saved.
-        var DEFAULT_AUTOSAVE_SECONDS = 60;
+        var DEFAULT_AUTOSAVE_SECONDS = 30;
         // "Saving code now..." will last for 3 seconds and disappear.
         var DISPLAY_AUTOSAVE_TEXT_SECONDS = 3;
 
--- a/client/components/LearnerViewDirective.js
+++ b/client/components/LearnerViewDirective.js
@@ -71,6 +71,9 @@ tie.directive('learnerView', [function() {
                     ng-click="resetCode()">
                   Reset Code
                 </button>
+                <div class="tie-code-auto-save" ng-show="displayAutoSaveText">
+                  Saving code now...
+                </div>
                 <button class="tie-run-button"
                     ng-class="{'active': !nextButtonIsShown}"
                     ng-click="submitCode(code)"
@@ -119,6 +122,12 @@ tie.directive('learnerView', [function() {
           position: absolute;
           top: calc(50% - 25px);
           width: 50px;
+        }
+        .tie-code-auto-save {
+          font-family: Roboto, 'Helvetica Neue', 'Lucida Grande', sans-serif;
+          float: left;
+          margin-top: 10px;
+          margin-left: 10px;
         }
         .tie-code-reset {
           float: left;
@@ -370,6 +379,13 @@ tie.directive('learnerView', [function() {
           ].join('\n')
         ];
 
+        var SECONDS_TO_MILLISECONDS = 1000;
+        // Default time interval in seconds that codes will be auto saved.
+        // Codes will be auto-saved every minute.
+        var DEFAULT_AUTO_SAVE_SECONDS = 60;
+        // "Saving code now..." will last for 3 seconds and disappear.
+        var DISPLAY_AUTO_SAVE_TEXT_SECONDS = 3;
+
         var congratulatoryFeedback = FeedbackObjectFactory.create();
         QuestionDataService.initCurrentQuestionSet(questionSetId);
         var questionSet = QuestionDataService.getCurrentQuestionSet(
@@ -381,6 +397,7 @@ tie.directive('learnerView', [function() {
         for (var i = 0; i < $scope.questionIds.length; i++) {
           $scope.questionsCompletionStatus.push(false);
         }
+        $scope.displayAutoSaveText = false;
         var question = null;
         var tasks = null;
         var currentTaskIndex = null;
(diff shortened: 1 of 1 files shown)


client/components/LearnerViewDirective.js
```diff
@@ -370,6 +379,13 @@ tie.directive('learnerView', [function() {
           ].join('\n')
         ];
 
+        var SECONDS_TO_MILLISECONDS = 1000;
+        // Default time interval in seconds that codes will be auto saved.
+        // Codes will be auto-saved every minute.
```
seanlip: Replace entire comment with: Default time interval, in seconds, after which code will be auto-saved.

client/components/LearnerViewDirective.js
```diff
@@ -381,6 +397,7 @@ tie.directive('learnerView', [function() {
         for (var i = 0; i < $scope.questionIds.length; i++) {
           $scope.questionsCompletionStatus.push(false);
         }
+        $scope.displayAutoSaveText = false;
```
seanlip: Let's go with `$scope.autosaveTextIsDisplayed = false;`, so that this does not start with a verb (which makes it seem like an action).

client/components/LearnerViewDirective.js
```diff
@@ -517,9 +534,29 @@ tie.directive('learnerView', [function() {
           loadQuestion(questionId, questionSet.getIntroductionParagraphs());
         };
 
+        var displayAutoSaveTextForSeconds = function(displaySeconds) {
+          $scope.displayAutoSaveText = true;
+          $scope.$apply();
```
seanlip: Why do you need the $scope.$apply()?

client/components/LearnerViewDirective.js
```diff
@@ -517,9 +534,29 @@ tie.directive('learnerView', [function() {
...
+          $timeout(function() {
+            $scope.displayAutoSaveText = false;
+          }, displaySeconds * SECONDS_TO_MILLISECONDS);
+        };
+
+        var codeAutoSave = function() {
```
seanlip: activateAutosaveFunctionality() or activateAutosaving() perhaps -- start with a verb.

client/components/LearnerViewDirective.js
```diff
@@ -517,9 +534,29 @@ tie.directive('learnerView', [function() {
...
+          setInterval(function() {
```
seanlip: Angular has $interval.

client/components/LearnerViewDirective.js
```diff
@@ -517,9 +534,29 @@ tie.directive('learnerView', [function() {
           loadQuestion(questionId, questionSet.getIntroductionParagraphs());
         };
 
+        var displayAutoSaveTextForSeconds = function(displaySeconds) {
```
seanlip: triggerAutosaveNotification

("Text" is a bit vague and feels like it can refer to the code. The arg name means you don't really need the "ForSeconds", I think.)

client/components/LearnerViewDirective.js
```diff
@@ -370,6 +379,13 @@ tie.directive('learnerView', [function() {
           ].join('\n')
         ];
 
+        var SECONDS_TO_MILLISECONDS = 1000;
+        // Default time interval in seconds that codes will be auto saved.
+        // Codes will be auto-saved every minute.
+        var DEFAULT_AUTO_SAVE_SECONDS = 60;
```
seanlip: Consider AUTO_SAVE --> AUTOSAVE here and elsewhere. It appears to be one word: https://en.wikipedia.org/wiki/Autosave

seanlip:
@rabidbit you might want to take a look at the UI.

client/components/LearnerViewDirective.js
wmcalyj: Done

client/components/LearnerViewDirective.js
wmcalyj: Done

client/components/LearnerViewDirective.js
wmcalyj: Done

client/components/LearnerViewDirective.js
wmcalyj: Done

client/components/LearnerViewDirective.js
wmcalyj: Done

client/components/LearnerViewDirective.js
wmcalyj: Somehow, if I don't do $scope.$apply, displayAutoSaveText won't be changed. I will double check, but I am pretty sure it wasn't working without $scope.$apply().

wmcalyj: Interesting, now, it's working without $scope.$apply(). Not sure if it's because I used $interval instead of setInterval(). Will play around it.

wmcalyj: Yep, looks like, if I use setInterval, the notification won't display.

client/components/LearnerViewDirective.js
wmcalyj: Done

wmcalyj:
@seanlip Address backend code change. PTAL

rabidbit:
Design comments:
1. Prefer "Saving code..." versus "Saving code now..." More succinct with unnecessary "now" qualifier
2. Prefer smaller font (13px)

From a UX perspective, rather than saving every minute, how hard would it be to activate auto-save 3 seconds after an edit event and spaced 5 seconds apart? That is, if I make a quick change, it auto-saves after 3 seconds. If I keep typing non-stop for while, it saves every 5 seconds while I type until I finish typing (it saves one last time 5 seconds after I finish typing the long code). If I don't change anything, it does not activate auto-save. I think this would be better than saving every minute, regardless of what the user does.

wmcalyj:
@rabidbit OK. I will make changes for 1 & 2. 
For the third change, I will do some research and see how to accomplish that. But I think that's possible and makes sense. Though I think it will require a lot of changes. So, if I will be still on this project next week, I will try to finish that change.

wmcalyj:
@rabidbit While trying to figure out how to do that, I found one issue. How do you define quick change? I mean, can we combine the two cases? Meaning, if the model is changed, I will do a debounce and save the change, things like that? How does that sound like?

seanlip:
LGTM from code perspective. @rabidbit should we merge it or hold off until further design changes are made? Your call.

seanlip:
Also @wmcalyj tests are failing. PTAL.

rabidbit:
I don't think this is the optimal approach, but it's better than what we have now (no auto-save at all). I recommend increasing the save frequency to 30sec (1min felt kind of long to me). I think we should update the auto-save feature to do what I described, but let's at least have an auto-save in there (merge this). I can open an issue with the requested design changes.

wmcalyj:
@rabidbit Addressed UI review changes. Can you assign me the issue once you created that? Want to work on that.

rabidbit:
OK, SG. Will do. Merging.

PR #210, opened 2017-04-14
Added very basic theming function and an alternative dark mode

theresa-lee:
There is now a Light (default) and Dark theme you can switch between using the selector at the bottom. I took out some borders and added a file for the CodeMirror styling. 

Probably can regulate the themes a bit better with some CSS framework but for now, this should work.

### DIFF
--- a/client/app.html
+++ b/client/app.html
@@ -13,9 +13,11 @@
     <script src="../third_party/ui-codemirror-0.3.0/ui-codemirror.min.js"></script>
     <script src="app.js"></script>
     <script src="../assets/questions/i18n.js"></script>
+    <script src="../assets/questions/palindrome.js"></script>
     <script src="../assets/questions/parens.js"></script>
     <script src="../assets/questions/reverseWords.js"></script>
     <script src="../assets/questions/rle.js"></script>
+    <script src="../assets/questions/bestMeetupLocation.js"></script>
     <script src="../assets/question_sets/strings.js"></script>
     <script src="components/CodeSnippetDirective.js"></script>
     <script src="components/LearnerViewDirective.js"></script>
--- a/client/app.html
+++ b/client/app.html
@@ -4,6 +4,7 @@
     <meta charset="UTF-8">
     <meta name="viewport" content="width=device-width, initial-scale=1.0, user-scalable=yes">
     <link rel="stylesheet" href="../third_party/codemirror-5.19.0/lib/codemirror.css">
+    <link rel="stylesheet" href="../third_party/codemirror-5.19.0/lib/material.css">
     <script src="../third_party/angular-1.6.1/angular.min.js"></script>
     <script src="../third_party/codemirror-5.19.0/lib/codemirror.js"></script>
     <script src="../third_party/codemirror-5.19.0/mode/python/python.js"></script>
@@ -43,7 +44,7 @@
     <script src="services/code_preprocessors/CodePreprocessorDispatcherService.js"></script>
     <script src="services/code_preprocessors/PythonCodePreprocessorService.js"></script>
   </head>
-  <body>
+  <body class="{{bodyClass}}">
     <learner-view></learner-view>
   </body>
 </html>
--- a/client/app.html
+++ b/client/app.html
@@ -33,6 +33,7 @@
     <script src="domain/TaskObjectFactory.js"></script>
     <script src="domain/TracebackCoordinatesObjectFactory.js"></script>
     <script src="domain/TranscriptObjectFactory.js"></script>
+    <script src="services/CodeStorageService.js"></script>
     <script src="services/FeedbackGeneratorService.js"></script>
     <script src="services/QuestionDataService.js"></script>
     <script src="services/SolutionHandlerService.js"></script>
--- a/client/app.html
+++ b/client/app.html
@@ -44,7 +44,7 @@
     <script src="services/code_preprocessors/CodePreprocessorDispatcherService.js"></script>
     <script src="services/code_preprocessors/PythonCodePreprocessorService.js"></script>
   </head>
-  <body class="{{bodyClass}}">
+  <body>
     <learner-view></learner-view>
   </body>
 </html>
--- a/client/components/LearnerViewDirective.js
+++ b/client/components/LearnerViewDirective.js
@@ -345,16 +345,15 @@ tie.directive('learnerView', [function() {
     `,
     controller: [
       '$scope', '$timeout', 'SolutionHandlerService', 'QuestionDataService',
-      'LANGUAGE_PYTHON', 'FeedbackObjectFactory',
+      'LANGUAGE_PYTHON', 'FeedbackObjectFactory', 'CodeStorageService',
       function(
           $scope, $timeout, SolutionHandlerService, QuestionDataService,
-          LANGUAGE_PYTHON, FeedbackObjectFactory) {
+          LANGUAGE_PYTHON, FeedbackObjectFactory, CodeStorageService) {
         var DURATION_MSEC_WAIT_FOR_SCROLL = 20;
         var language = LANGUAGE_PYTHON;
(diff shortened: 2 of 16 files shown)


client/app.html
```diff
@@ -43,7 +44,7 @@
     <script src="services/code_preprocessors/CodePreprocessorDispatcherService.js"></script>
     <script src="services/code_preprocessors/PythonCodePreprocessorService.js"></script>
   </head>
-  <body>
+  <body class="{{bodyClass}}">
```
seanlip: Do you think we could declare the bodyClass on the topmost DOM element inside the learner-view directive instead, rather than on the body tag? It might be necessary to introduce another wrapper div around the whole thing (I'm not sure) but I think that that would be fine if it's needed.

client/components/LearnerViewDirective.js
```diff
@@ -37,7 +37,7 @@ tie.directive('learnerView', [function() {
             </div>
           </div>
           <div class="tie-coding-ui">
-            <div class="tie-feedback-window">
+            <div class="tie-feedback-window {{tieFeedbackWindowClass}}">
```
seanlip: Take a look at ng-class: https://docs.angularjs.org/api/ng/directive/ngClass

client/components/LearnerViewDirective.js
```diff
@@ -89,7 +95,7 @@ tie.directive('learnerView', [function() {
             </div>
           </div>
           <div class="tie-question-ui">
-            <div class="tie-question-window">
+            <div class="tie-question-window {{tieQuestionWindowClass}}">
```
seanlip: Take a look at ng-class: https://docs.angularjs.org/api/ng/directive/ngClass

client/components/LearnerViewDirective.js
```diff
@@ -144,10 +153,10 @@ tie.directive('learnerView', [function() {
         }
         .tie-coding-terminal, .tie-question-window {
           background-color: rgb(255, 255, 255);
-          border-color: rgb(222, 222, 222);
-          border-radius: 3px;
-          border-style: solid;
-          border-width: 1px;
+          -webkit-font-smoothing: antialiased;
+        }
+        .tie-coding-terminal.night-mode, .tie-question-window.night-mode {
```
seanlip: Why do you need a duplicate rule here that repeats the same things as the CSS rule immediately above?

client/components/LearnerViewDirective.js
```diff
@@ -370,6 +397,11 @@ tie.directive('learnerView', [function() {
           ].join('\n')
         ];
 
+        $scope.themes = [
+          {themeName: "Light"},
```
seanlip: Here and below, let's go with single quotes for consistency.

client/components/LearnerViewDirective.js
```diff
@@ -443,6 +475,21 @@ tie.directive('learnerView', [function() {
...
+          if ($scope.theme === "Dark") {
+            $rootScope.bodyClass = "night-mode";
+            $scope.codeMirrorOptions.theme = "material";
+            $scope.tieFeedbackWindowClass = "night-mode";
+            $scope.tieQuestionWindowClass = "night-mode";
+          }
+          if ($scope.theme === "Light") {
+            $rootScope.bodyClass = "";
+            $scope.codeMirrorOptions.theme = "default";
+            $scope.tieFeedbackWindowClass = "tie-feedback-window";
```
seanlip: Here and above, it would be better to set a single boolean $scope.isInDarkMode or perhaps $scope.theme and then use ng-class to handle the rest.

client/components/LearnerViewDirective.js
```diff
@@ -67,6 +67,12 @@ tie.directive('learnerView', [function() {
                 <select class="tie-lang-select-menu" name="lang-select-menu">
                   <option value="Python" selected>Python</option>
                 </select>
+                <select class="tie-theme-select" name="theme-select" 
+                        ng-change="changeTheme()" ng-model="theme"
+                        ng-options="i.themeName as i.themeName for i in themes">
+                  <option style="display:none" value="">Theme</option>
```
seanlip: Nit: Add space after colon

client/components/LearnerViewDirective.js
```diff
@@ -353,11 +378,13 @@ tie.directive('learnerView', [function() {
       </style>
     `,
     controller: [
-      '$scope', '$timeout', 'SolutionHandlerService', 'QuestionDataService',
-      'LANGUAGE_PYTHON', 'FeedbackObjectFactory', 'CodeStorageService',
+      '$rootScope', '$scope', '$timeout', 'SolutionHandlerService',
```
seanlip: Hopefully we can remove $rootScope.

rabidbit:
Looks cool!

Design comments:
1. After selecting theme, the drop-down menu remains highlighted. It should reset to non-highlighted mode upon selection or cancelling selection.
2. In dark mode, there is a white outline on the circles. Please remove.
3. Please make relative shades in dark mode consistent with light mode (i.e., the question window on the right and the coding window on the bottom left are the same color while the feedback window on the upper left is different). Recommend changing question window on right to be the same color as coding window.

client/app.html
theresa-lee: Yeah, it required a wrapper around the top level html object in the learner-view directive (also is why I had to add the height attribute a couple of places), but no reference to rootScope anymore!

client/components/LearnerViewDirective.js
theresa-lee: Done.

client/components/LearnerViewDirective.js
theresa-lee: Done.

client/components/LearnerViewDirective.js
theresa-lee: Done.

client/components/LearnerViewDirective.js
theresa-lee: Sorry, I forgot to remove this from something I was trying earlier!

client/components/LearnerViewDirective.js
theresa-lee: Done.

client/components/LearnerViewDirective.js
theresa-lee: Done.

client/components/LearnerViewDirective.js
theresa-lee: Done.

client/components/LearnerViewDirective.js
```diff
@@ -477,7 +521,8 @@ tie.directive('learnerView', [function() {
           lineNumbers: true,
           mode: LANGUAGE_PYTHON,
           smartIndent: true,
-          tabSize: 4
+          tabSize: 4,
+          theme: "default"
```
seanlip: nit: use single quotes for consistency with code elsewhere in this file

seanlip:
LGTM from code perspective, thanks @theresa-lee! Deferring to @rabidbit for final UI review, approval and merge.

rabidbit:
Waiting on focus highlight fix.

theresa-lee:
Made the quote fix and removed focus highlighting on the theme select.

client/components/LearnerViewDirective.js
```diff
@@ -21,97 +21,116 @@ tie.directive('learnerView', [function() {
...
         body {
           background-color: rgb(242, 242, 242);
           font-family: Roboto, 'Helvetica Neue', 'Lucida Grande', sans-serif;
           font-size: 15px;
+          height: 100%;
+          margin: 0px;
+        }
+        .tie-wrapper {
+          height: 100%;
+        }
```
rabidbit: For the css style statements, please order alphabetically, unless it impacts cascading.

theresa-lee: Done.

PR #214, opened 2017-04-14
add string to dictionary question

alblin:
do not thinking having more than one input argument per test case is currently supported - implemented a workaround that may be slightly confusing to user

### DIFF
--- a/assets/question_sets/strings.js
+++ b/assets/question_sets/strings.js
@@ -26,5 +26,5 @@ globalData.questionSets['strings'] = {  // eslint-disable-line dot-notation
       "I will provide you with feedback."
     ].join('')
   ],
-  questionIds: ['reverseWords', 'parens', 'i18n', 'rle']
+  questionIds: ['reverseWords', 'parens', 'i18n', 'rle', 'stringToDict']
 };
--- a/assets/question_sets/strings.js
+++ b/assets/question_sets/strings.js
@@ -26,5 +26,5 @@ globalData.questionSets['strings'] = {  // eslint-disable-line dot-notation
       "I will provide you with feedback."
     ].join('')
   ],
-  questionIds: ['reverseWords', 'parens', 'i18n', 'rle', 'stringToDict']
+  questionIds: ['reverseWords', 'parens', 'i18n', 'rle']
 };
--- a/assets/questions/stringToDict.js
+++ b/assets/questions/stringToDict.js
@@ -20,16 +20,7 @@ globalData.questions['stringToDict'] = {  // eslint-disable-line dot-notation
   title: 'Split String to Words',
   starterCode: {
     python:
-`def preProcess(s): // necessary because no functionality for taking multiple inputs
-    // s is passed in as 'stringtoparse:words in dictionary' and converted two two parameters
-    spl = s.split(":")
-    dict_str = spl[1].split(" ")
-    dict = set()
-    for el in dict_str:
-      dict.add(el)
-    return stringToDict(spl[0], dict)
-
-def stringToDict(s, dict):
+`def stringToDict(lst):
     return ""
 `
   },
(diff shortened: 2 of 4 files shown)


assets/questions/stringToDict.js
```diff
@@ -0,0 +1,123 @@
...
+      [
+        'Note: You do not need to understand the preProcess function. Preprocess does some ',
+        'preprocessing that allows two arguments to be passed into the stringToDict function which ',
+        'you will implement. An input string is passed into preProcess that contains both the ',
+        'string you are to parse and the given dictionary for stringToDict, separated by a colon.'
+      ].join('')
+    ],
+    prerequisiteSkills: ['Arrays', 'Strings', 'Sets'],
+    acquiredSkills: ['String Parsing'],
+    inputFunctionName: null,
```
eyurko: This is why we have inputFunctionName -- it'll run the preprocess step for you so that you don't have to.

assets/question_sets/strings.js
```diff
@@ -26,5 +26,5 @@ globalData.questionSets['strings'] = {  // eslint-disable-line dot-notation
       "I will provide you with feedback."
     ].join('')
   ],
-  questionIds: ['reverseWords', 'parens', 'i18n', 'rle']
+  questionIds: ['reverseWords', 'parens', 'i18n', 'rle', 'stringToDict']
```
eyurko: Don't add anything to strings, just yet.

assets/questions/stringToDict.js
alblin: addressed in commit c2dac98. now using inputFunctionName, which returns a 2-item list (item 1 being string to parse, item 2 being the set of valid words) that is passed into the function the student is to implement

assets/questions/stringToDict.js
```diff
@@ -0,0 +1,117 @@
...
+// distributed under the License is distributed on an "AS-IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+/**
+ * @fileoverview Question data for String to Dict.
+ */
+
+globalData.questions['stringToDict'] = {  // eslint-disable-line dot-notation
```
eyurko: Your question title and function name aren't synced up -- are you splitting a string into words or converting it into a dictionary?

assets/questions/stringToDict.js
```diff
@@ -0,0 +1,117 @@
...
+  title: 'Split String to Words',
+  starterCode: {
+    python:
+`def stringToDict(lst):
```
eyurko: If you're trying to avoid a Python reserved word, I'd just use str

assets/questions/stringToDict.js
```diff
@@ -0,0 +1,117 @@
...
+      allowedOutputs: ['No Solution']
+    }],
+    buggyOutputTests: [],
+    performanceTests: []
+  }],
+  styleTests: [{
+    evaluationFunctionName: 'allowOnlyOneFunction',
+    expectedOutput: true,
+    message: [
+      'You should only be writing code in a words function. While ',
```
eyurko: "a words function"?

assets/questions/stringToDict.js
alblin: The input parameter is actually a list, the first element being the string to parse and the second element being the set of valid words. I explain this to the user in the question description. I'm not sure how to use inputFunctionName to allow the stringToDict to take in two parameters, so I made it a list of length two instead.

assets/questions/stringToDict.js
alblin: resolved in commit below - accidentally left that in when I was testing

assets/questions/stringToWords.js
```diff
@@ -0,0 +1,117 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+/**
+ * @fileoverview Question data for String to Dict.
+ */
+
+globalData.questions['stringToWords'] = {  // eslint-disable-line dot-notation
+  title: 'Split String to Words',
```
eyurko: Split String into Words

assets/questions/stringToWords.js
```diff
@@ -0,0 +1,117 @@
...
+`class AuxiliaryCode(object):
+    @classmethod
+    def returnOriginalString(cls, s):
+        return s
+
+    @classmethod
+    def processLongStringToStringAndDict(cls, s):
+      spl = s.split(":")
+      dict_str = spl[1].split(" ")
+      dict = set()
```
eyurko: This isn't a dict, then -- it's a set. Please update the names accordingly.

assets/questions/stringToWords.js
```diff
@@ -0,0 +1,117 @@
...
+      for el in dict_str:
```
eyurko: el?

assets/questions/stringToWords.js
```diff
@@ -0,0 +1,117 @@
...
+  auxiliaryCode: {
+    python:
+`class AuxiliaryCode(object):
+    @classmethod
+    def returnOriginalString(cls, s):
+        return s
+
+    @classmethod
+    def processLongStringToStringAndDict(cls, s):
+      spl = s.split(":")
```
eyurko: split_string. Feel free to specify your variable names so that they're as readable as possible -- we don't need to go for brevity, here.

assets/questions/stringToWords.js
alblin: comments addressed in commit below. Thank you!

assets/questions/stringToWords.js
```diff
@@ -0,0 +1,117 @@
...
+      set_string = split_string[1].split(" ")
+      set_object = set()
+      for word in set_string:
+        dict.add(word)
+      return [split_string[0], set_object]
+`
+  },
+  tasks: [{
+    instructions: [
+      [
```
eyurko: These lines are too long, I think. Keep lines to ~80.

assets/questions/stringToWords.js
```diff
@@ -0,0 +1,117 @@
...
+  styleTests: [{
+    evaluationFunctionName: 'allowOnlyOneFunction',
+    expectedOutput: true,
+    message: [
+      'You should only be writing code in the stringToWords function. While ',
+      "decomposition is generally a good idea, you shouldn't need more than ",
+      'just this function for this question.'
+    ].join('')
+  }]
+};
```
eyurko: Add a newline at the end of files.

alblin:
@eyurko thanks for the reviews! should i squash all these commits into one for the change?

assets/questions/stringToWords.js
```diff
@@ -0,0 +1,123 @@
...
+  tasks: [{
+    instructions: [
+      [
+        'For this question, you\'ll implement the stringToWords function. ',
+        'This function takes a list of two elements, the first element ',
+        ' being a string of alphabetic letters with no whitespaces and the ',
+        'second element being a set of valid words represented as a ',
+        'Python set. The function should take the input string and split it ',
+        'into two space-separated words that are both in the given ',
+        'set. If there is no valid solution, return \'No Solution\'.'
```
eyurko: Why not just return None?

alblin: changed No Solution to None

alblin:
Thanks for your help eric!

PR #226, opened 2017-04-18
Fix #213: Enable responsive autosave for user

### LINKED ISSUE #213: Make auto-save responsive to use interaction
#209 implements auto-save but it simply runs auto-save every 30seconds and is not based on user interactions.

Rather than saving every 30seconds (or any other static interval), auto-save should be activated 3 seconds after an edit event with saves set 8 seconds apart (that is "Saving code..." displayed for 3 seconds, then 5 second pause, then auto-save again). That is, if the user makes a quick change, it auto-saves after 3 seconds. If the user keeps typing non-stop for while, it saves every 8 seconds while the user types until the user finishes typing (it saves one last time 8 seconds after the user finishes typing the long code). If the user does not change anything, auto-save is not activated.

wmcalyj:
This change includes both ui and backend change.
I also added some unit test for LearnerViewDirective (See LearnerViewDirectiveSpec)

### DIFF
--- a/client/components/LearnerViewDirective.js
+++ b/client/components/LearnerViewDirective.js
@@ -422,7 +422,7 @@ tie.directive('learnerView', [function() {
         ];
 
         var autosaveCancelPromise;
-        var localCodeCache;
+        var codeCache;
         var congratulatoryFeedback = FeedbackObjectFactory.create();
         QuestionDataService.initCurrentQuestionSet(questionSetId);
         var questionSet = QuestionDataService.getCurrentQuestionSet(
@@ -445,11 +445,11 @@ tie.directive('learnerView', [function() {
           question = QuestionDataService.getQuestion(questionId);
           tasks = question.getTasks();
           currentTaskIndex = 0;
-          localCodeCache =
+          codeCache =
             CodeStorageService.loadStoredCode(questionId, language);
           $scope.title = question.getTitle();
-          $scope.code = localCodeCache ?
-              localCodeCache : question.getStarterCode(language);
+          $scope.code = codeCache ?
+              codeCache : question.getStarterCode(language);
           $scope.instructions = tasks[currentTaskIndex].getInstructions();
           $scope.previousInstructions = [];
           $scope.nextButtonIsShown = false;
@@ -572,7 +572,7 @@ tie.directive('learnerView', [function() {
                 ).then(setFeedback);
             }, DURATION_MSEC_WAIT_FOR_SCROLL);
           }, 0);
-          storeCodeAndUpdateLocalCodeCache(
+          storeCodeAndUpdateCodeCache(
             $scope.questionIds[$scope.currentQuestionIndex], code, language);
         };
 
@@ -595,13 +595,13 @@ tie.directive('learnerView', [function() {
             autosaveCancelPromise = $interval(function() {
               var currentQuestionId =
                 $scope.questionIds[$scope.currentQuestionIndex];
-              if (angular.equals(localCodeCache, $scope.code)) {
+              if (angular.equals(codeCache, $scope.code)) {
                 // No code change, stop autosave loop.
                 stopAutosave();
               } else {
                 // Code change detected, notify user, save code,
                 // update code cache and continue this loop.
-                storeCodeAndUpdateLocalCodeCache(
+                storeCodeAndUpdateCodeCache(
                   currentQuestionId, $scope.code, language);
                 triggerAutosaveNotification(DISPLAY_AUTOSAVE_TEXT_SECONDS);
               }
@@ -614,10 +614,10 @@ tie.directive('learnerView', [function() {
           $interval.cancel(autosaveCancelPromise);
         };
 
-        var storeCodeAndUpdateLocalCodeCache = function(
+        var storeCodeAndUpdateCodeCache = function(
           questionId, code, lang) {
           CodeStorageService.storeCode(questionId, code, lang);
-          localCodeCache = code;
+          codeCache = code;
         };
 
         loadQuestion(
--- a/client/components/LearnerViewDirective.js
+++ b/client/components/LearnerViewDirective.js
@@ -63,7 +63,7 @@ tie.directive('learnerView', [function() {
                   <div class="tie-coding-terminal">
                     <ui-codemirror ui-codemirror-opts="codeMirrorOptions"
                         ng-model="code"
-                        ng-change="activateAutosaving()"
+                        ng-change="autosave()"
                         class="tie-codemirror-container"></ui-codemirror>
                   </div>
                   <select class="tie-lang-select-menu" name="lang-select-menu">
@@ -422,7 +422,7 @@ tie.directive('learnerView', [function() {
         ];
 
         var autosaveCancelPromise;
-        var codeCache;
+        var cachedCode;
         var congratulatoryFeedback = FeedbackObjectFactory.create();
         QuestionDataService.initCurrentQuestionSet(questionSetId);
         var questionSet = QuestionDataService.getCurrentQuestionSet(
@@ -445,11 +445,11 @@ tie.directive('learnerView', [function() {
           question = QuestionDataService.getQuestion(questionId);
           tasks = question.getTasks();
           currentTaskIndex = 0;
-          codeCache =
+          cachedCode =
             CodeStorageService.loadStoredCode(questionId, language);
           $scope.title = question.getTitle();
-          $scope.code = codeCache ?
-              codeCache : question.getStarterCode(language);
+          $scope.code = cachedCode ?
+              cachedCode : question.getStarterCode(language);
           $scope.instructions = tasks[currentTaskIndex].getInstructions();
           $scope.previousInstructions = [];
           $scope.nextButtonIsShown = false;
@@ -572,7 +572,7 @@ tie.directive('learnerView', [function() {
                 ).then(setFeedback);
             }, DURATION_MSEC_WAIT_FOR_SCROLL);
           }, 0);
-          storeCodeAndUpdateCodeCache(
+          storeCodeAndUpdateCachedCode(
             $scope.questionIds[$scope.currentQuestionIndex], code, language);
         };
 
@@ -589,19 +589,19 @@ tie.directive('learnerView', [function() {
           }, displaySeconds * SECONDS_TO_MILLISECONDS);
         };
 
-        $scope.activateAutosaving = function() {
+        $scope.autosave = function() {
           if (!$scope.autosaveOn) {
             $scope.autosaveOn = true;
             autosaveCancelPromise = $interval(function() {
               var currentQuestionId =
                 $scope.questionIds[$scope.currentQuestionIndex];
-              if (angular.equals(codeCache, $scope.code)) {
+              if (angular.equals(cachedCode, $scope.code)) {
                 // No code change, stop autosave loop.
                 stopAutosave();
               } else {
                 // Code change detected, notify user, save code,
                 // update code cache and continue this loop.
-                storeCodeAndUpdateCodeCache(
+                storeCodeAndUpdateCachedCode(
                   currentQuestionId, $scope.code, language);
                 triggerAutosaveNotification(DISPLAY_AUTOSAVE_TEXT_SECONDS);
               }
@@ -614,10 +614,10 @@ tie.directive('learnerView', [function() {
           $interval.cancel(autosaveCancelPromise);
         };
 
-        var storeCodeAndUpdateCodeCache = function(
+        var storeCodeAndUpdateCachedCode = function(
           questionId, code, lang) {
           CodeStorageService.storeCode(questionId, code, lang);
-          codeCache = code;
+          cachedCode = code;
         };
 
         loadQuestion(
--- a/client/components/LearnerViewDirective.js
+++ b/client/components/LearnerViewDirective.js
@@ -80,7 +80,7 @@ tie.directive('learnerView', [function() {
                     Reset Code
                   </button>
                   <div class="tie-code-auto-save" ng-show="autosaveTextIsDisplayed">
-                    Saving code now...
+                    Saving code...
                   </div>
                   <button class="tie-run-button"
                       ng-class="{'active': !nextButtonIsShown}"
@@ -397,11 +397,13 @@ tie.directive('learnerView', [function() {
     controller: [
       '$scope', '$interval', '$timeout', 'SolutionHandlerService',
       'QuestionDataService', 'LANGUAGE_PYTHON', 'FeedbackObjectFactory',
-      'CodeStorageService',
+      'CodeStorageService', 'SECONDS_TO_MILLISECONDS', 'DEFAULT_AUTOSAVE_SECONDS',
+      'DISPLAY_AUTOSAVE_TEXT_SECONDS',
       function(
           $scope, $interval, $timeout, SolutionHandlerService,
           QuestionDataService, LANGUAGE_PYTHON, FeedbackObjectFactory,
-          CodeStorageService) {
+          CodeStorageService, SECONDS_TO_MILLISECONDS, DEFAULT_AUTOSAVE_SECONDS,
+          DISPLAY_AUTOSAVE_TEXT_SECONDS) {
         var DURATION_MSEC_WAIT_FOR_SCROLL = 20;
         var language = LANGUAGE_PYTHON;
         // TODO(sll): Generalize this to dynamically select a question set
@@ -419,14 +421,8 @@ tie.directive('learnerView', [function() {
           {themeName: 'Dark'}
         ];
 
-        var SECONDS_TO_MILLISECONDS = 1000;
-        // Default time interval, in seconds, after which code will
-        // be auto-saved.
-        var DEFAULT_AUTOSAVE_SECONDS = 5;
-        // "Saving code..." will last for 1 seconds and disappear.
-        var DISPLAY_AUTOSAVE_TEXT_SECONDS = 1;
-
         var autosaveCancelPromise;
+        var localCodeCache;
         var congratulatoryFeedback = FeedbackObjectFactory.create();
         QuestionDataService.initCurrentQuestionSet(questionSetId);
         var questionSet = QuestionDataService.getCurrentQuestionSet(
@@ -449,11 +445,11 @@ tie.directive('learnerView', [function() {
           question = QuestionDataService.getQuestion(questionId);
           tasks = question.getTasks();
           currentTaskIndex = 0;
-          var storedCode =
+          localCodeCache =
             CodeStorageService.loadStoredCode(questionId, language);
           $scope.title = question.getTitle();
-          $scope.code = storedCode ?
-              storedCode : question.getStarterCode(language);
+          $scope.code = localCodeCache ?
+              localCodeCache : question.getStarterCode(language);
           $scope.instructions = tasks[currentTaskIndex].getInstructions();
           $scope.previousInstructions = [];
           $scope.nextButtonIsShown = false;
@@ -560,7 +556,6 @@ tie.directive('learnerView', [function() {
           // their own code back if they click on the current question.
           CodeStorageService.storeCode(currentQuestionId,
             $scope.code, language);
-
           loadQuestion(questionId, questionSet.getIntroductionParagraphs());
         };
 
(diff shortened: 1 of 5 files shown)


client/components/LearnerViewDirective.js
```diff
@@ -384,9 +385,11 @@ tie.directive('learnerView', [function() {
         var SECONDS_TO_MILLISECONDS = 1000;
         // Default time interval, in seconds, after which code will
         // be auto-saved.
-        var DEFAULT_AUTOSAVE_SECONDS = 30;
-        // "Saving code now..." will last for 3 seconds and disappear.
-        var DISPLAY_AUTOSAVE_TEXT_SECONDS = 3;
+        var DEFAULT_AUTOSAVE_SECONDS = 5;
+        // "Saving code..." will last for 1 seconds and disappear.
```
eyurko: 1 second

client/components/LearnerViewDirective.js
```diff
@@ -543,21 +546,36 @@ tie.directive('learnerView', [function() {
...
-            CodeStorageService.storeCode(
-              currentQuestionId, $scope.code, language);
-          }, DEFAULT_AUTOSAVE_SECONDS * SECONDS_TO_MILLISECONDS);
+        $scope.activateAutosaving = function() {
+          if (!$scope.autosaveOn) {
+            $scope.autosaveOn = true;
+            autosaveCancelPromise = $interval(function() {
+              var currentQuestionId =
+                $scope.questionIds[$scope.currentQuestionIndex];
+              var storedCode = CodeStorageService.loadStoredCode(
```
eyurko: Is there any concern that loading the code every time will cause any responsiveness issues / lag for the user?

client/components/LearnerViewDirective.js
wmcalyj: I think it's already 1 seconds?

client/components/LearnerViewDirective.js
wmcalyj: I don't think so. it's just trying to get the code from localStorage to be compared. But I might be wrong. @rabidbit

client/components/LearnerViewDirective.js
eyurko: yes, I mean, "1 seconds" should be "1 second", in your comment.

client/components/LearnerViewDirectiveSpec.js
```diff
@@ -0,0 +1,201 @@
...
+    }
+    return generatedChars;
+  };
+
+  var NUM_CHARS_CODE = 100;
+  var LANGUAGE = 'python';
+
+  // Please make sure the following three variables are the same as
+  // what we have in LearnerViewDirective
+  var SECONDS_TO_MILLISECONDS = 1000;
```
eyurko: Can you make these tie constants and then just pull them into both LearnerViewDirective and the test? It seems like a bad idea to just have a comment asking politely to keep them consistent.

client/components/LearnerViewDirectiveSpec.js
```diff
@@ -0,0 +1,201 @@
...
+    DEFAULT_AUTOSAVE_SECONDS * SECONDS_TO_MILLISECONDS;
+
+  var AUTOSAVE_REPEAT_RANGE = 20;
+
+  beforeEach(inject(function($compile, $rootScope, _QuestionDataService_) {
+    $scope = $rootScope.$new();
+
+    // The reason why we have to go through this trouble to get $scope
+    // is the controller is anonymous, thus, there is no easy way to do
+    // it. Recommend future refactor.
```
eyurko: Add a TODO.

client/components/LearnerViewDirectiveSpec.js
```diff
@@ -0,0 +1,201 @@
...
+    $rootScope.$digest();
+    $scope.$digest();
+
+    QuestionDataService = _QuestionDataService_;
+
+  }));
+
+  describe("resetCode", function() {
+
+    it('should reset code to default code', function() {
```
eyurko: starter code

client/components/LearnerViewDirectiveSpec.js
```diff
@@ -0,0 +1,201 @@
...
+    beforeEach(inject(function(_$interval_, _$timeout_, _CodeStorageService_) {
+      $interval = _$interval_;
+      $timeout = _$timeout_;
+      CodeStorageService = _CodeStorageService_;
+      localStorage.clear();
+    }));
+
+    // Autosave text is using $timeout, thus,
+    // we need to flush both $interval and $timeout.
+    var timeFlies = function(timeToFly) {
```
eyurko: please just call this flushIntervalAndTimeout and the variable ... something informative.

client/components/LearnerViewDirectiveSpec.js
```diff
@@ -0,0 +1,201 @@
...
+
+    it("should activate autosaving for only once", function() {
+      for (var i = 0; i < $scope.questionIds.length; i++) {
+        if (i !== 0) {
+          // Default question index is 0, we don't want to
+          // navigate to question 0 again because navigateToQuestion
+          // will store code for current question, thus, fail the test
+          $scope.navigateToQuestion(i);
+        }
+        expect(!$scope.autosaveOn).toBe(true);
```
eyurko: why not just expect($scope.autosaveOn).toBe(false);?

client/components/LearnerViewDirectiveSpec.js
```diff
@@ -0,0 +1,201 @@
...
+      $interval.flush(timeToFly);
+      $timeout.flush(timeToFly);
+    };
+
+    it("should activate autosaving for only once", function() {
+      for (var i = 0; i < $scope.questionIds.length; i++) {
+        if (i !== 0) {
+          // Default question index is 0, we don't want to
+          // navigate to question 0 again because navigateToQuestion
+          // will store code for current question, thus, fail the test
```
eyurko: End sentences with periods --

"because navigateToQuestion will already have stored code for that question and fail the test."

client/components/LearnerViewDirectiveSpec.js
```diff
@@ -0,0 +1,201 @@
...
+          $scope.navigateToQuestion(i);
+        }
+        expect(!$scope.autosaveOn).toBe(true);
+        var questionId = $scope.questionIds[i];
+        // There should be no code stored in local storage
```
eyurko: localStorage

client/components/LearnerViewDirectiveSpec.js
```diff
@@ -0,0 +1,201 @@
...
+        // before autosave is triggerred.
+        expect(CodeStorageService.loadStoredCode(
```
eyurko: What is going on with this entire section?

client/components/LearnerViewDirectiveSpec.js
```diff
@@ -0,0 +1,201 @@
...
+          // navigate to question 0 again because navigateToQuestion
+          // will store code for current question, thus, fail the test
+          $scope.navigateToQuestion(i);
+        }
+        // Repeat 1 - 20 times
+        var repeatTimes = Math.floor(Math.random() * AUTOSAVE_REPEAT_RANGE) + 1;
+        expect(!$scope.autosaveOn).toBe(true);
+        var questionId = $scope.questionIds[i];
+        // There should be no code stored in local storage
+        // before autosave is triggerred.
```
eyurko: triggered

client/components/LearnerViewDirectiveSpec.js
```diff
@@ -0,0 +1,201 @@
...
+    }));
+
+    // Autosave text is using $timeout, thus,
+    // we need to flush both $interval and $timeout.
+    var timeFlies = function(timeToFly) {
+      $interval.flush(timeToFly);
+      $timeout.flush(timeToFly);
+    };
+
+    it("should activate autosaving for only once", function() {
```
eyurko: should only activate autosave once

client/components/LearnerViewDirectiveSpec.js
```diff
@@ -0,0 +1,201 @@
...
+        timeFlies(4 * SECONDS_TO_MILLISECONDS);
+        expect($scope.autosaveTextIsDisplayed).toBe(false);
+        expect($scope.autosaveOn).toBe(false);
+        expect(CodeStorageService.loadStoredCode(
+          questionId, LANGUAGE)).toEqual(starterCode);
+      }
+    });
+
+    it("should store the latest code into localStorage", function() {
+      for (var i = 0; i < $scope.questionIds.length; i++) {
```
eyurko: These seem pretty similar -- could you write a helper function that does both things?

client/components/LearnerViewDirective.js
wmcalyj: I see. Done

client/components/LearnerViewDirectiveSpec.js
wmcalyj: Done

client/components/LearnerViewDirectiveSpec.js
wmcalyj: Alright...

wmcalyj: Change variable to flushIntervalAndTimeout. Change parameter to timeToFlush.

client/components/LearnerViewDirectiveSpec.js
wmcalyj: Done

client/components/LearnerViewDirectiveSpec.js
wmcalyj: Done

client/components/LearnerViewDirectiveSpec.js
wmcalyj: It's because there is no $scope.autosaveOn ($scope.autosaveOn === undefined) if autosave is never triggered.

client/components/LearnerViewDirective.js
rabidbit: I think I'm just repeating what Eric is saying, but could the storedCode initial declaration be done outside activateAutosaving? Then set storedCode = $scope.code when storedCode != $scope.code?

client/components/LearnerViewDirectiveSpec.js
wmcalyj: Done

client/components/LearnerViewDirectiveSpec.js
wmcalyj: This section is to test when the user makes only one change, what will happen.
The change should trigger autosave, and 5 seconds later, autosave text should be displayed for 1 second. Then, the autosave text should be gone while the autosave is still running. Then, at 10 seconds, autosave will compare the code again, since they are the same, autosave will stop. Thus, at 10.1 seconds, autosave should be canceled, thus, autosaveOn should be false.

client/components/LearnerViewDirectiveSpec.js
wmcalyj: Honestly, I don't think I will need the first part where it's similar to the previous test. I put it there just to be safe I guess. What I really want to test here is, with continuous change, the autosave is working properly. I want to test that by checking the latest code is stored in localStorage. Thus, the initial status doesn't really matter here. Do you think I should just get rid of the similar part? @eyurko

client/components/LearnerViewDirectiveSpec.js
wmcalyj: Done

client/components/LearnerViewDirectiveSpec.js
wmcalyj: Done

client/components/LearnerViewDirective.js
wmcalyj: @rabidbit I can do that. But let me do some more tests about that since it's not a one line change. I actually need to update storedCode once currentQuestionId changes. (e.g. user navigates to next question)

client/components/LearnerViewDirectiveSpec.js
wmcalyj: Done

client/components/LearnerViewDirectiveSpec.js
wmcalyj: Get rid of the redundant testing part

client/components/LearnerViewDirective.js
wmcalyj: done

wmcalyj:
PTAL

client/components/LearnerViewDirective.js
```diff
@@ -447,11 +445,11 @@ tie.directive('learnerView', [function() {
           question = QuestionDataService.getQuestion(questionId);
           tasks = question.getTasks();
           currentTaskIndex = 0;
-          var storedCode =
+          localCodeCache =
```
eyurko: cachedCode, to go with the storedCode pattern

client/components/LearnerViewDirectiveSpec.js
```diff
@@ -0,0 +1,193 @@
...
+  var AUTOSAVE_REPEAT_RANGE = 20;
+
+  beforeEach(inject(function($compile, $rootScope, _QuestionDataService_,
+    _SECONDS_TO_MILLISECONDS_, _DEFAULT_AUTOSAVE_SECONDS_) {
+    $scope = $rootScope.$new();
+
+    // The reason why we have to go through this trouble to get $scope
+    // is the controller is anonymous, thus, there is no easy way to do
+    // it. Recommend future refactor.
+    // TODO
```
eyurko: That's not how a TODO works -- it needs a username and should go before the description of work that needs done.. Check other TODOs in the codebase and make sure you're following existing patterns.

client/components/LearnerViewDirectiveSpec.js
eyurko: You should include that as a set of comments, then, because that's not super clear to someone reading the code, and as a team supporting this code in the future we should make sure that the code is well-documented.

eyurko: It might not even be a bad idea to pull it out as a separate helper function so that you can comment it thoroughly without impacting the file's code flow.

client/components/LearnerViewDirectiveSpec.js
eyurko: I would say $scope.autosaveOn === undefined || $scope.autosaveOn, then. Reads cleaner.

client/components/LearnerViewDirective.js
wmcalyj: done

client/components/LearnerViewDirectiveSpec.js
wmcalyj: Done

wmcalyj:
PTAL

client/components/LearnerViewDirective.js
```diff
@@ -63,6 +63,7 @@ tie.directive('learnerView', [function() {
                   <div class="tie-coding-terminal">
                     <ui-codemirror ui-codemirror-opts="codeMirrorOptions"
                         ng-model="code"
+                        ng-change="activateAutosaving()"
```
eyurko: should this just be autosave()?

client/components/LearnerViewDirective.js
eyurko: you did codeCache instead of cachedCode. Please fix.

client/components/LearnerViewDirectiveSpec.js
```diff
@@ -0,0 +1,204 @@
...
+    var possible = "abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789";
+
+    for (var i = 0; i < number; i++) {
+      generatedChars += possible.charAt(
+        Math.floor(Math.random() * possible.length));
+    }
+    return generatedChars;
+  };
+
+  var NUM_CHARS_CODE = 100;
```
eyurko: Are these constants? Also, why all the weird spacing between them?

client/components/LearnerViewDirectiveSpec.js
```diff
@@ -0,0 +1,204 @@
...
+  var LANGUAGE = 'python';
+
+  var SECONDS_TO_MILLISECONDS;
+  var DEFAULT_AUTOSAVE_SECONDS;
+
+  var DELTA_MILLISECONDS = 100;
+
+  var AUTOSAVE_MILLISECONDS;
```
eyurko: This is formatted as a constant but doesn't have a value set. Is that intentional?

client/components/LearnerViewDirectiveSpec.js
```diff
@@ -0,0 +1,204 @@
...
+        expect(CodeStorageService.loadStoredCode(
+          questionId, LANGUAGE)).toEqual(null);
+        $scope.activateAutosaving();
+        checkActivateAutosaveOnceDetail(questionId, starterCode);
+        expect(CodeStorageService.loadStoredCode(
+          questionId, LANGUAGE)).toEqual(starterCode);
+      }
+    });
+
+    // This helper function is to test when autosave is triggered, what will
```
eyurko: Could you clean up a bit of the grammar / language, here? This comment is pretty verbose.

client/components/LearnerViewDirectiveSpec.js
```diff
@@ -0,0 +1,204 @@
...
+    // happen. After it's triggered, 5 seconds later, autosave text should be
+    // displayed for 1 second. Then, the autosave text should be gone while the
+    // autosave is still running. Then, at 10 seconds, autosave will compare the
+    // code again, since they are the same, autosave will stop. Thus, at 10.1
+    // seconds, autosave should be canceled, thus, autosaveOn should be false.
+    var checkActivateAutosaveOnceDetail = function(questionId, starterCode) {
```
eyurko: do you mean for this function to check to make sure that autosave is only activated once? 

If so, verifyAutosaveIsOnlyActivatedOnce

client/services/CodeStorageServiceSpec.js
```diff
@@ -77,12 +77,9 @@ describe('CodeStorageService', function() {
   describe('loadStoredCode', function() {
     it('should retrieve stored code from browser', function() {
       expect(localStorage.length).toEqual(0);
-      var hashKeys = [];
-      for (var i = 0; i < NUM_QUESTIONS; i++) {
-        hashKeys[i] = sampleQuestionIds[i] + ":" + LANGUAGE;
-        localStorage.setItem(hashKeys[i], sampleQuestionCodes[i]);
-      }
       sampleQuestionIds.forEach(function(questionId, index) {
+        var hashKey = questionId + ":" + LANGUAGE;
```
eyurko: It's not a hash, so just call it key.

client/components/LearnerViewDirectiveSpec.js
wmcalyj: It's updated in the before each since it requires tie constant

client/components/LearnerViewDirective.js
wmcalyj: Done

client/components/LearnerViewDirectiveSpec.js
wmcalyj: No, I am not trying to verify it's only activated once. What I am trying to verify here is, the text is displayed when it should be and the flag is correctly turned on/off. The reason why "once" is here is because of the timing.

client/services/CodeStorageServiceSpec.js
wmcalyj: Used name based on Sean's suggestion. Will do

client/components/LearnerViewDirectiveSpec.js
wmcalyj: Yes, they are.

wmcalyj:
PTAL

rabidbit:
This looks good from a UX perspective (responds and saves as I would expect). Deferring to Eric for coding/logic OK.

wmcalyj:
@rabidbit Thank you, John!

client/components/LearnerViewDirectiveSpec.js
```diff
@@ -0,0 +1,200 @@
...
+      // Flush 4000 milliseconds -- time: 10.1s
+      flushIntervalAndTimeout(4 * SECONDS_TO_MILLISECONDS);
+      expect($scope.autosaveTextIsDisplayed).toBe(false);
+      expect($scope.autosaveOn).toBe(false);
+    };
+
+    it("should store the latest code into localStorage", function() {
+      for (var i = 0; i < $scope.questionIds.length; i++) {
+        if (i !== 0) {
+          // Default question index is 0, we don't want to
```
eyurko: The default question index is 0, so we don't want to navigate to question 0 again as navigateToQuestion will already have stored code for the current question and will fail the test.

wmcalyj:
PTAL

PR #233, opened 2017-04-18
Fix #215: Question window enhancements

### LINKED ISSUE #215: Provision code and other data to question window
**Making the question window more intuitive**
As of now, the question window is just an array of strings being displayed as a `<p>` tag. We can make it more developer friendly by providing a means to add code and other resources like images or videos. 

screenshot from 2017-04-14 11 45 52
screenshot from 2017-04-14 11 48 48
screenshot from 2017-04-14 15 26 08

**Proposed solution**
The solution consists of the following two steps:
a) Data schema changes - Eric has proposed an architectural level change to the data schema. We wish to change the 'instructions' array in 'tasks' to a dictionary. 

Before:

```
instructions: [
	[
		'First, implement the abbreviate function. It takes a string as input ',
		'and returns an abbreviation of the string of the form ',
		'<first character><length of the middle of the string><last character> ',
		'but only if it shortens the input. ',
		'For example, "internationalization" should be abbreviated as "i18n".'
	].join('')
],
```

After:

```
instructions: [
	{
		'content': 'In this question, you will implement two functions.......',
		'type': 'text'
	},
	{
		'content': '    return answer.....',
		'type': 'code'
	},
],
```

This dictionary will help us in handling different parts of the question more elegantly. The UI in turn, has to take care of displaying the contents based on their types. These changes are non-trivial and will effect a lot of other components in the code.

b) UI changes - We need to find a way to make the UI intelligent enough to gracefully handle different types of instructions. The two proposed solutions by John are as follows:

i) Use JS to dynamically create the html tags based on the content type. for e.g. `<p>` for rendering text and `<pre>` for rendering code.
ii) use a common `<pre>` tag for all types and let angular handle the css using conditions

Things to do:
- [ ] Update the instructions data schema to a dictionary
- [ ] Modify the tests to handle the updated schema
- [ ] Fix all the components that have been flagged as broken by the updated tests
- [ ] Update the instructions for all the questions as per the data schema
- [ ] Update the design doc accordingly
- [ ] Research on finding the more appropriate way of handling the schema changes on the UI
- [ ] Build the UI to handle the updated data schema gracefully using angular conditions
- [ ] Test for any new issues and fix them

abinash-behera:
a) Updated the 'instructions' data schema for all the questions.
b) Updated the 'instructions' object validtions in TaskSchemaValidationService for asserting that each instruction is an associative array

screenshot from 2017-04-18 15 40 12

### DIFF
--- a/assets/questions/bestMeetupLocation.js
+++ b/assets/questions/bestMeetupLocation.js
@@ -46,19 +46,28 @@ globalData.questions['bestMeetupLocation'] = {  // eslint-disable-line dot-notat
   },
   tasks: [{
     instructions: [
-      [
-        'N people live in a city with streets that form a regular 2D grid. ',
-        'They start out at different intersections and want to meet up. ',
-        'They are only able to move along the roads (no crossing blocks diagonally etc.). ',
-        'Write a function bestMeetupLocation(people) that returns a meeting place that ',
-        'minimizes the total travel distance.'
-      ].join(''),
-      [
-        'The input is a list of strings indicating the coordinates of the people in a 2D grid. ',
-        'Your function should return a string indicating the best meeting point for N people. ',
-        'For instance, the best meeting point for 3 people at ',
-        '"0, 0", "4, 2" and "2, 4" is "2, 2".'
-      ].join('')
+      {
+        content: [
+          'N people live in a city with streets that form a regular 2D grid. ',
+          'They start out at different intersections and want to meet up. ',
+          'They are only able to move along the roads (no crossing blocks ',
+          'diagonally etc.). Write a function bestMeetupLocation(people) that ',
+          'returns a meeting place that minimizes the total travel distance.'
+        ].join(''),
+        type: 'text'
+      },
+      {
+        content: [
+          'The input is a list of strings indicating the coordinates of the ',
+          'people in a 2D grid. Your function should return a string ',
+          'indicating the best meeting point for N people.'
+        ].join(''),
+        type: 'text'
+      },
+      {
+        content: 'Input: "0, 0", "4, 2" and "2, 4"\nOutput: "2, 2"',
+        type: 'code'
+      }
     ],
     prerequisiteSkills: ['Math', 'Sorting'],
     acquiredSkills: ['Math', 'Sorting'],
@@ -96,10 +105,14 @@ globalData.questions['bestMeetupLocation'] = {  // eslint-disable-line dot-notat
     }]
   }, {
     instructions: [
-      [
-        'You might notice that some inputs have more than one possible "best" meeting point. ',
-        'Modify your bestMeetupLocation function to return a list of all of the best meeting points.'
-      ].join('')
+      {
+        content: [
+          'You might notice that some inputs have more than one possible ',
+          '"best" meeting point. Modify your bestMeetupLocation function to ',
+          'return a list of all of the best meeting points.'
+        ].join(''),
+        type: 'text'
+      }
     ],
     prerequisiteSkills: ['Math', 'Sorting'],
     acquiredSkills: ['Math', 'Sorting'],
@@ -125,9 +138,9 @@ globalData.questions['bestMeetupLocation'] = {  // eslint-disable-line dot-notat
     evaluationFunctionName: 'allowOnlyOneFunction',
     expectedOutput: true,
     message: [
-      'You should only be writing code in a bestMeetupLocation function. While ',
-      "decomposition is generally a good idea, you shouldn't need more than ",
-      'just this function for this question.'
+      'You should only be writing code in a bestMeetupLocation function. ',
+      'While decomposition is generally a good idea, you shouldn\'t need more',
+      'than just this function for this question.'
     ].join('')
   }]
 };
--- a/assets/questions/findChar.js
+++ b/assets/questions/findChar.js
@@ -40,14 +40,23 @@ globalData.questions['findFirstNonRepeatingCharacter'] = { // eslint-disable-lin
   tasks: [
     {
       instructions: [
-        [
-          'For this question, you will implement a ',
-          'findFirstNonRepeatingCharacter function. ',
-          'This function takes a string as input and returns ',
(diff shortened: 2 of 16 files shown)


client/components/LearnerViewDirective.js
```diff
@@ -96,12 +96,24 @@ tie.directive('learnerView', [function() {
               <h3 class="tie-question-title">{{title}}</h3>
               <div class="tie-previous-instructions">
                 <div ng-repeat="previousInstruction in previousInstructions track by $index">
-                  <p ng-repeat="paragraph in previousInstruction track by $index">{{paragraph}}</p>
+                  <div ng-repeat="instruction in previousInstruction track by $index">
+                    <p ng-if="instruction.type == 'text'">{{instruction.content}}</p>
+                    <div ng-if="instruction.type == 'example'">
+                      Example: <pre class="tie-question-example">{{instruction.content}}</pre>
```
rabidbit: I think we should be more flexible in what uses <pre> format. Meaning, remove the 'Example: ' in the line above.

Also means we probably want to name this type something more general than 'example'. Maybe 'preformatted' or 'codetext'

client/components/LearnerViewDirective.js
```diff
@@ -96,12 +96,24 @@ tie.directive('learnerView', [function() {
...
+                    </div>  
+                    <pre class="tie-question-code" ng-if="instruction.type == 'code'">{{instruction.content}}</pre>
```
rabidbit: Ignore my previous (not deleted) comment about 'example' type. Is there a reason we need an 'example' type? It appears it just appends 'Example:' prior to the preformatted. If that's the case, why not just have the content author write 'Example:' in the prior text type?

client/components/LearnerViewDirective.js
```diff
@@ -278,6 +290,28 @@ tie.directive('learnerView', [function() {
...
+        .tie-question-example {
+          background: rgb(242, 242, 242);
+          padding: 10px;
+          border: 1px solid #ccc;
+          white-space: pre-wrap;
+          white-space: -moz-pre-wrap;
+          white-space: -pre-wrap;
+          white-space: -o-pre-wrap;
+          word-wrap: break-word;  
+        }
```
rabidbit: Please order the css attributes in alphabetical order too, unless it impacts cascading.

rabidbit:
This looks fine to me from a design perspective. Added a couple of comments. Deferring to @eyurko  or @seanlip for feedback from data schema and general logic perspective.

assets/questions/bestMeetupLocation.js
```diff
@@ -46,19 +46,27 @@ globalData.questions['bestMeetupLocation'] = {  // eslint-disable-line dot-notat
...
-      [
-        'The input is a list of strings indicating the coordinates of the people in a 2D grid. ',
-        'Your function should return a string indicating the best meeting point for N people. ',
-        'For instance, the best meeting point for 3 people at ',
-        '"0, 0", "4, 2" and "2, 4" is "2, 2".'
-      ].join('')
+      {
+        content: [
+          'N people live in a city with streets that form a regular 2D grid. ',
+          'They start out at different intersections and want to meet up. ',
```
eyurko: Line length needs to be 80 characters for all lines.

assets/questions/findChar.js
```diff
@@ -40,14 +40,22 @@ globalData.questions['findFirstNonRepeatingCharacter'] = { // eslint-disable-lin
...
+            'For this question, you will implement a ',
+            'findFirstNonRepeatingCharacter function. ',
+            'This function takes a string as input and returns ',
+            'the first character that isn\'t repeated elsewhere in the string.'
+          ].join(''),
+          type: 'text'
+        },
+        {
+          content: [
+            'For instance, findFirstNonRepeatingCharacter("mom") would return "o". ',
```
eyurko: Will all this be in code font?

assets/tests/TaskSchemaValidationService.js
```diff
@@ -38,11 +38,13 @@ tie.factory('TaskSchemaValidationService', [
         var instructions = task.getInstructions();
         return instructions.length > 0;
       },
-      verifyInstructionsIsArrayOfStrings: function(task) {
+      verifyInstructionsIsAssociativeArray: function(task) {
```
eyurko: You should also verify that the types are within the types that you allow, not just type: "HAADASD" or whatever.

client/components/LearnerViewDirective.js
eprym: You can refer to #215 about the limitation of the current text type. Basically speaking, it's not user-friendly to embed other format of instructions in the question window.

assets/questions/bestMeetupLocation.js
abinash-behera: As a side note, it would be a good idea to add the '[max-len](http://eslint.org/docs/rules/max-len)' rule to our ESLint package. I have raised an [issue #236](https://github.com/google/tie/issues/236) for the same.

client/components/LearnerViewDirective.js
rabidbit: What I mean is, why not just have types 'text' and 'code' (instead of 'text', 'code', and 'example')? I don't see the value in having an 'example' type when it appears the only difference between 'example' and 'code' is that the text 'Example:' is prepended.

assets/tests/TaskSchemaValidationService.js
abinash-behera: Fixed.

client/components/LearnerViewDirective.js
abinash-behera: Fixed.

client/components/LearnerViewDirective.js
abinash-behera: Fixed.

assets/questions/findChar.js
abinash-behera: Changed the sentence to a concrete example.

rabidbit:
LGTM from a UX perspective.

rabidbit:
I recommend removing seanlip from this review.

Issue #37, opened 2017-03-09
Show better error message if user deletes or modifies starter code.


eyurko:
Repro steps:

1. Open app.html
2. Navigate to the i18n problem.
3. Change "abbreviate" to "isBalanced"
4. Run code.
5. You should see this error: ["Your code threw an error: AttributeError: 'StudentCode' object has no attribute 'abbreviate' on line 42"]

It might be worth catching that error and saying something closer to "You seem to be missing an 'abbreviate' method. Check the starter code and make sure that it hasn't been deleted or modified accidentally."

eyurko:
Another solid error message would be "You seem to have modified the starter code -- here it is again, if you need it:

{STARTER CODE}"

That way they can copy-paste it back in.

seanlip:
Hm, this actually feels like an extension of the style tests we were considering. Maybe we should look into fleshing that out and implementing it.

seanlip:
Fixed in #237, thanks @a-zakem! Closing.

PR #237, opened 2017-04-20
Pre requisites

### LINKED ISSUE #100: No message to warn against library imports
Reproduction Steps:
1. Open "app.html"

2. Navigate to a question

3. Add a library import like `import string` or `import re`

4. Use the import in the function like `word[0] in string.whitespace` or `re.compile(r'\w')`

5. Click Run

Current message is:
`NameError: name 'string' is not defined on line 3 `

Which could be:
`Library imports are not permitted for this problem`

screen shot 2017-04-04 at 10 15 23 am

Standard library imports like re and string may not be allowed as they make the problem solution too simple and unchallenging. However, it might be a god idea to say that explicitly.

[EDIT Mar 4, 11:24]
Suggested Solutions:
**Option 1: Ban the use all libraries for all questions**
This is a blanket solution. For each solution to any problem, check for an import statement. If there is one, don't compile the problem. Instead, show a message saying imports are not permitted.
This is how the system is currently designed and would only require adding a check and a message. This can be a good minimal solution.

**Option 2: Ban libraries specific to questions**
We could have two sets of libraries - banned in every question e.g. `pandas` and banned for this specific question e.g. `re` for string manipulation questions. This is a more elegant and sustainable solution and requires a little more work than the first option

a-zakem:
Fix #100 #37 

Adding PreRequisite checks.  Check if user has modified/deleted starter code or tried to import external libraries, and give appropriate feedback.


client/services/code_evaluators/PreRequisiteCheckDispatcherService.js
```diff
@@ -0,0 +1,34 @@
...
+      checkCode: function(language, starterCode, code) {
+        if (language === LANGUAGE_PYTHON) {
+          return PythonPreRequisiteCheckService.checkCode(starterCode, code);
+        } else {
+          throw Error('Language not supported: ' + language);
+        }
+      }
+    };
+  }
+]);
```
eyurko: Add newline

client/domain/PreRequisiteEvalResultObjectFactory.js
```diff
@@ -0,0 +1,52 @@
...
+  function() {
+    var PreRequisiteEvalResult = function(
+      code, preReqErrorMessage, starterCode) {
+      this._code = code;
+      this._preReqErrorMessage = preReqErrorMessage;
+      this._starterCode = starterCode;
+    };
+
+    // Instance methods
+    PreRequisiteEvalResult.prototype.getCode = function() {
```
eyurko: If you're also using the CodeEvalResult, do you think you need to store the code here, too? Or maybe store the code on the snapshot and just have the eval results contain error messages? I'm just spitballing, here.

client/services/FeedbackGeneratorService.js
```diff
@@ -221,6 +221,23 @@ tie.factory('FeedbackGeneratorService', [
         feedback.appendCodeParagraph(errorString);
         return feedback;
       },
+      getPreRequisiteFailureFeedback: function(preReqEvalResult) {
+        var feedback = FeedbackObjectFactory.create(false);
+        var errorMessage =
+          preReqEvalResult.getPreReqErrorMessage();
+        feedback.appendTextParagraph(
+          "Oops!  There is a problem with your code:");
```
eyurko: I wouldn't say "Oops!"; it's a bit too informal. Go for "It looks like there's a problem with your code: "

client/services/SolutionHandlerService.js
```diff
@@ -19,53 +19,74 @@
...
-        return CodeRunnerDispatcherService.runCodeAsync(
-          language, studentCode
-        ).then(function(rawCodeEvalResult) {
-          var potentialSyntaxErrorString = rawCodeEvalResult.getErrorString();
-          if (potentialSyntaxErrorString) {
-            var feedback = FeedbackGeneratorService.getSyntaxErrorFeedback(
-              potentialSyntaxErrorString);
+          task, starterCode, studentCode, auxiliaryCode, language) {
+        // First, check pre-requisites for the submitted code
+        return PreRequisiteCheckDispatcherService.checkCode(
```
eyurko: As a nit, I think it should be just Prerequisite, not PreRequisite

client/services/code_evaluators/PreRequisiteCheckDispatcherService.js
eyurko: Add newline

client/services/code_evaluators/PythonPreRequisiteCheckService.js
```diff
@@ -0,0 +1,123 @@
...
+      }
+      var codeLineIndex = 0;
+      for (var i = 0; i < starterCodeLines.length; i++) {
+        var starterCodeLine = starterCodeLines[i].trim();
+        // ignore blank lines in starter code
+        if (starterCodeLine.length === 0) {
+          continue;
+        }
+        //handle "return" statements in starter code
+        if (starterCodeLine.replace(/^\s+/,"").startsWith("return ")) {
```
eyurko: Always pull out regexes as constants at the start of the file. Makes it more obvious to the person reading your code.

client/services/code_evaluators/PythonPreRequisiteCheckService.js
```diff
@@ -0,0 +1,123 @@
...
+    };
+
+    return {
+      // Returns a promise.
+      checkCode: function(starterCode, code) {
+        if (!(checkStarterCodePresent(starterCode, code))) {
+          var errorMessage = ['It looks like you deleted or modified ',
+            'the starter code!  Our evaluation program requires the ',
+            'function names given in the starter code.  Here\'s the '
+            ,'starter code again:'].join('');
```
eyurko: I think we also have a reset button, now? Might be worth mentioning, if that's a thing that exists.

client/services/code_evaluators/PythonPreRequisiteCheckServiceSpec.js
```diff
@@ -0,0 +1,102 @@
...
+		it('does not capture strings with \'import\' in text', function() {
+			var code = ['def myFunction(arg): \n',
+				'\tresult = str(arg) + \'import this\'',
+				'\treturn result'].join('');
+			var codeLibs =
+				PythonPreRequisiteCheckService.getCodeLibs(code);
+			expect(codeLibs).toEqual([]);
+		});
+	});
+});
```
eyurko: Add newline at EOF

client/domain/PrerequisiteEvalResultObjectFactory.js
```diff
@@ -0,0 +1,52 @@
...
+ * @fileoverview Factory for creating new frontend instances of
+ * PreRequisiteCheckResult domain objects.
+ */
+
+tie.factory('PrerequisiteEvalResultObjectFactory', [
+  function() {
+    var PrerequisiteEvalResult = function(
+      code, prereqErrorMessage, starterCode) {
+      this._code = code;
+      this._prereqErrorMessage = prereqErrorMessage;
```
seanlip: This is a bit confusing. The purpose of this domain object would presumably be to store enough information to reconstruct the result of the prerequisite check.

So, presumably the only field here should be a list called something like this._prerequisiteCheckFailures. Each entry is a dict with a 'type' key (whose value is PREREQ_CHECK_TYPE_BAD_IMPORT = 'badImport' or PREREQ_CHECK_TYPE_MISSING_STARTER_CODE = 'missingStarterCode' -- these named constants can be declared in app.js). If the key is PREREQ_CHECK_TYPE_MISSING_STARTER_CODE then the dict also has an extra 'expectedStarterCode' key; you may need something similar for badImport as well.

The error messages themselves should presumably go in the service which generates the feedback objects based on the essential info in this domain object.

client/services/SolutionHandlerService.js
```diff
@@ -19,53 +19,74 @@
 
 tie.factory('SolutionHandlerService', [
   '$q', 'CodePreprocessorDispatcherService', 'CodeRunnerDispatcherService',
-  'FeedbackGeneratorService', 'SnapshotObjectFactory', 'TranscriptService',
-  'CodeSubmissionObjectFactory',
+  'FeedbackGeneratorService', 'PrerequisiteCheckDispatcherService',
+  'SnapshotObjectFactory', 'TranscriptService', 'CodeSubmissionObjectFactory',
```
seanlip: Nit: let's move 'CodeSubmissionObjectFactory' to the line below for parity with line 27. Easier to check that the dependencies line up, that way.

client/services/code_evaluators/PythonPrerequisiteCheckService.js
```diff
@@ -0,0 +1,126 @@
...
+tie.factory('PythonPrerequisiteCheckService', [
+  'PYTHON_STANDARD_LIBRARIES',
+  'PrerequisiteEvalResultObjectFactory',
+  function(
+      PYTHON_STANDARD_LIBRARIES,
+      PrerequisiteEvalResultObjectFactory) {
+
+    const LEFT_TRIM_PATTERN = /^\s+/;
+
+    var checkStarterCodePresent = function(starterCode, code) {
```
seanlip: This is far too complicated / ad hoc. It'll be hard to maintain.

So, just to verify: the point of this prereq check is to ensure that the functions in the starter code still exist within the learner's code, right? If so, could we instead restructure the logic in this function roughly as follows?

```
    expectedTopLevelFunctions = extractTopLevelFunctions(starterCode);
    verifyFunctionsExist(code, expectedTopLevelFunctions);
```

client/services/code_evaluators/PythonPrerequisiteCheckService.js
```diff
@@ -0,0 +1,126 @@
...
+          }
+        }
+        if (codeLineIndex >= codeLines.length && i < starterCodeLines.length) {
+          return false;
+        }
+      }
+      return true;
+    };
+
+    var getCodeLibs = function(code) {
```
seanlip: getImportedLibraries

client/services/code_evaluators/PythonPrerequisiteCheckService.js
```diff
@@ -0,0 +1,126 @@
...
+      var codeLines = code.split('\n');
+      var coddLibs = [];
```
seanlip: Spelling is wrong. Might want to use importedLibraries, anyway.

client/services/code_evaluators/PythonPrerequisiteCheckService.js
```diff
@@ -0,0 +1,126 @@
...
+      var pattern = new RegExp('^import\\ \\w+$');
+      for(var i=0; i<codeLines.length; i++) {
```
seanlip: I'm surprised that the linter isn't catching this, but there should be single spaces around = and '<', as well as before '('. Ditto elsewhere.

client/services/code_evaluators/PythonPrerequisiteCheckService.js
```diff
@@ -0,0 +1,126 @@
...
+        if(pattern.test(codeLines[i])) {
+          var words = codeLines[i].split(' ');
+          coddLibs.push(words[1]);
```
seanlip: You should be able to get the pattern directly from the regexp check by using capture groups. I recommend doing that instead, it's much cleaner.

client/services/code_evaluators/PythonPrerequisiteCheckService.js
```diff
@@ -0,0 +1,126 @@
...
+        }
+      }
+      return coddLibs;
+    };
+
+    return {
+      // Returns a promise.
+      checkCode: function(starterCode, code) {
+        if (!(checkStarterCodePresent(starterCode, code))) {
+          var errorMessage = ['It looks like you deleted or modified ',
```
seanlip: This stuff should go in the service that creates the feedback messages that get displayed to the user. The object storing the prereq check result should only contain the error type and the minimal amount of info needed to reproduce what check actually failed.

client/services/code_evaluators/PythonPrerequisiteCheckService.js
```diff
@@ -0,0 +1,126 @@
...
+            PrerequisiteEvalResultObjectFactory.create(
+              code, errorMessage, starterCode));
+        }
+
+        var libErrorMessage = ['It looks like you are importing an external ',
+                        'library.  Only standard libraries are supported.  ',
+                        'The following libraries are not ',
+                        'supported:\n'].join('');
+        codeLibs = getCodeLibs(code);
+        var extLibs = [];
```
seanlip: Again, all this stuff is way too complicated. You could instead do something like this:

    var unsupportedImports = importedLibraries.filter(function(importedLibrary) {
      return PYTHON_STANDARD_LIBRARIES.indexOf(importedLibrary) === -1;
    });

Make sure that the importedLibraries list is already trimmed, etc. Then pass the unsupportedImports as a customization to the prereq check result factory.

client/services/code_evaluators/PythonPrerequisiteCheckServiceSpec.js
```diff
@@ -0,0 +1,102 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+/**
+ * @fileoverview Unit tests for the PythonPreRequisiteCheckService.
+ */
+
+describe('PythonPrerequisiteCheckService', function() {
+	var PythonPrerequisiteCheckService;
```
seanlip: Use spaces (and 2-space indentation), not tabs. Again I'm surprised the linter isn't catching this, I'll take a look. Is this synced with the latest version of master?

client/services/code_evaluators/PythonPrerequisiteCheckServiceSpec.js
```diff
@@ -0,0 +1,102 @@
...
+		it('correctly parses import statements', function() {
+			var code = ['import numpy\n',
+				'import pandas\n', '\n',
+				'def myFunction(arg): \n',
+				'\treturn arg'].join('');
+			var codeLibs =
+				PythonPrerequisiteCheckService.getCodeLibs(code);
+			expect(codeLibs).toEqual(['numpy','pandas']);
+		});
+		it('does not capture strings with \'import\' in text', function() {
```
seanlip: Nice test!

client/services/code_evaluators/PythonPrerequisiteCheckServiceSpec.js
```diff
@@ -0,0 +1,102 @@
...
+			function() {
+				var code = ['def myFunction(arg):\n',
+					'\tresult = arg.rstrip()\n',
+					'\treturn myResult\n'].join('');
+				var starterCodePresent =
+					PythonPrerequisiteCheckService.checkStarterCodePresent(
+						starterCode, code);
+				expect(starterCodePresent).toEqual(true);
+		});
+		it('returns false when starter code lines are present, but out of order',
```
seanlip: Eh, I'm not sure the order matters...

a-zakem:
Hi Sean,

I can confirm that the linter is not catching the style errors you pointed
out; I fetched from origin and merged in the master branch, but was still
able to push while receiving a "Linter tests passed" message.

I'll work on fixing the issues above.

--Aaron

seanlip:
Thanks, @a-zakem! OK, good to know -- I'll investigate the linter issues.

client/domain/SnapshotObjectFactory.js
```diff
@@ -19,13 +19,22 @@
...
+    var Snapshot = function(codePrereqCheckResult, codeEvalResult, feedback) {
+      this._codePrereqCheckResult = codePrereqCheckResult;
       this._codeEvalResult = codeEvalResult;
       this._feedback = feedback;
       this._timestamp = '';
     };
 
     // Instance methods.
+    Snapshot.prototype.getCodePrereqCheckResult = function() {
+      return this._prereqEvalResult;
```
seanlip: I think you forgot to update this.

client/domain/SnapshotObjectFactory.js
```diff
@@ -19,13 +19,22 @@
...
+    };
+
+    Snapshot.prototype.setCodePrereqCheckResult = function(prereqEvalResult) {
+      this._prereqEvalResult = prereqEvalResult;
```
seanlip: Ditto. The names are wrong.

client/services/FeedbackGeneratorService.js
```diff
@@ -222,6 +225,43 @@ tie.factory('FeedbackGeneratorService', [
         feedback.setSyntaxErrorIndex(feedback.getParagraphs().length - 1);
         return feedback;
       },
+      getPrereqFailureFeedback: function(codePrereqCheckResult) {
+        var feedback = FeedbackObjectFactory.create(false);
+        var preReqFailures = codePrereqCheckResult.getPrereqCheckFailures();
```
seanlip: Here and elsewhere: be careful of casing. "Prerequisite" is one word, and so this should be `prereqFailures` (or, for parity with the method name, `prereqCheckFailures`).

client/services/FeedbackGeneratorService.js
```diff
@@ -222,6 +225,43 @@ tie.factory('FeedbackGeneratorService', [
...
+
+        /* This function shouldn't be invoked unless there is at least one
```
seanlip: I don't understand. Wouldn't it make more sense to throw an Error in this case if we require the input arg to have at least one element?

client/services/FeedbackGeneratorService.js
```diff
@@ -222,6 +225,43 @@ tie.factory('FeedbackGeneratorService', [
...
+        pre-requisite check failure, but the code below is provided to prevent
+        an index out of bounds error. */
+        if (preReqFailures.length === 0) {
+          feedback.appendTextParagraph("Prerequisite checks passed!");
+          return feedback;
+        }
+
+        // Check first error type and generate appropriate feedback message.
+        var preReqFailure = preReqFailures[0];
+        if (preReqFailure.get('type') === PREREQ_CHECK_TYPE_MISSING_STARTER_CODE) {
```
seanlip: There ought to be a domain object for prereqFailure, and we should be able to call prereqFailure.isMissingStarterCode(), prereqFailure.isBadImport().

client/services/FeedbackGeneratorService.js
```diff
@@ -222,6 +225,43 @@ tie.factory('FeedbackGeneratorService', [
...
+          return feedback;
+        } else if (preReqFailure.get('type') === PREREQ_CHECK_TYPE_BAD_IMPORT) {
+          feedback.appendTextParagraph(['It looks like you are importing an ',
+            'external library.  Only standard libraries are supported.  ',
+            'The following libraries are not supported:\n'].join(''));
+          var badImports = preReqFailure.get('badImports');
+          feedback.appendCodeParagraph(badImports.join('\n'));
+          return feedback;
+        } else {
+          /* If pre-requisite check type is not handled above, generate generic
```
seanlip: No -- throw an Error.

client/services/FeedbackGeneratorService.js
```diff
@@ -222,6 +225,43 @@ tie.factory('FeedbackGeneratorService', [
...
+
+        // Check first error type and generate appropriate feedback message.
+        var preReqFailure = preReqFailures[0];
+        if (preReqFailure.get('type') === PREREQ_CHECK_TYPE_MISSING_STARTER_CODE) {
+          feedback.appendTextParagraph(['It looks like you deleted or ',
+            ' modified the starter code!  Our evaluation program requires the ',
+            'function names given in the starter code.  You can press the ',
+            '\'Reset Code\' button below to start over.  Or, you can copy ',
+            'the starter code below:'].join(''));
+          feedback.appendCodeParagraph(preReqFailure.get('starterCode'));
```
seanlip: Should be prereqFailure.getStarterCode(), and similarly for badImports.

client/services/code_evaluators/PythonPrereqCheckService.js
```diff
@@ -0,0 +1,119 @@
...
+      PREREQ_CHECK_TYPE_BAD_IMPORT, PREREQ_CHECK_TYPE_MISSING_STARTER_CODE,
+      PYTHON_STANDARD_LIBRARIES) {
+
+    var extractTopLevelFunctions = function(starterCode) {
+      var starterCodeLines = starterCode.split('\n');
+      var topLevelFunctions = [];
+      for (var i = 0; i < starterCodeLines.length; i++) {
+        var line = starterCodeLines[i].trim();
+        if (line.startsWith("def ")) {
+          topLevelFunctions.push(line);
```
seanlip: Let's call this topLevelFunctionLines, otherwise it looks like it's storing just the function names.

client/services/code_evaluators/PythonPrereqCheckService.js
```diff
@@ -0,0 +1,119 @@
...
+        }
+      }
+      return topLevelFunctions;
+    };
+
+    var verifyTopLevelFunctionsExist = function(
+      code, expectedTopLevelFunctions) {
```
seanlip: expectedTopLevelFunctionLines (similarly)

client/services/code_evaluators/PythonPrereqCheckService.js
```diff
@@ -0,0 +1,119 @@
...
+      var codeLines = code.split('\n');
+      for (var i = 0; i < codeLines.length; i++) {
+        codeLines[i] = codeLines[i].trim();
+      }
+      for (var i = 0; i < expectedTopLevelFunctions.length; i++) {;
```
seanlip:
```
return expectedTopLevelFunctionLines.every(function(expectedLine) {
  return codeLines.indexOf(expectedLine) !== -1;
});

File: client/services/code_evaluators/PythonPrereqCheckService.js [hunk 30]
```diff
@@ -0,0 +1,119 @@
...
+        }
+      }
+      return topLevelFunctions;
+    };
+
+    var verifyTopLevelFunctionsExist = function(
+      code, expectedTopLevelFunctions) {
+      var codeLines = code.split('\n');
+      for (var i = 0; i < codeLines.length; i++) {
+        codeLines[i] = codeLines[i].trim();
```
seanlip: Why trim, btw? I thought we'd want no whitespace at the start, otherwise we risk matching inner functions.

client/services/code_evaluators/PythonPrereqCheckService.js
```diff
@@ -0,0 +1,119 @@
...
+      for (var i = 0; i < starterCodeLines.length; i++) {
+        var line = starterCodeLines[i].trim();
+        if (line.startsWith("def ")) {
+          topLevelFunctions.push(line);
+        }
+      }
+      return topLevelFunctions;
+    };
+
+    var verifyTopLevelFunctionsExist = function(
```
seanlip: Nit: let's call this doTopLevelFunctionLinesExist()

(I'd tend to expect verify() to throw an exception if the check fails. The proposed new name makes it clearer that it returns a boolean.)

client/services/code_evaluators/PythonPrereqCheckService.js
```diff
@@ -0,0 +1,119 @@
...
+
+    var checkStarterCodePresent = function(starterCode, code) {
+      var expectedTopLevelFunctions = extractTopLevelFunctions(starterCode);
+      return verifyTopLevelFunctionsExist(code, expectedTopLevelFunctions);
+    };
+
+    var getImportedLibraries = function(code) {
+      var codeLines = code.split('\n');
+      var importedLibraries = [];
+      var importPattern = new RegExp('(^import\\ )(\\w+)($)');
```
seanlip: You don't really need some of these capture groups. I.e. you could do:

    new RegExp('^import\\ (\\w+)$');

client/services/code_evaluators/PythonPrereqCheckService.js
```diff
@@ -0,0 +1,119 @@
...
+          importedLibraries.push(match[2]);
+        }
+      }
+      return importedLibraries;
+    };
+
+    var getUnsupportedImports = function(importedLibraries) {
+      var unsupportedImports = importedLibraries.filter(function(
+        importedLibrary) {
+        return PYTHON_STANDARD_LIBRARIES.indexOf(importedLibrary) === -1;
```
seanlip: Indent this line by 2.

Also, I forgot to mention this earlier, but can we rename the constant to SUPPORTED_PYTHON_LIBRARIES or SUPPORTED_PYTHON_LIBS? PYTHON_STANDARD_LIBRARIES kind of implies "all standard libraries in python".

client/services/code_evaluators/PythonPrereqCheckService.js
```diff
@@ -0,0 +1,119 @@
...
+      );
+      return unsupportedImports;
+    };
+
+    return {
+      // Returns a promise.
+      checkCode: function(starterCode, code) {
+        // check that starter code is present
+        if (!(checkStarterCodePresent(starterCode, code))) {
+          var prereqCheckFailures = [{'type':
```
seanlip: There should be a domain object for the inner dicts -- don't use raw JS dicts to pass around the info.

client/services/code_evaluators/PythonPrereqCheckServiceSpec.js
```diff
@@ -0,0 +1,119 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+/**
+ * @fileoverview Unit tests for the PythonPrereqCheckService.
+ */
+
+describe('PythonPrereqCheckService', function() {
+	var PythonPrereqCheckService;
```
seanlip: As mentioned previously, please use spaces, not tabs. You can double-check how it looks at https://github.com/google/tie/pull/237/files#diff-8011e749cdc5745088f463a875d42ccdR19

client/services/code_evaluators/PythonPrereqCheckServiceSpec.js
```diff
@@ -0,0 +1,119 @@
...
+		});
+		it('returns false if starter code line is not found (deleted)',
+			function() {
+				var code = ['\tresult = arg.rstrip()\n',
+				'\treturn result\n'].join('');
+				var starterCodePresent =
+					PythonPrereqCheckService.checkStarterCodePresent(
+						starterCode, code);
+				expect(starterCodePresent).toEqual(false);
+		});
```
seanlip: I mentioned before that the test "it returns false when starter code lines are present, but out of order" seems odd, but I should have been clearer. I think this is a good test to include here, but we should instead check that it returns `true`.

seanlip:
Thank you, @a-zakem! On the whole, this is looking way better. I have another round of comments.

Also, one request: please could you reply to all comments? In particular, for comments that you've addressed, please reply "Done." -- this helps keep track of which comments have been addressed and which haven't. (I think you missed one.)

client/domain/SnapshotObjectFactory.js
a-zakem: Fixed.

client/domain/SnapshotObjectFactory.js
a-zakem: Fixed.

client/services/FeedbackGeneratorService.js
a-zakem: Fixed.

client/services/FeedbackGeneratorService.js
a-zakem: Changed to throw an Error.

client/services/FeedbackGeneratorService.js
a-zakem: Added PrereqCheckFailureObjectFactory under client/domain with functions PrereqCheckFailure.isMissingStarterCode() and PrereqCheckFailure.isBadImport().

client/services/FeedbackGeneratorService.js
a-zakem: Added functions prereqCheckFailure.getStarterCode() and prereqCheckFailure.getBadImports() to PrereqCheckFailure object.

client/services/FeedbackGeneratorService.js
a-zakem: Changed to throw an Error.

client/services/code_evaluators/PythonPrereqCheckService.js
a-zakem: Changed.

client/services/code_evaluators/PythonPrereqCheckService.js
a-zakem: Changed.

client/services/code_evaluators/PythonPrereqCheckService.js
a-zakem: Changed.

client/services/code_evaluators/PythonPrereqCheckService.js
a-zakem: I mistakenly thought Javascript's trim() was equivalent to Python's rstrip().  I added a function rightTrim() to PythonPrereqCheckService which uses regular expressions and string replace to trim white space at the end of a string.    I'm using rightTrim() so that additional white space at the end of the line in either the user's code or starter code will not trigger a prereq failure for missing starter code function lines (although it should probably trigger a style warning in the future).

client/services/code_evaluators/PythonPrereqCheckService.js
a-zakem: Changed.

client/services/code_evaluators/PythonPrereqCheckService.js
a-zakem: Changed.

client/services/code_evaluators/PythonPrereqCheckService.js
a-zakem: Changed.

client/services/code_evaluators/PythonPrereqCheckService.js
a-zakem: Added PrereqCheckFailure domain object.

client/services/code_evaluators/PythonPrereqCheckServiceSpec.js
a-zakem: Changed.

client/services/code_evaluators/PythonPrereqCheckServiceSpec.js
a-zakem: Added test.

client/services/code_evaluators/PythonPrereqCheckService.js
```diff
@@ -0,0 +1,126 @@
...
+  'CodePrereqCheckResultObjectFactory', 'PrereqCheckFailureObjectFactory',
+  'PREREQ_CHECK_TYPE_BAD_IMPORT', 'PREREQ_CHECK_TYPE_MISSING_STARTER_CODE',
+  'SUPPORTED_PYTHON_LIBS',
+  function(
+      CodePrereqCheckResultObjectFactory, PrereqCheckFailureObjectFactory,
+      PREREQ_CHECK_TYPE_BAD_IMPORT, PREREQ_CHECK_TYPE_MISSING_STARTER_CODE,
+      SUPPORTED_PYTHON_LIBS) {
+
+    var rightTrim = function(str) {
+      //remove trailing white space at end of string
```
seanlip: Nit: format as
```
// Remove trailing white space at end of string.
```
(start with capital letter, have full stop at end, have space after "//")

client/services/code_evaluators/PythonPrereqCheckService.js
```diff
@@ -0,0 +1,126 @@
...
+      var RIGHT_TRIM_PATTERN = /\s+$/;
+      return str.replace(RIGHT_TRIM_PATTERN, '');
+    };
+
+    var extractTopLevelFunctionLines = function(starterCode) {
+      var starterCodeLines = starterCode.split('\n');
+      var topLevelFunctionLines = [];
+      for (var i = 0; i < starterCodeLines.length; i++) {
+        var line = rightTrim(starterCodeLines[i]);
+        if (line.startsWith("def ")) {
```
seanlip: Nit: use single quotes for consistency with rest of codebase

client/services/code_evaluators/PythonPrereqCheckService.js
```diff
@@ -0,0 +1,126 @@
...
+          return SUPPORTED_PYTHON_LIBS.indexOf(importedLibrary) === -1;
+        }
+      );
+      return unsupportedImports;
+    };
+
+    return {
+      // Returns a promise.
+      checkCode: function(starterCode, code) {
+        // check that starter code is present
```
seanlip: Same comments as above for the styling of all these comments (start with capital letter, end with period).

client/services/code_evaluators/PythonPrereqCheckService.js
a-zakem: Fixed.

client/services/code_evaluators/PythonPrereqCheckService.js
a-zakem: Fixed.

client/services/code_evaluators/PythonPrereqCheckService.js
a-zakem: Fixed.

client/domain/PreRequisiteEvalResultObjectFactory.js
seanlip: Hm, I think we can do that in a separate pass if needed -- I've added it to my radar. Thanks for noting this!

seanlip:
@eyurko -- can you PTAL and approve if you have no further comments? Then I'll handle the merge.

Thanks!

client/domain/PreRequisiteEvalResultObjectFactory.js
seanlip: Hi @a-zakem,

I just realized something -- the code that's used for PrerequisiteEvalResult is not actually the same as the code in the CodeEvalResult, right? If I'm not mistaken, the first one is the raw code and the second is the code after a bunch of extra testing stuff is added to it...

a-zakem:
Hi Sean,

I think that depends on when the CodeEvalResult object is created.  If the
CodeEvalResult object is created as a result of a syntax error during the
the initial run of the raw code, then the CodeEvalResult object will
contain the raw code.  If the student submission passes the syntax error
check and proceeds to pre-processing, then the CodeEvalResult object will
contain the pre-processed code including the testing code.

In the PreRequisites branch as merged, PreRequisiteEvalResultObjectFactory
was renamed CodePrereqCheckResultObjectFactory, and the
CodePrereqCheckResult does not store the submitted code (raw or
otherwise).  At present, I believe the stored code in the CodeEvalResult is
only used in the FeedbackGeneratorService's _getBuggyOutputTestFeedback
function, to check that the student has submitted different code before
incrementing the hintIndex for the same error.

client/domain/PreRequisiteEvalResultObjectFactory.js
a-zakem: Hi Sean,

I think that depends on when the CodeEvalResult object is created.  If the
CodeEvalResult object is created as a result of a syntax error during the
the initial run of the raw code, then the CodeEvalResult object will
contain the raw code.  If the student submission passes the syntax error
check and proceeds to pre-processing, then the CodeEvalResult object will
contain the pre-processed code including the testing code.

In the PreRequisites branch as merged, PreRequisiteEvalResultObjectFactory
was renamed CodePrereqCheckResultObjectFactory, and the
CodePrereqCheckResult does not store the submitted code (raw or
otherwise).  At present, I believe the stored code in the CodeEvalResult is
only used in the FeedbackGeneratorService's _getBuggyOutputTestFeedback
function, to check that the student has submitted different code before
incrementing the hintIndex for the same error.

seanlip: Ah ok -- I checked SolutionHandlerService, and the codeEvalResult is created after running the preprocessed code, so it would be different from the code stored in the prereq check result object (if we had indeed stored code in it). Since we didn't, though, that's a moot point, and I think what is currently in master looks good. We can always add the raw code later to the prereq result if we need it.

Given this, I don't think we need to pull the code out of CodeEvalResult, either now or in the future. Would you agree?

a-zakem: For my two cents, I think it is fine to leave the code in CodeEvalResult at present, as there isn't a need to access the student code that caused a PrereqCheckFailure at some later point in time.  If there is a feature in the future that could use that information, the submitted code could be pulled out of CodeEvalResult and added to Snapshot instead.

seanlip: OK, sounds good -- though the code in PrereqCheckFailure would still be different from the code in CodeEvalResult, so we might want to still keep the code with the results of the run.

In any case, though, this depends on what we do in the future, so let's punt the decision. Thanks for your thoughts!

PR #238, opened 2017-04-20
Add ability to change question sets.

eyurko:
Wasn't sure if this was something we wanted, but it's something we _could_ have, now.

And it's got a switch statement! Success all around.


eyurko:
Incorporated into #250.

PR #239, opened 2017-04-20
Dark Theme Fixes: Fix #232, Temp Fix #234

### LINKED ISSUE #232: "Saving code now..." in "Dark" theme is in black font
Expected Behavior: "Saving code now..." in light colored font, so that it is readable in dark background.

Observed Behavior: "Saving code now..." in black font.

Steps to Reproduce: Switch to "Dark" theme setting and wait 20-30 seconds for the "Saving code now..." message to appear next to right of the "Reset Code" button.

### LINKED ISSUE #234: Dark background theme does not cover whole scrollable area
Expected behavior: When making window small and scrolling left or down, background color in "Dark" them covers entire area, including areas scrolled to.

Observed behavior: Background color in "Dark" theme only covers area size of window. See screenshot below...

dark-bg-not-all-dark

Steps to reproduce: Size window to a small window such that you need to scroll to see the whole app. Notice the areas you scroll to do not have dark background color.

theresa-lee:
'Saving code...' text is now readable in Dark Mode. Temp fix for the background color height issue and also added Dark Night for code blocks in instructions.

### DIFF
--- a/client/components/LearnerViewDirective.js
+++ b/client/components/LearnerViewDirective.js
@@ -117,7 +117,7 @@ tie.directive('learnerView', [function() {
                   <div ng-repeat="previousInstruction in previousInstructions track by $index">
                     <div ng-repeat="instruction in previousInstruction track by $index">
                       <p ng-if="instruction.type == 'text'">{{instruction.content}}</p>
-                      <pre class="tie-question-code" ng-if="instruction.type == 'code'">{{instruction.content}}</pre>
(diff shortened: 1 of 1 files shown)


rabidbit:
Issue #234 still seems to persist for me.

theresa-lee:
Okay, I added some min width as well (horizontally resizing was working with prev fix but this should fix the vertical one, which I didn't see originally).

rabidbit:
Might need to sync with master, as a bug fix was pushed this morning.

PR #243, opened 2017-04-21
Add angular min.js.map file.

seanlip:


rabidbit:
Note that this addresses the map file not found error but other CodeStorageService related error persists for Safari. Will file issue.

PR #247, opened 2017-04-22
Added MINGW_NT-10.0 Compatibility to setup.sh

BenjaminHolland:


seanlip:
Hi @BenjaminHolland! We need a CLA signed before we can accept the PR. Would you mind following the instructions [here](https://github.com/google/tie/pull/247#issuecomment-296342463)?

Also, it looks like the Travis tests are failing...

BenjaminHolland:
Ah. I signed the CLA but didn't comment. 

It's a little weird the travis tests are failing, as they passed on my machine both individually and during the git push. I haven't used a whole lot of CI stuff, but I'll run the tests again locally and see what's up.

BenjaminHolland:
Oh, I know what it is.  Gonna have to find a way to normalize line endings. It might be worth putting in a .gitattributes as well.

BenjaminHolland:
I'm a bit worried about whether or not I did the .gitattributes correctly, since I'm pretty new to cross-platform development. It might be worth getting someone slightly more knowledgeable than I am to look it over before merging.

scripts/setup.sh
```diff
@@ -79,16 +84,27 @@ if [ ! -d "$NODE_DIR" ]; then
     else
       NODE_FILE_NAME=node-v6.9.1-linux-x86
     fi
+  elif [ ${OS}==MINGW64_NT-10.0 ]; then
```
seanlip: Add spaces around == for consistency.

scripts/setup.sh
```diff
@@ -23,7 +23,11 @@ export OS=`uname`
 export MACHINE_TYPE=`uname -m`
 export TOOLS_DIR=./tools
 export NODE_DIR=$TOOLS_DIR/node-6.9.1
-export NPM_CMD=$NODE_DIR/bin/npm
+if [ ${OS}==MINGW_NT-10.0 ]; then
```
seanlip: Will this mess up the PATH on line 33?

BenjaminHolland: Shouldn't, since the PATH on line 33 references $NODE_DIR instead of $NODE_CMD. $NODE_DIR is the same regardless of OS, but $NPM_CMD is different in the windows distro of node, which doesn't like to use /bin directories for some reason.

seanlip: Ah, sorry, I misread. Thanks!

seanlip:
Hi @BenjaminHolland! Just to clarify, btw, I didn't mean replicate the .gitattributes wholesale. (This codebase doesn't have .jscsrc, for example.) The file I pointed to was intended just as an example of a format that seems to have worked fine in the past.

BenjaminHolland:
OK. Pulled on both linux and windows and got successful tests. Everything should be good here now. 
Re the .gitattributes
I got a bit frustrated with git being a bit fiddly, so I just copied in the entire known good file as a starting point. It's not perfect, but It works, and I'm refining it now. Hopefully after that I can start working on something a bit more interesting.

BenjaminHolland:
Woot. Now people can pull and run stuff using Git Bash. 
Proper cross compatibility will take some time, Ideally, the tests should be configured to run differently on different platforms, but since I know even less about the test platforms than I do about git and line endings, this should do for the moment.

.gitattributes
```diff
@@ -0,0 +1,21 @@
+# Source files
+# ============
+*.css      text eol=lf
+*.html     text eol=lf
+*.md       text eol=lf
+*.map	   text eol=lf
```
seanlip: Slightly OCD nit: please indent text by one space to line up with the rest.

.gitattributes
```diff
@@ -0,0 +1,21 @@
...
+*.js       text eol=lf
+*.json     text eol=lf
+*.yml      text eol=lf
```
seanlip: This codebase doesn't have .yml, I think...

.gitattributes
```diff
@@ -0,0 +1,21 @@
...
+
+# Misc files
+# ==========
+.gitattributes	text eol=lf
+.gitignore   	text eol=lf
+AUTHORS      	text eol=lf
+CHANGELOG    	text eol=lf
+CONTRIBUTORS 	text eol=lf
+LICENSE      	text eol=lf
+Vagrantfile  	text eol=lf
```
seanlip: This codebase doesn't have Vagrantfile either.

scripts/setup.sh
seanlip: Re-flagging this comment.

seanlip:
Thanks @BenjaminHolland! Looks great :)

PR #250, opened 2017-04-25
Improve questions, add question sets, and add ability to switch between question sets.

eyurko:
I think I got most of the things I've noticed.

For a variety of trying-not-to-shoot-myself-in-the-foot-long-term reasons, this also includes the "Add ability to change question sets" PR (hate them merge conflicts), so I'm closing #238.


eyurko:
Would like CSS / front-end approval from @theresa-lee, as well.

client/components/LearnerViewDirective.js
```diff
@@ -80,6 +80,12 @@ tie.directive('learnerView', [function() {
                   <select class="tie-lang-select-menu" name="lang-select-menu">
                     <option value="Python" selected>Python</option>
                   </select>
+                  <select class="tie-question-set-select" name="question-set-select"
+                          ng-change="changeQuestionSet()" ng-model="currentQuestionSetId"
+                          ng-options="i.questionSetId as i.questionSetId for i in questionSetIds">
+                    <option style="display: none" value="">Question Set</option>
```
seanlip: Why do you need this and the next inner option?

client/components/LearnerViewDirective.js
```diff
@@ -450,6 +462,12 @@ tie.directive('learnerView', [function() {
         // TODO(sll): Generalize this to dynamically select a question set
         // based on user input.
         var questionSetId = 'strings';
+        var questionSetIds = [
+          {questionSetId: 'strings'},
+          {questionSetId: 'all'},
+          {questionSetId: 'other'}
```
seanlip: Perhaps "other" before "all"? "all" seems like it's just for testing.

client/components/LearnerViewDirective.js
```diff
@@ -568,6 +586,33 @@ tie.directive('learnerView', [function() {
           }
         };
 
+        $scope.changeQuestionSet = function() {
```
seanlip: I suggest actually passing in the questionSetId as an argument.

client/components/LearnerViewDirective.js
```diff
@@ -568,6 +586,33 @@ tie.directive('learnerView', [function() {
...
+          questionSetId = '';
+          switch ($scope.currentQuestionSetId) {
```
seanlip: Cleaner to have, somewhere:

    var ALLOWED_QUESTION_SET_IDS = [...];

and then do

    if (ALLOWED_QUESTION_SET_IDS.indexOf(questionSetId) === -1) {
      return;
    }

    ...

client/components/LearnerViewDirective.js
```diff
@@ -568,6 +586,33 @@ tie.directive('learnerView', [function() {
...
+          case 'all':
+            questionSetId = $scope.currentQuestionSetId;
+            break;
+          default:
+            return;
+          }
+          QuestionDataService.initCurrentQuestionSet(questionSetId);
+          questionSet = QuestionDataService.getCurrentQuestionSet(
+            questionSetId);
+          $scope.currentQuestionIndex = 0;
```
seanlip: Don't copy/paste. Extract all this setup stuff into another method (initQuestionSet, perhaps).

client/components/LearnerViewDirective.js
eyurko: I think that one's the title, but I don't think I need the next one.

eyurko: Yeah -- that's done for consistency with the "Theme" dropdown, below. It also has the <option></option>, would you like me to remove that one as well?

client/components/LearnerViewDirective.js
eyurko: This is also consistent with the implementation of the themes -- should we change the way we do theming, as well?

client/components/LearnerViewDirective.js
seanlip: Yeah, let's remove stuff that's not needed. Thanks!

client/components/LearnerViewDirective.js
seanlip: Yes, that would be better, I think, if that's OK!

client/components/LearnerViewDirective.js
eyurko: done

client/components/LearnerViewDirective.js
eyurko: done, though I will miss the switch statements.

client/components/LearnerViewDirective.js
eyurko: sure, sounds good.

client/components/LearnerViewDirective.js
eyurko: aight

client/components/LearnerViewDirective.js
eyurko: aight

client/components/LearnerViewDirective.js
```diff
@@ -557,17 +562,42 @@ tie.directive('learnerView', [function() {
           $scope.$apply();
         };
 
-        $scope.changeTheme = function() {
-          if ($scope.theme === 'Dark') {
+        $scope.changeTheme = function(theme) {
```
seanlip: Suggest newTheme, for parity with newQuestionSetId.

client/components/LearnerViewDirective.js
```diff
@@ -446,10 +456,17 @@ tie.directive('learnerView', [function() {
           CodeStorageService, SECONDS_TO_MILLISECONDS, DEFAULT_AUTOSAVE_SECONDS,
           DISPLAY_AUTOSAVE_TEXT_SECONDS) {
         var DURATION_MSEC_WAIT_FOR_SCROLL = 20;
+        var ALLOWED_QUESTION_SET_IDS = ['strings', 'other', 'all'];
         var language = LANGUAGE_PYTHON;
         // TODO(sll): Generalize this to dynamically select a question set
         // based on user input.
         var questionSetId = 'strings';
+        var questionSetIds = [
```
seanlip: Not sure you need the "var" variable -- you can just define $scope.questionSetIds directly.

That said, probably better not to hardcode two things with the same data. Can $scope.questionSetIds be derived from ALLOWED_QUESTION_SET_IDS somehow?

client/components/LearnerViewDirective.js
eyurko: done

client/components/LearnerViewDirective.js
eyurko: done

seanlip:
LGTM, though I think you still want @theresa-lee's approval. Thanks @eyurko!

Issue #186, opened 2017-04-08
Give users positive reinforcement about anticipating edge cases and feedback on failing tests


apicardgoog:
When a user's code passes a broad category of tests (e.g. short inputs) we would like to give positive reinforcement for their work.
Also, if tests that used to pass stop working, we would like to remind them that their code needs to be fixed.

We would like to do this by adding tags to the test inputs that group them into relevant categories (e.g. "the general case", "short inputs", "large inputs", "sorted inputs", etc.). 

Then, all inputs would need to be evaluated with each run in order to determine the tags that are passing and the tags that are still failing.

Design doc is here:
https://goto.google.com/tie-positive

Suggested implementation:
1. Maintain a list of tags whose tests are all passing (passing_tags_list starts off empty) and a list of failing tests that are displayed in the output (displayed_failing_tests_list starts off empty)
2. Run all the tests compiling a list of all passing tests and a separate list of all failing tests.
3. If all the tests with a particular tag are passing and that tag is not in passing_tags_list, add "☑ Handles <name of tag>" to the output and add the tag to the list.
4. If a test fails that is associated with a tag in passing_tags_list, change the check mark to a red X for the entry that refers to the tag name. Add the test to displayed_failing_tests_list and display a message of the form "☒ Fails on 'abc'".
5. If a test in  displayed_failing_tests_list goes back to passing, update the X to a check mark. Evaluate whether the check mark next to corresponding tag message should become a checkmark too (depends on whether other tests with that tag are still failing)
6. Optional, if no formerly passing test has started failing, prompt the user about the next failing test.

seanlip:
Fixed in #251. Thanks @shaman-rajan!

PR #251, opened 2017-04-25
Solving #186 Positive reinforcement

shaman-rajan:
I'm still working on editing the data schema to all questions, but created the PR to take comments in the meantime. 2 autosave karma tests fail, and I'm working on fixing that too.


shaman-rajan:
All tests pass now and all questions have tags included. @eyurko PTAL.

assets/tests/TaskSchemaValidationService.js
```diff
@@ -157,6 +157,12 @@ tie.factory('TaskSchemaValidationService', [
           });
         });
       },
+      verifyCorrectnessTestsHaveTags: function(task) {
+        var correctnessTests = task.getCorrectnessTests();
+        return correctnessTests.every(function(test) {
+          return test.getTag() !== undefined && test.getTag().length > 0;
```
seanlip: angular.isString(test.getTag()) is probably a stronger check.

client/components/LearnerViewDirective.js
```diff
@@ -481,6 +505,19 @@ tie.directive('learnerView', [function() {
         var autosaveCancelPromise;
         var cachedCode;
         var congratulatoryFeedback = FeedbackObjectFactory.create();
+        var reinforcement = ReinforcementObjectFactory.create();
+        QuestionDataService.initCurrentQuestionSet(questionSetId);
+        var questionSet = QuestionDataService.getCurrentQuestionSet(
+          questionSetId);
+        $scope.currentQuestionIndex = 0;
```
seanlip: Don't copy/paste please -- refactor common stuff into a function instead. That said, I just made a similar comment on #250 which has just been merged, so merging from master into this branch and resolving the conflicts should address this.

client/domain/ReinforcementBulletObjectFactory.js
```diff
@@ -0,0 +1,57 @@
...
+ * reinforcementBullet domain objects.
+ */
+
+tie.factory('ReinforcementBulletObjectFactory', [
+  function() {
+
+    var ReinforcementBullet = function(passed, content) {
+      this._passed = passed;
+      this._content = content;
+      if (passed) {
```
seanlip: this._imgName = (passed ? 'pass.png' : 'fail.png');

but then, actually, you don't even need this._imgName at all. Just add a method ReinforcementBullet.prototype.getImageFilename() and do the computation there.

client/services/ReinforcementGeneratorService.js
```diff
@@ -0,0 +1,98 @@
...
+/**
+ * @fileoverview Service for generating positive reinforcement when a user
+ * submits code based on the tasks that the code passes/fails on.
+ */
+
+tie.factory('ReinforcementGeneratorService', [
+  function() {
+
+    return {
+      getReinforcement: function(task, codeEvalResult, runtimeFeedback) {
```
seanlip: It's an antipattern to add new keys to an object mid-stream -- let's avoid that. It means that the same-looking object has a different definition depending on which file in the code you're looking at.

Instead, construct the Reinforcement object here, and return just that.

client/services/SolutionHandlerService.js
```diff
@@ -60,7 +60,9 @@ tie.factory('SolutionHandlerService', [
               task, codeEvalResult, codeSubmission.getRawCodeLineIndexes());
             TranscriptService.recordSnapshot(
               SnapshotObjectFactory.create(codeEvalResult, runtimeFeedback));
-            return runtimeFeedback;
+            var finalFeedback = ReinforcementGeneratorService.getReinforcement(
```
seanlip: Don't modify the feedback object. Instead, when passing stuff to the learner view directive, pass the reinforcement object should be passed as a separate object from the usual feedback. Do all the computations in services so that there is no logic in the directive controller.

shaman-rajan:
@seanlip PTAL. Comments addressed except the one we can fix during merge.

client/components/LearnerViewDirective.js
shaman-rajan: This isn't needed, will remove it.

shaman-rajan:
Looks like there are still issues here. When submitting with a syntax error, the app hangs. Will update once it's fixed.

seanlip:
OK, thanks for the update @shaman-rajan! I'll take another look when it's ready.

shaman-rajan:
@seanlip PTAL. Fixed the syntax error issue, and addressed all comments.

client/components/LearnerViewDirective.js
```diff
@@ -318,6 +324,21 @@ tie.directive('learnerView', [function() {
         .tie-previous-instructions {
           opacity: 0.5;
         }
+        .tie-reinforcement li {
+          list-style: none;
+          margin: 0;
+          margin-top: 1px;
+          position: relative;
+        }
+        .bullet-img {
```
seanlip: Could you use `.tie-bullet-img` or `.tie-reinforcement .bullet-img` instead? We prefix all our CSS classes with `.tie-` in order to distinguish them from third-party libs.

client/components/LearnerViewDirective.js
```diff
@@ -318,6 +324,21 @@ tie.directive('learnerView', [function() {
...
+          width: 15px;
```
seanlip: Nit: recommend alphabetizing these attrs (i.e. bottom, then height, then position, then width). As the number of attributes increases, this makes it easier to locate a particular one.

client/components/LearnerViewDirective.js
```diff
@@ -553,9 +588,28 @@ tie.directive('learnerView', [function() {
             } else if (syntaxErrorIndex === null) {
               $scope.syntaxErrorString = '';
               $scope.syntaxErrorFound = false;
+
+              // Updating reinforcement bullets only if no syntax errors
+              reinforcement.clear();
+              for (var caseList in reinforcementObject.passedList) {
+                if (reinforcementObject.passedList[caseList]) {
```
seanlip: The reinforcement object should be fully prepared in a service and just passed to the controller -- there should be none of this logic in the directive's controller.

If necessary, you can handle the "don't show if there are syntax errors" bit with an if statement -- but that's about as much logic as should exist at this level of the code.

client/domain/ReinforcementBulletObjectFactory.js
```diff
@@ -0,0 +1,52 @@
...
+    ReinforcementBullet.prototype.isPassedBullet = function() {
+      return this._passed;
+    };
+
+    ReinforcementBullet.prototype.getContent = function() {
+      return this._content;
+    };
+
+    ReinforcementBullet.prototype.getImgName = function() {
+      return this._passed ? "pass.png" : "fail.png";
```
seanlip: Use single quotes here for consistency with the other files.

client/services/ReinforcementGeneratorService.js
```diff
@@ -0,0 +1,99 @@
...
+        }
+        if (!task.pastFailsList) {
+          task.pastFailsList = {};
+        }
+
+        var splitTestsByTag = function(tests) {
+          var splitTests = {};
+          for (var i = 0; i < tests.length; ++i) {
+            var test = tests[i];
+            test.index = i;
```
seanlip: Please avoid modifying dicts like this. If you change the keys midstream then it's hard to tell what a dict's structure is at any given time.

client/services/ReinforcementGeneratorService.js
```diff
@@ -0,0 +1,99 @@
...
+
+tie.factory('ReinforcementGeneratorService', [
+  function() {
+
+    return {
+      getReinforcement: function(task, codeEvalResult) {
+
+        // Initializing question reinforcement data if not done already
+        if (!task.passedList) {
+          task.passedList = {};
```
seanlip: Please avoid modifying dicts like this. If you change the keys midstream then it's hard to tell what a dict's structure is at any given time.

Also, why is this defined on the task? I thought it would be part of the reinforcement object, conceptually?

client/services/ReinforcementGeneratorService.js
```diff
@@ -0,0 +1,99 @@
...
+          task.pastFailsList = {};
+        }
+
+        var splitTestsByTag = function(tests) {
+          var splitTests = {};
+          for (var i = 0; i < tests.length; ++i) {
+            var test = tests[i];
+            test.index = i;
+            var testTag = test.getTag();
+            if (testTag in splitTests) {
```
seanlip: Use splitTests.hasOwnProperty(testTag) for checking object membership.

client/services/ReinforcementGeneratorService.js
```diff
@@ -0,0 +1,99 @@
...
+        // Initializing question reinforcement data if not done already
+        if (!task.passedList) {
+          task.passedList = {};
+        }
+        if (!task.pastFailsList) {
+          task.pastFailsList = {};
+        }
+
+        var splitTestsByTag = function(tests) {
+          var splitTests = {};
```
seanlip: testsByTag or testsKeyedByTag might be a better name.

client/services/ReinforcementGeneratorService.js
```diff
@@ -0,0 +1,99 @@
...
+            if (testTag in splitTests) {
+              splitTests[testTag].push(test);
+            } else {
+              splitTests[testTag] = [test];
+            }
+          }
+          return splitTests;
+        };
+
+        // Go through correctness tests to update reinforcement data
```
seanlip: Add a period at the end of the comment.

client/services/SolutionHandlerService.js
```diff
@@ -39,7 +39,10 @@ tie.factory('SolutionHandlerService', [
               potentialSyntaxErrorString);
             TranscriptService.recordSnapshot(
               SnapshotObjectFactory.create(rawCodeEvalResult, feedback));
-            return $q.resolve(feedback);
+            return {
+              feedbackObject: feedback,
+              reinforcementObject: {}
```
seanlip: This should be a blank Reinforcement object. Avoid passing dicts around.

client/services/ReinforcementGeneratorService.js
```diff
@@ -0,0 +1,99 @@
...
+  function() {
+
+    return {
+      getReinforcement: function(task, codeEvalResult) {
+
+        // Initializing question reinforcement data if not done already
+        if (!task.passedList) {
+          task.passedList = {};
+        }
+        if (!task.pastFailsList) {
```
seanlip: I think I understand why you're doing it like this -- you want to maintain history. However, defining stuff on the task dict midstream is not the right way to do this, since the task is supposed to be independent of the individual student's playthrough.

So, there are two ways we can go here:
(a) The "proper" way to do this is probably to preserve info about "past fails" somewhere. Perhaps you can store these in the transcript, and extract them from there.
(b) If you want to get this PR in quicker, though, that may be too much to do. In that case, let's drop the "past fails" requirement for this PR, and handle things in a stateless manner.

I'm happy either way. Let me know if you want to discuss any of this.

BenjaminHolland: I kind of like the idea of having a more robust transcript system, especially for repeat playthroughs. Eventually we might be able to turn that into a progress or improvement system of sorts.

seanlip:
Also, FYI, there are merge conflicts.

client/components/LearnerViewDirective.js
shaman-rajan: I was using a style similar to congratulatoryFeedback defined on line 515. The reinforcement object we get from the service is always a new one, from which entries are taken and the main stored object is updated.

seanlip: Ah, I see. Let's not do that -- congratulatoryFeedback is a special case, and is just meant to show a message at the end of the task (which is just a display thing, and no logic is required). However, for something complex like reinforcement data, it's better to just pass the object wholesale from the service and do the logic within the service.

client/services/ReinforcementGeneratorService.js
shaman-rajan: I like the idea of adding it to the transcript. Will work on doing that.

shaman-rajan:
@seanlip PTAL

client/domain/ReinforcementObjectFactory.js
```diff
@@ -0,0 +1,85 @@
...
+      }
+
+      return bullets;
+    };
+
+    Reinforcement.prototype.getTask = function() {
+      return this._task;
+    };
+
+    Reinforcement.prototype.addToPassedList = function(key, value) {
```
seanlip: What exactly do `key` and `value` represent here? (E.g. is the "key" a "tag"?)

Please add comments to explain, or use clearer variable names. Ditto below.

client/domain/ReinforcementObjectFactory.js
```diff
@@ -0,0 +1,85 @@
...
+              "Handles " + caseList));
+        } else {
+          bullets.push(
+            ReinforcementBulletObjectFactory.createFailedBullet(
+              "Fails " + caseList));
+        }
+      }
+
+      for (var testCase in this._pastFailsList) {
+        if (this._pastFailsList[testCase]) {
```
seanlip: In principle, is it possible for testCase here to overlap a caseList in this._passedList? If so, how do you guard against that (within this object)?

client/services/ReinforcementGeneratorService.js
```diff
@@ -0,0 +1,101 @@
...
+ * submits code based on the tasks that the code passes/fails on.
+ */
+
+tie.factory('ReinforcementGeneratorService', [
+  'ReinforcementObjectFactory', 'TranscriptService',
+  function(ReinforcementObjectFactory, TranscriptService) {
+    return {
+      getReinforcement: function(task, codeEvalResult) {
+
+        var testsKeyedByTag = function(tests) {
```
seanlip: Move this between lines 22 and 23 -- no need for inner functions.

Also: getTestsKeyedByTag. Names of functions should start with verbs.

client/services/ReinforcementGeneratorService.js
```diff
@@ -0,0 +1,101 @@
...
+          var splitTests = {};
+          for (var i = 0; i < tests.length; ++i) {
+            var test = tests[i];
+            test.originalIndex = i;
```
seanlip: Please do not modify objects midstream -- we talked about this before.

client/services/ReinforcementGeneratorService.js
```diff
@@ -0,0 +1,101 @@
...
+        if (previousSnapshot !== null) {
+          previousReinforcement = previousSnapshot.getReinforcement();
+        }
+
+        var reinforcement = ReinforcementObjectFactory.create(task);
+
+        // Copy the previous reinforcement object if we are on the same task.
+        if (previousReinforcement !== null &&
+            task === previousReinforcement.getTask()) {
+          for (var tag in previousReinforcement.getPassedList()) {
```
seanlip: Define a local variable for previousReinforcement.getPassedList() and use that instead; don't keep calling the method repeatedly (especially in a for loop). Ditto for the past fails list.

It may make sense to add a reinforcement.clonePassedList() method that uses angular.copy() to bulk-populate the list; if you go that route, make sure to document the expected inputs.

client/services/ReinforcementGeneratorService.js
```diff
@@ -0,0 +1,101 @@
...
+              previousReinforcement.getPassedList()[tag]);
+          }
+          for (var pastCase in previousReinforcement.getPastFailsList()) {
+            reinforcement.addToPastFailsList(pastCase,
+              previousReinforcement.getPastFailsList()[pastCase]);
+          }
+        }
+
+        // Go through correctness tests to update reinforcement data.
+        var correctnessTests = testsKeyedByTag(task.getCorrectnessTests());
```
seanlip: correctnessTestsByTag

client/services/ReinforcementGeneratorService.js
```diff
@@ -0,0 +1,101 @@
...
+        observedOutputs = observedOutputs[observedOutputs.length - 1];
+        var failedCaseSeenOverall = false;
+        for (var testTag in correctnessTests) {
+          var failedCaseSeenInTag = false;
+          for (var testIdx = 0; testIdx < correctnessTests[testTag].length;
+              ++testIdx) {
+            var test = correctnessTests[testTag][testIdx];
+            var observedOutput = observedOutputs[test.originalIndex];
+            if (test.matchesOutput(observedOutput)) {
+              if (test.getInput() in reinforcement.getPastFailsList()) {
```
seanlip: reinforcement.hasPastFailure(test.getInput())

client/services/ReinforcementGeneratorService.js
```diff
@@ -0,0 +1,101 @@
...
+                reinforcement.addToPastFailsList(test.getInput(), true);
```
seanlip: But it's already in the list. Should this be "modify" instead?

client/services/ReinforcementGeneratorService.js
```diff
@@ -0,0 +1,101 @@
...
+              }
+            } else if (!failedCaseSeenInTag) {
+              // We want to display only 1 new failed case among all tasks.
+              if (!failedCaseSeenOverall) {
+                reinforcement.addToPastFailsList(test.getInput(), false);
+                failedCaseSeenOverall = true;
+              }
+              failedCaseSeenInTag = true;
```
seanlip: I'm confused. If we change failedCaseSeenInTag to true when we see it is false, then what's the point at all of keeping track of failedCaseSeenInTag?

client/services/ReinforcementGeneratorService.js
```diff
@@ -0,0 +1,101 @@
...
+            }
+          }
+          if (failedCaseSeenInTag) {
+            if (testTag in reinforcement.getPassedList()) {
+              reinforcement.addToPassedList(testTag, false);
```
seanlip: I'm starting to think that "passed list" is a bit of a misnomer here. What are your two lists tracking, exactly? Can the names better reflect their contents?

client/services/TranscriptService.js
```diff
@@ -27,6 +27,9 @@ tie.factory('TranscriptService', [
       },
       recordSnapshot: function(snapshot) {
         transcript.recordSnapshot(snapshot);
+      },
+      getPreviousSnapshot: function() {
```
seanlip: getMostRecentSnapshot() perhaps? "Previous" implies any previous one.

client/services/ReinforcementGeneratorService.js
shaman-rajan: This is inside an "else if" clause, so is entered only if the first if condition evaluates to false.

seanlip: Ah. Perhaps, then:

    } else {
      failedCaseSeenInTag = true;
      if (!failedCaseSeenOverall) {
        ...
      }
    }

might be slightly less efficient, but clearer?

client/domain/ReinforcementObjectFactory.js
```diff
@@ -21,15 +21,15 @@ tie.factory('ReinforcementObjectFactory', [
   'ReinforcementBulletObjectFactory', function(ReinforcementBulletObjectFactory) {
     var Reinforcement = function(task) {
       this._task = task;
-      this._passedList = {};
-      this._pastFailsList = {};
+      this._passedTagsList = {};
```
seanlip: One other thing I forgot to mention earlier (sorry) -- can we call these this._passedTags and this._pastFailedCases? They're not lists; lists are ordered.

client/domain/ReinforcementObjectFactory.js
```diff
@@ -59,20 +59,24 @@ tie.factory('ReinforcementObjectFactory', [
...
     };
 
-    Reinforcement.prototype.getPastFailsList = function() {
-      return this._pastFailsList;
+    Reinforcement.prototype.getPastFailedCasesList = function() {
+      return this._pastFailedCasesList;
+    };
+
+    Reinforcement.prototype.hasPastFailedCase = function(caseInput) {
+       return (caseInput in this._pastFailedCasesList);
```
seanlip: return this._pastFailedCasesList.hasOwnProperty(caseInput)

Btw, are the caseInput's guaranteed to be strings / hashable?

client/domain/ReinforcementObjectFactory.js
```diff
@@ -59,20 +59,24 @@ tie.factory('ReinforcementObjectFactory', [
       return this._task;
     };
 
-    Reinforcement.prototype.addToPassedList = function(key, value) {
-      this._passedList[key] = value;
+    Reinforcement.prototype.addToPassedTagsList = function(key, value) {
```
seanlip: Still need docstrings.

shaman-rajan:
@seanlip Addressed comments. 1 thing left to consider is whether all inputs are hashable. If we are using templates to enter test cases, they will initially be strings that may later be parsed to other data structures. I will investigate on whether the objects I am using to hash are strings taken from the template or parsed objects and update.

shaman-rajan:
@seanlip PTAL. Addressed all comments including the hashability of inputs.

client/domain/CorrectnessTestObjectFactory.js
```diff
@@ -24,13 +24,21 @@ tie.factory('CorrectnessTestObjectFactory', [
...
     };
 
     // Instance methods.
     CorrectnessTest.prototype.getInput = function() {
       return this._input;
     };
 
+    CorrectnessTest.prototype.getStringifiedInput = function() {
+      if (!this._stringifiedInput) {
+        this._stringifiedInput = JSON.stringify(this._input);
```
seanlip: Again, do not modify dicts mid-stream. Either always compute on the fly, or store at the outset and do a simple lookup.

client/domain/ReinforcementObjectFactory.js
```diff
@@ -0,0 +1,104 @@
...
+ * @fileoverview Factory for creating new frontend instances of Reinforcement
+ * domain objects.
+ */
+
+tie.factory('ReinforcementObjectFactory', [
+  'ReinforcementBulletObjectFactory', function(
+      ReinforcementBulletObjectFactory) {
+    var Reinforcement = function(task) {
+      this._task = task;
+      this._passedTags = {};
```
seanlip: Can you document the keys/values (and their types) for these two objects?

client/domain/ReinforcementObjectFactory.js
```diff
@@ -0,0 +1,104 @@
...
+
+      return bullets;
+    };
+
+    Reinforcement.prototype.getTask = function() {
+      return this._task;
+    };
+
+    // Add a tag to the list of tags the code passes/fails
+    Reinforcement.prototype.addToPassedTags = function(tag, passedBool) {
```
seanlip: addPassedTag

client/domain/ReinforcementObjectFactory.js
```diff
@@ -0,0 +1,104 @@
...
+      this._passedTags[tag] = passedBool;
+    };
+
+    // Update the pass/fail status of a tag
+    Reinforcement.prototype.updatePassedTags = function(tag, passedBool) {
```
seanlip: updatePassedTag

client/domain/ReinforcementObjectFactory.js
seanlip: Also, here and below, would testPassed be a better name for the boolean? In general you'll notice that the variable/arg names don't include type information. It seems like a good idea to point out that this is a boolean but I would recommend doing it in the comment/docstring.

client/domain/ReinforcementObjectFactory.js
```diff
@@ -0,0 +1,104 @@
...
+      this._passedTags[tag] = passedBool;
+    };
+
+    // Add a test case to the list of failed test cases
+    Reinforcement.prototype.addToPastFailedCases = function(
+        caseInput, passedBool) {
```
seanlip: The comments here and below should indicate that caseInput is a string, or perhaps you can rename to stringifiedInput.

client/domain/ReinforcementObjectFactory.js
```diff
@@ -0,0 +1,104 @@
...
+    Reinforcement.prototype.getPassedTags = function() {
+      return this._passedTags;
+    };
+
+    Reinforcement.prototype.getPastFailedCases = function() {
+      return this._pastFailedCases;
+    };
+
+    Reinforcement.prototype.hasPastFailedCase = function(caseInput) {
+      return (caseInput in this._pastFailedCases);
```
seanlip: return this._pastFailedCases.hasOwnProperty(stringifiedInput);

client/services/ReinforcementGeneratorService.js
```diff
@@ -0,0 +1,120 @@
...
+ * submits code based on the tasks that the code passes/fails on.
+ */
+
+tie.factory('ReinforcementGeneratorService', [
+  'ReinforcementObjectFactory', 'TranscriptService',
+  function(ReinforcementObjectFactory, TranscriptService) {
+
+    var getTestsKeyedByTagAndIndexes = function(tests) {
+      var splitTests = {};
+      var inputToOriginalIndex = {};
```
seanlip: I think it's worth adding comments to explain what the keys/values of these two objects are, and why they're needed. The names are reasonable but a bit hard to parse for someone who doesn't have context.

client/services/ReinforcementGeneratorService.js
```diff
@@ -0,0 +1,120 @@
...
+
+        var reinforcement = ReinforcementObjectFactory.create(task);
+
+        // Copy the previous reinforcement object if we are on the same task.
+        if (previousReinforcement !== null &&
+            task === previousReinforcement.getTask()) {
+          var previousReinforcementPassedTags =
+            previousReinforcement.getPassedTags();
+          for (var tag in previousReinforcementPassedTags) {
+            reinforcement.addToPassedTags(tag,
```
seanlip: Here and below: if a line of code spans multiple lines, break after "(".

client/services/ReinforcementGeneratorService.js
```diff
@@ -0,0 +1,120 @@
...
+        inputToOriginalIndex[test.getStringifiedInput()] = i;
+        var testTag = test.getTag();
+        if (splitTests.hasOwnProperty(testTag)) {
+          splitTests[testTag].push(test);
+        } else {
+          splitTests[testTag] = [test];
+        }
+      }
+      return {
+        splitTests: splitTests,
```
seanlip: Avoid returning composite blobs like this where possible. In this case, have two separate methods instead, each of which only returns one thing.

client/services/ReinforcementGeneratorService.js
```diff
@@ -0,0 +1,120 @@
...
+        }
+
+        // Go through correctness tests to update reinforcement data.
+        var correctnessTestsByTagAndIndexes = getTestsKeyedByTagAndIndexes(
+          task.getCorrectnessTests());
+        var correctnessTestsByTag = correctnessTestsByTagAndIndexes.splitTests;
+        var testIndexes = correctnessTestsByTagAndIndexes.testIndexes;
+        var observedOutputs = codeEvalResult.getCorrectnessTestResults();
+        // Get results only for the last task.
+        observedOutputs = observedOutputs[observedOutputs.length - 1];
```
seanlip: Don't redefine a variable. Instead, create a new method along the lines of codeEvalResult.getLastTaskResults()

client/domain/CorrectnessTestObjectFactory.js
shaman-rajan: I didn't want to store that unless this function is used. That said, I can do this computation during object creation also.

seanlip: Let's either always do it at object creation time, or always compute it on the fly. I'm fine either way.

shaman-rajan:
@seanlip PTAL

client/domain/CorrectnessTestObjectFactory.js
```diff
@@ -22,15 +22,21 @@ tie.factory('CorrectnessTestObjectFactory', [
...
+      this._tag = correctnessTestDict.tag;
     };
 
     // Instance methods.
     CorrectnessTest.prototype.getInput = function() {
       return this._input;
     };
 
+    CorrectnessTest.prototype.getStringifiedInput = function() {
+      return JSON.stringify(this._input);
```
seanlip: If you are storing the stringified input, and if the input doesn't change, why are you redoing the serialization?

In other words, either do this, or keep line 25, but not both.

client/domain/ReinforcementObjectFactory.js
```diff
@@ -0,0 +1,120 @@
...
+    *  task: The corresponding task for the reinforcement comments.
+    *  passedTags: Dictionary containing tags (test sets) that the code
+    *              passed/failed.
+    *              Key: The tag
+    *              Value: Boolean indicating whether the code passed the tag.
+    *  pastFailedCases: The list of test cases that will be displayed in the
+    *                   reinforcement comments and the corresponding pass/fail
+    *                   status.
+    *                   Key: Stringified input of test case.
+    *                   Value: Boolean indicating where the code passed the
```
seanlip: where --> whether?

client/domain/ReinforcementObjectFactory.js
```diff
@@ -0,0 +1,120 @@
...
+tie.factory('ReinforcementObjectFactory', [
+  'ReinforcementBulletObjectFactory', function(
+      ReinforcementBulletObjectFactory) {
+    /**
+    * The reinforcement object contains the current state of reinforcement
+    * comments about a particular task in a question.
+    * Fields:
+    *  task: The corresponding task for the reinforcement comments.
+    *  passedTags: Dictionary containing tags (test sets) that the code
+    *              passed/failed.
```
seanlip: Add: A tag is considered "failed" if at least one of the correctness tests corresponding to the tag have failed.

client/domain/ReinforcementObjectFactory.js
```diff
@@ -0,0 +1,120 @@
...
+            ReinforcementBulletObjectFactory.createPassedBullet(
+              "Handles " + tag));
+        } else {
+          bullets.push(
+            ReinforcementBulletObjectFactory.createFailedBullet(
+              "Fails " + tag));
+        }
+      }
+
+      for (var testCase in this._pastFailedCases) {
```
seanlip: var stringifiedTestCase

Recommend similar changes elsewhere throughout this file (and possibly others), if necessary, so that it's always clear whether we're dealing with a stringified version or the raw version.

shaman-rajan:
@seanlip PTAL

seanlip:
@shaman-rajan -- this looks great to me, and I'm going to merge it. Thank you very much for all your work on this!

Issue #93, opened 2017-04-04
[Sample question issue] RLE question gives confusing description


jianhaolyu:
In the last phase of this question, the user are required to encode the word in a comprehensible manner, in which "5aaaa" can be decoded properly.
However, the most straight forward way is to encode "5" as "1x5", which conflict with the previous phase's requirement(encoding should be as short as possible).

seanlip:
Thanks, @jianhaolyu!

@eyurko -- any thoughts?

eyurko:
Seems like the messaging should be "provided the encoding is correct / reversible, it should be as short as possible", so that the student prioritizes correctness over brevity.

jianhaolyu:
@eyurko what means "prioritize"? Are students supposed to fulfill both requirements? Since encoding "5" as "1x5" totally violate the brevity requirement.

eyurko:
So, short answer, there's no good way to create a perfect encoding scheme. Part of the last task is understanding that and trying to build something that prioritizes correctness but generally tries to be as short as possible. I think, in the case you've specified, you should not return 1x5, because there's no way to have a decoder misinterpret 5 (or any string of length <= 2). In general, what I'm trying to say is that we should relax the brevity "requirement" from the first segment into more of a suggestion.

If you're shooting for a mix of brevity and correctness, 1x54xa is actually not better (it's longer), but the best version would be either returning 5aaaa or 5a3xa. It seems like it could be worth mentioning "you may find it necessary to relax the brevity requirement in order to accomplish this, but still try to keep the encoded strings as short as possible" as a hint of some kind, but I wouldn't necessarily mention that to the student first if I were asking this as an interview question (and I have, in the past).

seanlip:
Handled in #256.

PR #256, opened 2017-04-26
Made various run-length encoding improvements and added outputFunctionName to the results.

### LINKED ISSUE #86: Last Question (encode and decode)
The problem statement is not so clear. We don't know what to expect as the result of the decoder. The problem said something about one-to-one mapping between the encoder and decoder, 23xa means 2aaa and not 23 of 'a'. However, it is not just clear in the problem statement.

### LINKED ISSUE #95: Confusing result when outputFunctionName != null
Retro steps:
In step3 of Question 4, Run-Length Encoding: 
We write a one-line program to return the input directly, then click 'run'.
screen shot 2017-04-03 at 7 04 53 pm

Output:
The output of our function will be 'bbbbb' when input is '5xb'. 

Expected behavior:
The output of our function will be '5xb' when input is '5xb'.

Possible causes:
'bbbbb' is the result after using the function 'AuxiliaryCode.decodeEncodedString' defined in 'outputFunctionName' attribute in JS script of questions on our output '5xb'. It may confuse some users cause the output in the console is not the actual output of their function. 

Suggestion:
Maybe we can modify the text when outputFunctionName != null. For this question, a possible output could be 'Output after decoding:', 'Expected Output after decoding'. Or we can modify the question: give them the decoding function/strategy to ask them write correct encoding function to make sure the string remain the same after encoding and decoding. Or in the last step we can ask them to implement the decoding function by themselves.

eyurko:
Fixes #86 + #93.

screen shot 2017-04-25 at 5 54 18 pm

Fixes #95.

screen shot 2017-04-25 at 5 54 09 pm

### DIFF
--- a/client/domain/TaskObjectFactory.js
+++ b/client/domain/TaskObjectFactory.js
@@ -19,10 +19,12 @@
 
 tie.factory('TaskObjectFactory', [
   'CorrectnessTestObjectFactory', 'BuggyOutputTestObjectFactory',
-  'PerformanceTestObjectFactory',
+  'PerformanceTestObjectFactory', 'CLASS_NAME_AUXILIARY_CODE',
+  'CLASS_NAME_SYSTEM_CODE',
   function(
       CorrectnessTestObjectFactory, BuggyOutputTestObjectFactory,
-      PerformanceTestObjectFactory) {
+      PerformanceTestObjectFactory, CLASS_NAME_AUXILIARY_CODE,
+      CLASS_NAME_SYSTEM_CODE) {
     var Task = function(taskDict) {
       // A list of strings (where each string corresponds to a paragraph in the
       // UI).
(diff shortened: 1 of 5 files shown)


client/domain/TaskObjectFactory.js
```diff
@@ -67,6 +69,19 @@ tie.factory('TaskObjectFactory', [
       return this._outputFunctionName;
     };
 
+    Task.prototype.getOutputFunctionNameWithoutClass = function() {
+      if (this._outputFunctionName.startsWith(
```
seanlip: This will fail in IE. Use indexOf(...) === 0 instead.

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/startsWith

eyurko: Done.

eyurko:
PTAL

PR #263, opened 2017-04-26
Add BuggyOutputFunction validation test / Fix BuggyOutputFunction error in isPalindrome.

eyurko:
Sorry about the long commit tree -- I'll delete this branch after commit. This fixes #262 and #261.

Issue #261, opened 2017-04-26
Error when running code for isPalindrome question


shaman-rajan:
There is an internal error when running any code for the isPalindrome question.
![screen shot 2017-04-26 at 10 48 51 am](https://cloud.githubusercontent.com/assets/1597618/25448702/301c8c96-2a6e-11e7-9fad-afd6c80f931b.png)

**Repro steps:**
1. Open app.html
2. Navigate to question 3
3. Click run

eyurko:
oh that's fun.

eyurko:
Fixed in #266.

Issue #257, opened 2017-04-26
Wrap student code more carefully.


seanlip:
To reproduce:

1. Open the TIE app and navigate to the RLE problem.
2. Type the following code:
```
def encoded(a):
    return ""

def encode(word):
    return ""
```
3. Run the code. A syntax error will be displayed.

(Note that, if "encoded" in the sample code is replaced by "encoxded", this issue no longer arises.)

The issue occurs because the code preprocessing process replaces the first line of code with `StudentCode(self, ).encoded(a)`, since it sees a function starting with the string "encode" and naively assumes that this needs to be wrapped. Instead, however, it should note that the string doesn't terminate after "encode" -- and, based on that check, it should not do the replacement.

(Thanks @BenjaminHolland for noting this.)

BenjaminHolland:
Specific Location:
/client/services/code_preprocessors/PythonCodePreprocessorService.js::_transformCodeToInstanceMethods() line 71

BenjaminHolland:
Uploading instrumented run to illustrate issue more clearly.
[Issue257 Error Illustration.txt](https://github.com/google/tie/files/957219/Issue257.Error.Illustration.txt)

BenjaminHolland:
Since python treats methods and functions as first-class objects, we should note that the check should be something along the lines of the following:

trim all spaces immediately following the found name.
Check the next character for a valid variable name character. If there is one, do not wrap the function. 
if the next non-whitespace character is ( ) ] or , wrap the function as normal.
if the next non-whitespace character is not in the previous groups, this is a python syntax error
if the statement is an assignment and the next whitespace character is a \n, wrap the function as normal.
otherwise, there's a python syntax error.

PR #264, opened 2017-04-26
Add BuggyOutputFunction validation test / Fix BuggyOutputFunction error in isPalindrome.

### LINKED ISSUE #262: TaskSchemaValidation doesn't properly validate if function names appear in auxiliary code
See issue #261 

buggyFunctionName is AuxiliaryCode.forgetToIgnoreSpace, but forgetToIgnoreSpaces appears in the AuxiliaryCode, so this passes.

eyurko:
This fixes #262 and #261, and of course still contains all the garbage from #263.

Love GitHub, so much.


seanlip:
So, this shouldn't have all the extra stuff. Can you do the following instead?

- do a fresh checkout and pull of the master branch
- create a new branch off of that
- do "git log" to verify that the last commit matches the one [here](https://github.com/google/tie/commits/master)
- make a commit there 
- push that branch

Thanks!

seanlip:
Also, does this address #257?

eyurko:
Nope, but I can fix that.

eyurko:
Actually, it seems like it's a very different problem, so I'd prefer to have a different PR address that rather than conflating it with this one.

seanlip:
Ah, ok with me. Just checking, since it seemed related.

PR #268, opened 2017-04-27
Alternate Layout - Coding window on the right and feedback combined into question.

theresa-lee:


rabidbit:
Thanks! General look and feel comments: (a) make feedback window stand out a bit from the coding window (make feedback background color #fffff7 in light mode and make code background color #333a42 in dark mode), (b) align the question set and dark/light buttons under feedback window but keep the language and reset code buttons aligned under the code window, (c) we need a way to switch between the two alternative UIs (a toggle button or a new client html file, whichever is easiest--don't need to be elaborate since we will likely drop one of the UIs after user testing).

Other UX issues that should be addressed (a) when switching between questions, feedback is lost and reset, (b) when there are multiple syntax errors, only the most recent one is displayed. These are outside the original scope for the alternative UI, but they are much more apparent in this UI. I'll open issues for these two items.

seanlip:
Note: for (c) I recommended to Theresa that we just replace the existing UI in this branch (no toggle button needed) for the sake of technical simplicity. Switching UIs during the user test can be done using git.

theresa-lee:
Okay, I added the colors to make the feedback window stand out and also moved around the buttons (and consolidated the styles as such). Yeah, for (c) switching between the UIs would mean that we would just have this as another directive and a large amount of redundant code between the two.

client/components/LearnerViewDirective.js
```diff
@@ -37,99 +37,103 @@ tie.directive('learnerView', [function() {
...
+                  <div class="tie-feedback-syntax-error">
                   <a href class="tie-feedback-syntax-error-link",
                       ng-click="toggleSyntaxErrorHint()",
                       ng-show="syntaxErrorFound">
                     {{isSyntaxErrorShown ? 'Hide error details' : 'Display error details'}}
                   </a>
                 </div>
                 <br>
-                <span class = "tie-feedback-error-string", ng-show="isSyntaxErrorShown">
+                <span class= "tie-feedback-error-string", ng-show="isSyntaxErrorShown">
```
rabidbit: Space after class=

client/components/LearnerViewDirective.js
```diff
@@ -37,99 +37,103 @@ tie.directive('learnerView', [function() {
...
-              <div class="tie-feedback-window" ng-class="{'night-mode': isInDarkMode}">
-                <div class="tie-feedback">
-                  <p ng-repeat="paragraph in feedbackParagraphs track by $index"
-                      class="tie-feedback-paragraph"
-                      ng-class="{'tie-feedback-paragraph-code': paragraph.isCodeParagraph()}">
-                    <span ng-if="$first">{{feedbackTimestamp}}</span>
-                    <span ng-if="paragraph.isTextParagraph()">
+            <div class="tie-question-ui">
+              <div class="tie-question-window" ng-class="{'night-mode': isInDarkMode}">
+                <div class="tie-greetings">
```
rabidbit: Recommend keeping 'feedback' nomenclature, as the feedback window will more often contain feedback than greetings. Same for other instanced of 'greeting' below.

client/components/LearnerViewDirective.js
theresa-lee: Done.

client/components/LearnerViewDirective.js
```diff
@@ -37,99 +37,103 @@ tie.directive('learnerView', [function() {
...
+                <div class="tie-instructions">
+                  <div ng-repeat="instruction in instructions">
+                    <p ng-if="instruction.type == 'text'">{{instruction.content}}</p>
+                    <pre class="tie-question-code" ng-class="{'night-mode': isInDarkMode}" ng-if="instruction.type == 'code'">{{instruction.content}}</pre>
+                  </div>
+                </div>
+                <div class="tie-feedback">
+                  <div ng-repeat="set in feedbackStorage">
+                    <hr>
+                    <span class="tie-feedback-timestamp">{{set.timestamp}}<span>
```
rabidbit: I'm thinking we no longer need the timestamp for this UI. The timestamp was introduced in the other UI because feedback was overwritten rather than appended; so, if the same error or feedback was given multiple times consecutively, it would not be apparent to the user. Here, since we are always appending feedback, I don't think we need a timestamp anymore.

rabidbit:
As discussed offline, when user gets the answer correct, feedback is cleared, the next step is appended, and a congratulatory note is appended after that. I would expect the existing feedback to remain, append the congratulatory feedback first, then append the next step.

rabidbit:
The Next button overlay is not off center. It was statically set for the other UI; so, it doesn't align properly for this version. See screenshot below:

feedback-order-and-next-button-alignment

client/components/LearnerViewDirective.js
```diff
@@ -37,99 +37,103 @@ tie.directive('learnerView', [function() {
...
-                    <div class="tie-next-curtain"></div>
-                    <div class="tie-arrow-highlighter"></div>
-                    <div ng-click="showNextTask()" class="tie-next-arrow">
-                      <span class="tie-next-button-text">Next</span>
-                    </div>
-                  </div>
                 </div>
               </div>
+              <select class="tie-select-menu" name="question-set-select"
+                      ng-change="changeQuestionSet(currentQuestionSetId)" ng-model="currentQuestionSetId"
```
rabidbit: nit pick: 80 chars or less for line, if possible.

rabidbit: @seanlip I personally find it more readable to do 4 space indentation for continuation of lines. Do you still want to stick with 2 space indentation for everything, including continuation of lines?

seanlip: For continuation lines of HTML tags only, I prefer to have the attrs aligned as they are now (i.e. left edge of "ng-change" aligned with left edge of "class"). So I think the current version of lines 96-97 is fine in that regard. This helps distinguish from actual nesting of HTML tags within others, which we do with 2-space indentation.

The reason for using 2-space rather than 4-space is that HTML tags can get pretty nested. So I would lean towards the former in that case.

Does the above sound OK?

rabidbit: SGTM for html tags. Re nesting, I meant only use 4 spaces in cases where line exceeds 80 characters and was broken up (otherwise, use 2); didn't mean 4 spaces for all lines.

seanlip: Ah, understood. OK, thanks!

theresa-lee:
Okay, I've re-merged with master with some of the new stuff, and I think this solution works but may need some additional testing. 

I've basically removed the instruction that used to be in the old window that pointed you to other instruction window because it was just redundant at this point. It still has the congratulatory one when you finish though. Also the next button CSS should be fixed.

theresa-lee:
Hrm, still need to make some changes because the merging has caused a bit of a mess.

theresa-lee:
Okay, I think for the most part I've cleaned it up, but there might be slight issues with the look of the new reinforcement stuff which I can probably handle in a separate PR.

rabidbit:
Looks great!

seanlip:
Um, I thought we were only going to merge this once we've confirmed it should replace the current version?

I guess I could create a new branch for the previous three-window version if we still want to test...

rabidbit:
Yeah, I think it was a pain for theresa-lee to keep up with the branch syncs, we seem to be planning on the 2 window design, we had users and UX person point to 2 window, and we have prior commits we can test on. We can do the quick internal user checks before the end of the week to confirm. Feel free to use the internal team mail for further discussion on this.

Issue #271, opened 2017-04-27
time.time used in performance tests.


BenjaminHolland:
This is the most general method, but also probably the least accurate choice for the purposes of testing algorithm performance, which may account for some of the wonkyness of the performance tests.  

time.perf_counter would seem to be the most appropriate choice, being the highest available resolution and being meant for short duration timing rather than long running or user-facing clocks.  However, this is only available in python 3. (I haven't quite figured out which python version we're using yet.)

if we're using 2.7, time.clock may be a better option, as it seems to have better resolution. Apparently this is the function used to analyze python library algos. 

https://docs.python.org/2.7/library/time.html
https://docs.python.org/3.6/library/time.html

BenjaminHolland:
Attaching additional reference material regarding the specific implementation used in Skulpt. Confirms that the implementation of clock and time are different, and that the resolution of time is less than that of clock.

https://github.com/skulpt/skulpt/blob/master/src/lib/time.js

seanlip:
Fixed in #272. Thanks @BenjaminHolland!

(Btw, if you start the PR comment with "Fix #issuenum: ...", GitHub will auto-close the issue when the PR is merged. Just fyi!)

PR #272, opened 2017-04-27
Fix for Issue #271

BenjaminHolland:
Small change, but may help if we're still wanting to get performance testing back up.


BenjaminHolland:
Also starting small seems like a good idea. I'm still getting used to the whole "Legit Project I Don't Control" workflow.

seanlip:
Thanks! Please remove console.logs. You can see a diff of what's changing by clicking on the "Files Changed" tab in GitHub: https://github.com/google/tie/pull/272/files -- it's a good idea to make sure that's correct and exactly what you intend, before getting a review.

(Also, passing to @eyurko for review, once you've got this PR finalized.)

BenjaminHolland:
Whoops! Forgot I'd done that.

client/services/code_preprocessors/PythonCodePreprocessorService.js
```diff
@@ -131,9 +134,11 @@ tie.factory('PythonCodePreprocessorService', [
             break;
           }
           if (lastFunctionNameLocation !== functionNameStart) {
+            
             codeToAdd = functionPrefix;
+            var functionAdded=codeToAdd+code.slice(lastFunctionNameLocation,functionName.length);
```
seanlip: Is this used?

seanlip:
Hi @BenjaminHolland, thanks! As a general rule, could you also try to make sure the diff is as minimal as possible? In other words, the stuff in the "Files changed" tab should contain a few changes from time.time to time.clock, and nothing else.

BenjaminHolland:
I'll check that in the future. A lot of times Im doing commits in quick
breaks during other projects, so I don't always check everything before the
commit as closely as i should.

client/services/code_preprocessors/PythonCodePreprocessorService.js
eyurko: If it is, please add spaces around the + and after the ,

client/services/code_preprocessors/PythonCodePreprocessorService.js
```diff
@@ -274,12 +281,13 @@ tie.factory('PythonCodePreprocessorService', [
...
           outputFunctionName, correctnessTests, buggyOutputTests,
           performanceTests) {
+            
         // Transform the student code (without changing the number of lines) to
         // put it within a class.
         var transformedStudentCode = this._transformCodeToInstanceMethods(
           codeSubmission.getRawCode(), CLASS_NAME_STUDENT_CODE);
         codeSubmission.replace(transformedStudentCode);
-
+
```
eyurko: trailing whitespace

BenjaminHolland:
Regarding the referenced waiting pulls, would you prefer each one be done individually or have them merged into a single one?

seanlip:
Ah -- please see comment in https://github.com/google/tie/issues/277#issuecomment-297873487. Each should be done individually (especially if they're for unrelated things) so that they can be reviewed separately, but you can have multiple branches open at once.

PR #275, opened 2017-04-27
Correctly add class names to student code.

eyurko:


client/services/code_preprocessors/PythonCodePreprocessorService.js
```diff
@@ -26,7 +26,8 @@ tie.factory('PythonCodePreprocessorService', [
       VARNAME_CORRECTNESS_TEST_RESULTS, VARNAME_BUGGY_OUTPUT_TEST_RESULTS,
       VARNAME_PERFORMANCE_TEST_RESULTS, VARNAME_MOST_RECENT_INPUT) {
     var PYTHON_FUNCTION_DEF_REGEX = new RegExp(
-      'def\\s+([A-Za-z_][A-Za-z_0-9]*)\\s*\\(', 'g');
+      'def\\s+([A-Za-z_][A-Za-z_0-9]*\\s*\\()', 'g');
+    var VALID_WHITESPACE_REGEX = new RegExp('[\\s]');
```
seanlip: Why do you need the '[' and ']'?

client/services/code_preprocessors/PythonCodePreprocessorService.js
```diff
@@ -111,6 +112,8 @@ tie.factory('PythonCodePreprocessorService', [
         }
         var matchingString = regexResult[0];
         var functionName = regexResult[1];
+        var functionNameWithoutParen = functionName.substring(
```
seanlip: I would call this functionName, and the previous line functionNameWithParen.

client/services/code_preprocessors/PythonCodePreprocessorService.js
```diff
@@ -126,11 +129,32 @@ tie.factory('PythonCodePreprocessorService', [
...
+            var whitespaceCheckLocation = (
+              lastFunctionNameLocation + functionNameWithoutParen.length);
+            // We assume we found the matched function, but we might have found
+            // a similarly-named other function / variable. User-inputted code
+            // can be ... tricky.
+            var isMatchedFunction = true;
+            // We can safely assume there's a left paren because we matched it
+            // with the regular expression.
+            while (code[whitespaceCheckLocation] != '(') {
+              var isValidWhitespace = VALID_WHITESPACE_REGEX.test(
```
seanlip: I'm a little confused about why you need to do all this. We've already determined that the current location is "function name + possible spaces + open paren", right? If so, why not just scan until the next open paren?

client/services/code_preprocessors/PythonCodePreprocessorService.js
eyurko: Oops, don't, holdover from when I was selecting a greater character set.

client/services/code_preprocessors/PythonCodePreprocessorService.js
eyurko: I was confused about this for a while as well (and honestly still am, so I had to re-test the code), but here's my best explanation and let me know if this doesn't make sense.

The matching string is "match anything that looks like a python function definition _and_ ends with a parenthesis". The problem is that we might capture:

"def encode     ("

And would strip that down to "encode". That's perfectly fine. But, when we start searching for the string in the code we can't do an exact match on the function name, because we might have "encoded()" as a method, and then we would append the StudentCode(). wrapper twice. On the other hand, we can't do an exact match on the function name with the parenthesis because we might have "encode(", "encode (", and "encode   (" in the code, which I guess are all valid Python?

It's only a problem if a function name is a complete substring of another function name, which is uncommon but should still be supported, I guess.

client/services/code_preprocessors/PythonCodePreprocessorService.js
eyurko: Makes sense.

client/services/code_preprocessors/PythonCodePreprocessorService.js
seanlip: Ah, OK, I see, thanks! You might want to add this sort of thing to the Karma test, it feels like a breakage can easily occur here.

I think the reasoning is sound, but would suggest adding a comment to the code explaining this case in particular ("X can be ... tricky" doesn't quite communicate the nuances). Also maybe create a separate isWhitespaceChar(c) method.

Also, do you need a check for "end of student code" in the "while" clause so that there's a guarantee of termination?

eyurko: I did -- we have a test for functions that are similarly-named, now.

And yes, essentially -- we consistently move up lastFunctionNameLocation at the end of the while loop (or any time before we continue), so when we search for the index we _start_ from the new lastFunctionNameLocation such that we're always looking forward in the code string. Eventually, we will either hit no more matches (and return -1) and / or run out of code (and return -1), so we'll break.

eyurko: I also added a specific _checkMatchedFunctionForWhitespace method to break it down, but I briefly forgot to break out the isWhitespaceChar, so I'll do that now.

client/services/code_preprocessors/PythonCodePreprocessorService.js
```diff
@@ -151,6 +165,21 @@ tie.factory('PythonCodePreprocessorService', [
...
+      while (code[whitespaceCheckLocation] != '(') {
+        var isValidWhitespace = VALID_WHITESPACE_REGEX.test(
+          code[whitespaceCheckLocation])
+        whitespaceCheckLocation++;
+        if (!isValidWhitespace) {
+          return false;
+        }
+      }
+      return true;
+    }
```
eyurko: Weird that this doesn't get flagged by the linter.

eyurko:
Reimplemented by #288.

PR #276, opened 2017-04-27
Add some tests to verify that we handle learner code running forever.

eyurko:
Note that these tests take about 3 seconds to run, because of how the timeout is set.


client/services/code_evaluators/CodeRunnerDispatcherServiceSpec.js
```diff
@@ -31,6 +31,37 @@ describe('CodeRunnerDispatcherService', function() {
       };
       expect(errorFunction).toThrowError(Error);
     });
+
+    it('should throw a timeout error if passed infinitely looping code',
+      function() {
+        var code =
+        `while True:
```
seanlip: Maybe use [..., ...].join('') format, for consistency?

client/services/code_evaluators/CodeRunnerDispatcherServiceSpec.js
```diff
@@ -31,6 +31,37 @@ describe('CodeRunnerDispatcherService', function() {
...
+          print 'a'
+        `;
+
+        CodeRunnerDispatcherService.runCodeAsync("python", code).then(
+          function(rawCodeEvalResult) {
+            expect(rawCodeEvalResult.getErrorString()).toBe(
+              'TimeLimitError: Program exceeded run time limit. on line 2')
```
seanlip: Why is there a period in the middle of the sentence? 

Edit: on thinking about it more, it seems like the "on line 2" part of this doesn't make sense. Perhaps we should remove that, in this case?

client/services/code_evaluators/CodeRunnerDispatcherServiceSpec.js
```diff
@@ -31,6 +31,37 @@ describe('CodeRunnerDispatcherService', function() {
...
+          });
+    });
+
+    it('should throw a timeout error if passed infinitely recursing code',
+      function() {
+        var code = [
+        'def forgetToIgnoreSpaces(string):',
```
seanlip: Indent this and the next two lines by 2.

client/services/code_evaluators/CodeRunnerDispatcherServiceSpec.js
```diff
@@ -31,6 +31,37 @@ describe('CodeRunnerDispatcherService', function() {
...
+        `;
+
+        CodeRunnerDispatcherService.runCodeAsync("python", code).then(
+          function(rawCodeEvalResult) {
+            expect(rawCodeEvalResult.getErrorString()).toBe(
+              'TimeLimitError: Program exceeded run time limit. on line 2')
+          });
+    });
+
+    it('should throw a timeout error if passed infinitely recursing code',
```
seanlip: The error message here looks more like "exceeded call stack size" than timeout...

client/services/code_evaluators/CodeRunnerDispatcherServiceSpec.js
eyurko: Done.

client/services/code_evaluators/CodeRunnerDispatcherServiceSpec.js
eyurko: Sounds good.

client/services/code_evaluators/CodeRunnerDispatcherServiceSpec.js
eyurko: I added a check to getErrorString.

client/services/code_evaluators/CodeRunnerDispatcherServiceSpec.js
eyurko: Sure, sounds good.

seanlip:
Couple of things:

- The PR seems to contain a bunch of commits that don't belong to it.
- The review comments don't seem to have been addressed in the "Files changed" diff.

eyurko:
Yes, sorry, I keep trying to avoid bringing in extra commits but the whole

git checkout master
git pull origin master
git checkout -b new-branch

thing seems to still pull in extra commits. Any advice on how to fix that in the future?

Also the comments should be rectified, now -- accidentally ran git status and then git push and forgot to commit in the middle.

client/services/code_evaluators/CodeRunnerDispatcherServiceSpec.js
seanlip: The expected result for this test still doesn't look right?

eyurko: Yikes, you're completely right. Sorry about that -- it's been addressed.

seanlip:
@eyurko -- could you please redo this PR cleanly? The "Commits" tab has extra stuff.

eyurko:
See #289.

PR #287, opened 2017-04-28
Fixed setup.sh file so the e2e tests can be run locally.

theresa-lee:


scripts/setup.sh
```diff
@@ -23,7 +23,7 @@ export OS=`uname`
 export MACHINE_TYPE=`uname -m`
 export TOOLS_DIR=./tools
 export NODE_DIR=$TOOLS_DIR/node-6.9.1
-if [ ${OS}==MINGW_NT-10.0 ]; then
+if [ ${OS}== "MINGW_NT-10.0" ]; then
```
seanlip: Nit: could you please add a space before == as well? Thanks!

Issue #282, opened 2017-04-28
Git hook doesn't run when pushing on Mac


shaman-rajan:
I'm not sure if this is an issue on all OS, but when I push from my mac it doesn't seem to run the tests from the hook script.

### Repro steps:
1. Make edits and commit (I'm using a Macbook Air)
2. Push

### Expected behaviour
Linter and karma tests are run before push

seanlip:
Fixed in #294.

PR #291, opened 2017-05-05
Fix #236 and #282: make pre-push hook run, and make linter stricter.

### LINKED ISSUE #236: ESLint max-len configuration
Enable the [max-len](http://eslint.org/docs/rules/max-len) rule in ESLint to flag long lines. The default limit is 80.

seanlip:
This PR fixes the pre-push hook by setting the directory correctly and by ensuring that the hook is executable. It also removes the previous leniency of the linter for lines that contain strings (causing it to miss several lines that were > 80 chars long).


seanlip:
NB: once this is merged, developers should re-run `bash scripts/setup.sh` on their local machines.

hooks/pre-push
```diff
@@ -50,7 +50,7 @@ bash scripts/run_linter.sh
 check_test_status "Linter"
 
 echo "Running karma tests"
-bash scripts/run_karma_tests.sh --browsers PhantomJS
```
shaman-rajan: The phantomJS browser was here in case the actual browser cannot run (for example when in a SSH session). Removing may cause errors in such cases.

seanlip: Yep -- but it turns out that PhantomJS couldn't use es6, so it was giving errors for the backtick character in template files. So it seemed better to go back to Chrome, rather than have pre-push testing that didn't work at all...

shaman-rajan: I had got around that by using es6-shim ([here](https://www.npmjs.com/package/es6-shim)). [This line](https://github.com/google/tie/blob/master/karma.conf.js#L33) should ensure that errors are not brought up for the incompatibility for any of the questions. If there are other directories that use es6 we can include them there. The 2 lines above that may need to be uncommented for this (comment was added [here](https://github.com/google/tie/commit/1fa187b21eddbdc72f6060e21f2bf81fb02f8f3f#diff-a2a3b7b0c9c3b4b93b4aebf4e3ec3cfb)).

seanlip: Ah, I see the issue. Will fix that. Thanks!

seanlip: Now fixed.

seanlip:
On further investigation, I found a bunch of files that were never linted (see modifications [here](https://github.com/google/tie/pull/291/files#diff-5b10a7fb8b822e317eedb11c21f871e4R28)). This PR now includes lint corrections for these files. PTAL.

PR #300, opened 2017-05-16
Fix #279

### LINKED ISSUE #279: Reset Code button resets feedback window
Expected behavior: When resetting code, only the code is reset. There should be a separate button to reset feedback (it's possible the user wants to reset one but not the other).

Observed behavior: Reset Code button resets both the code and the feedback window.

Steps to reproduce: Click reset code and note that feedback window is also reset.

theresa-lee:
I've move the button to the right question/feedback side, and it mirrors the same location and similar style to the Run Code button. Also we decided that the words 'Start Over' are more clear and consistent than Reset Code.

### DIFF
--- a/client/components/LearnerViewDirective.js
+++ b/client/components/LearnerViewDirective.js
@@ -105,6 +105,10 @@ tie.directive('learnerView', [function() {
                       ng-options="i.themeName as i.themeName for i in themes">
                 <option style="display: none" value="">Theme</option>
               </select>
+              <button class="tie-code-reset protractor-test-reset-code-btn" name="code-reset"
+                ng-click="resetCode()">
+                Start Over
+              </button>
             </div>
             <div class="tie-coding-ui">
               <div class="tie-lang-terminal">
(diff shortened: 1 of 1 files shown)


rabidbit:
I feel the red color draws too much attention to the button and compels the user to try it (or at least focus on it for a bit). I'm actually OK with keeping the button's styling as it was (smaller white button), as it makes the Run button prominent and distinct from the other buttons that are more settings-like/optional.

There are a couple of problems with the original button though. It turns blue on activate (should probably stay white with a faint border change to indicate activation). It also should use explicit styling for the color, border and size, because if you zoom in, the button style changes.

In the longer term, we probably want to consider getting a consistent style across all UI assets, like buttons, drop-downs and such.  That would be a bigger UX change that we could get internal UX review on.

theresa-lee:
Hrm, it looks like the problems with the original might be OSX specific as those things aren't happening on Chrome on my desktop (there's no blue when I click on it on linux, and it also doesn't change when I zoom in). Should I still proceed to make changes specific to what's happening on the mac?

rabidbit:
I think it would be useful to make the style explicit, since it may address a large portion of the user base, it shouldn't break other browsers, and you never know if how a given browser handles default styles changes in the future.

theresa-lee:
Closing because this is obsolete after feedback from users that there should be buttons to clear code and feedback separately.

PR #307, opened 2017-05-23
Fix #274 : Syntax errors now stored as feedback

### LINKED ISSUE #274: Syntax errors only display the last error
Expected behavior: when running into multiple syntax errors, all past errors can be viewed. This is important for the alternative UI (#268), which maintains a history of feedback.

Observed behavior: In alternative UI, past syntax error feedback is displayed, but only the most recent error message can be viewed.

Steps to reproduce: In alternative UI (#268), generate multiple syntax errors and try to view all the errors. Notice only details of the last error is displayed.

theresa-lee:
Since we were already storing feedback, I thought it would make sense to store the syntax error messages as well with the other feedback but had to add another kind of paragraph (error paragraphs).


client/components/LearnerViewDirective.js
```diff
@@ -72,15 +72,17 @@ tie.directive('learnerView', [function() {
                       <span ng-if="paragraph.isCodeParagraph()">
                         <code-snippet content="paragraph.getContent()">test</code-snippet>
                       </span>
+                      <span ng-if="paragraph.isErrorParagraph()">
+                        <a href class="tie-feedback-syntax-error-link"
+                          ng-click="showErrorString = !showErrorString">
```
seanlip: I lean against doing anything more than calling a single function in ng-click. Can we keep the toggleSyntaxErrorHint() and isSyntaxErrorShown (or perhaps syntaxErrorIsShown) terminology that we used to have? The toggleSyntaxErrorHint() function could just flip the boolean flag.

client/components/LearnerViewDirective.js
```diff
@@ -584,12 +581,7 @@ tie.directive('learnerView', [function() {
             var syntaxErrorIndex = feedback.getSyntaxErrorIndex();
```
seanlip: Do we even need the concept of "syntaxErrorIndex" any more? If not, would it be worth trying to remove it from the entire codebase?

theresa-lee: Mm for the reinforcement strings to show up, we rely on knowing whether or not there was a syntax error. So I left it because although we don't specifically need the index, the fact that if there isn't an index shows that there wasn't a syntax error. Should I change it to a boolean instead?

seanlip: A boolean sounds better, yep. But I'm also wondering whether this can be computed on the fly -- i.e. have a hasSyntaxError() method on the feedback object and have that check the types of the feedback paragraphs and compute the appropriate boolean?

client/components/LearnerViewDirective.js
```diff
@@ -72,15 +72,17 @@ tie.directive('learnerView', [function() {
                       <span ng-if="paragraph.isCodeParagraph()">
                         <code-snippet content="paragraph.getContent()">test</code-snippet>
                       </span>
+                      <span ng-if="paragraph.isSyntaxErrorParagraph()">
+                        <a href class="tie-feedback-syntax-error-link"
+                          ng-click="toggleSyntaxErrorHint(this)">
```
seanlip: Hmm. Could we do `ng-click="toggleSyntaxErrorHint(paragraph)"`?

client/components/LearnerViewDirective.js
theresa-lee: Okay, per our conversation, I removed the index altogether and just am doing a simple forloop to check if there's a syntax error. Also renamed the function to be more specific.

seanlip:
Oh, also, there are merge conflicts, but I guess you might not have finished making code changes yet :) Let me know whenever you'd like another review!

client/components/LearnerViewDirective.js
theresa-lee: I think it needs to be 'this' because it refers to that specific scope that the ng-repeat creates?

seanlip: For posterity: `this` actually seems to refer to some Angular construct and not the feedback paragraph, so we decided to change it to `paragraph` instead.

PR #310, opened 2017-05-26
Fix #309: Do not compute reinforcements when there is a runtime error.

### LINKED ISSUE #309: Persistent thinking dots when issue with parsing code
Expected behavior: Floating dots (the dots UI that indicated that the application is "thinking" / "processing") do not display indefinitely.

Observed behavior: There are cases when the floating dots display indefinitely when there is an error in the user's code, even after "Reset Code" is pressed. This can prevent other feedback from displaying.

Steps to reproduce: Simple example is return ss in the first common char question.

High Priority: * This may be impacting user study results! *

seanlip:
I introduced this bug in the #304 refactor, sorry. /cc @eyurko


rabidbit:
This appears to fix the issue based on my test! Adding Eric for another pair of eyes.

PR #316, opened 2017-05-31
Get coverage reports working on GitHub.

seanlip:


seanlip:
NB: I think the coverage "cross mark" will go away once we have a coverage report for the `master` branch, so it can probably be ignored.

PR #319, opened 2017-06-02
Add Tests for client/services/code_evaluators folder

dchen97:
* Added tests for code coverage for the directory `client/services/code_evaluators`
* There is still one line that is not covered in `PythonCodeRunnerService.js`, but will try to see if it gets covered by expanding tests in other files.

### DIFF
--- a/client/services/code_evaluators/PrereqCheckDispatcherServiceSpec.js
+++ b/client/services/code_evaluators/PrereqCheckDispatcherServiceSpec.js
@@ -43,7 +43,7 @@ describe('PrereqCheckDispatcherService', function() {
 
     it('throws an error when an unsupported language is used', function() {
       expect(function() {
-        PrereqCheckDispatcherService.checkCode('java', '', '');
+        PrereqCheckDispatcherService.checkCode('invalidLanguage', '', '');
       }).toThrow();
     });
   });
--- a/client/services/code_evaluators/PrereqCheckDispatcherServiceSpec.js
+++ b/client/services/code_evaluators/PrereqCheckDispatcherServiceSpec.js
@@ -1,0 +1,50 @@
+// Copyright 2017 The TIE Authors. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS-IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+/**
+ * @fileoverview Unit tests for the PrereqCheckDispatcherServiceSpec.
+ */
+describe('PrereqCheckDispatcherService', function() {
+  var PrereqCheckDispatcherService;
+  var LANGUAGE_PYTHON = 'python';
+
+  beforeEach(module('tie'));
+  beforeEach(inject(function($injector) {
+    PrereqCheckDispatcherService = $injector.get(
+      'PrereqCheckDispatcherService');
+  }));
+
+  describe('checkCode', function() {
+    it('correctly calls PythonPrereqCheckService when the language is python',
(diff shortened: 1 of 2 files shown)


client/services/code_evaluators/PrereqCheckDispatcherServiceSpec.js
```diff
@@ -0,0 +1,50 @@
...
+        ].join('\n');
+        var prereqCheckResults = PrereqCheckDispatcherService.checkCode(
+          LANGUAGE_PYTHON, code, code);
+        expect(prereqCheckResults).toEqual(null);
+      }
+    );
+
+    it('throws an error when an unsupported language is used', function() {
+      expect(function() {
+        PrereqCheckDispatcherService.checkCode('java', '', '');
```
seanlip: Let's use a string like 'invalidLanguage' instead, so that when we (hopefully) support Java in the future we won't have to amend this.

client/services/code_evaluators/PrereqCheckDispatcherServiceSpec.js
```diff
@@ -0,0 +1,50 @@
...
+    PrereqCheckDispatcherService = $injector.get(
+      'PrereqCheckDispatcherService');
+  }));
+
+  describe('checkCode', function() {
+    it('correctly calls PythonPrereqCheckService when the language is python',
+      function() {
+        var code = [
+          'def myFunction(arg):',
+          '\treturn result',
```
seanlip: Just to check -- in some tests we use 4 spaces and in other tests we use \t. What is it in practice, and can we stick to that throughout? I seem to remember we standardized it in the frontend (so that if the user hits Tab it gets converted to 4 spaces, or the other way round).

client/services/code_evaluators/PythonPrereqCheckServiceSpec.js
```diff
@@ -156,4 +156,75 @@ describe('PythonPrereqCheckService', function() {
...
+
+  describe('checkCode', function() {
+    it(['returns the correct PrereqCheckFailureObject when starter code is ',
+      'missing'].join(''), function() {
+      var starterCode = [
+        'def myFunction(arg):',
+        '\treturn arg',
+        ''
+      ].join('\n');
+      var prereqCheckFailure = PythonPrereqCheckService.checkCode(starterCode,
```
seanlip: If the contents of (...) span multiple lines, break after the '('.

client/services/code_evaluators/PythonPrereqCheckServiceSpec.js
```diff
@@ -156,4 +156,75 @@ describe('PythonPrereqCheckService', function() {
...
+          '');
+      expect(prereqCheckFailure.isMissingStarterCode()).toEqual(true);
```
seanlip: Add some additional checks for the case where the starter code has the right method name but the wrong args, and the wrong method name but the right args.

client/services/code_evaluators/PrereqCheckDispatcherServiceSpec.js
theresa-lee: It looks like when using tab, it gets converted to 4 spaces. Does that mean we should be using 4 spaces instead of \t?

theresa-lee: Discussed with Sean, we should use 4 spaces instead!

dchen97: Done!

client/services/code_evaluators/PythonPrereqCheckServiceSpec.js
```diff
@@ -156,4 +156,116 @@ describe('PythonPrereqCheckService', function() {
...
+
+  describe('getUnsupportedImports', function() {
+    it('correctly returns a list of the unsupported imports', function() {
+      var code = [
+        'import numpy',
+        'import pandas',
+        'import math',
+        '',
+        'def myFunction(arg):',
+        '\treturn arg',
```
seanlip: Shouldn't these and others be changed as well (tabs-to-spaces)?

dchen97: All these tabs should now be spaces.

seanlip:
Thanks! LGTM.

Issue #323, opened 2017-06-07
Handling calls to template function


rabidbit:
User noticed the following:

Input code:

```
def reverseWords(s):
    return s

reverseWords('hi')
```

Results in the following response:

> Looks like your code had a runtime error when evaluating the input None. Here's the trace:
> 
> `NameError: name 'StudentCode' is not defined on line 4`

Input code:

```
def reverseWords(s):
    reverseWords('hi')
```

Results in the following response:

> Looks like your code had a runtime error when evaluating the input "moo cow bark dog". Here's the trace:
> 
> `ExternalError: RangeError: Maximum call stack size exceeded on line 2`

Should we consider special handling of calls to the current function like this?

seanlip:
Discussed with @eyurko:
- For the second one, we can change the message to: "Looks like your code is hitting an infinite recursive loop. Check to see that your recursive calls terminate."
- For the first one, we can say: "Please keep your code within the existing predefined functions -- we can't process code in the global scope."

rabidbit:
SGMT. Thanks for looking into this!

rabidbit:
Verified UI appears to work as described (UI behavior LGTM).

seanlip:
Fixed in #324. Thanks @dchen97!

PR #324, opened 2017-06-08
Fix issue #323

dchen97:
* Resolves the issues brought up in Issue #323 
    * Adds a Prereq error type for when there is code declared in the global scope
    * Throws the necessary Prereq error when code is declared in the global scope
    * Adds feedback for when there is an infinite recursive loop
* Added the associated tests

### DIFF
--- a/client/services/code_evaluators/PythonPrereqCheckService.js
+++ b/client/services/code_evaluators/PythonPrereqCheckService.js
@@ -36,7 +36,7 @@ tie.factory('PythonPrereqCheckService', [
       var codeLines = code.split('\n');
       for (var i = 0; i < codeLines.length; i++) {
         var line = rightTrim(codeLines[i]);
-        var pattern = /(^\s\s+(\w|\s)*)|((^\s\s\s\s)+(\w|\s)*)|(^def)|(^import)/
+        var pattern = /(^\s\s+)|((^\s\s\s\s)+)|(^def)|(^import)/;
         if (line.search(pattern) === -1 && line !== '') {
           return true;
         }
--- a/client/services/code_evaluators/PythonPrereqCheckService.js
+++ b/client/services/code_evaluators/PythonPrereqCheckService.js
@@ -36,8 +36,8 @@ tie.factory('PythonPrereqCheckService', [
       var codeLines = code.split('\n');
       for (var i = 0; i < codeLines.length; i++) {
         var line = rightTrim(codeLines[i]);
-        if (line.search(/((^\s\s\s\s)+(\w|\s)*)|(^def)|(^import)/) === -1 &&
-          line !== '') {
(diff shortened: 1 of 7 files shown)


seanlip:
Ah, sorry, I just merged your other PR. Mind merging master into this branch and re-pushing in order to resolve the merge conflicts?

Thanks!

seanlip:
Thanks @dchen97! Adding @rabidbit for UI verification and @theresa-lee for initial review pass.

dchen97:
All merge conflicts should be resolved now!

rabidbit:
Actually, I see something odd. If you enter something like this:

```
def reverseWords(s):
  reversed_word = ''
  reversed_string = ''
  for character in s:
    if not character.isspace():
      reversed_word = character + reversed_word
    else:
      reversed_string += reversed_word
      reversed_word = ''
      reversed_string += character
  reversed_string += reversed_word
  return reversed_string
```

You see:

> Please keep your code within the existing predefined functions -- we cannot process code in the global scope.

Which is unexpected.

rabidbit:
Interestingly, whitespace appears to impact this(?)

dchen97:
Hmmm so I think the problem is that the tabs are supposed to be parsed as 4 spaces, but in that example, there are only 2 spaces in the tab, so my RegEx doesn't recognize the "tabbed" lines.

dchen97:
Should I change it so it recognizes tabs as 2 and/or 4 spaces?

dchen97:
OK I've adjusted the RegEx to accept tabs that are 2 spaces instead of 4 as well as a simple test to double check.

seanlip:
Hm. If there is even at least one space, it stops being in global scope, right?

dchen97:
Hmm I guess so. I'll adjust it

dchen97:
Done! Now even one space passes the Prereq Check

rabidbit:
Verified UI appears to work as described (UI behavior LGTM).

client/services/code_evaluators/PythonPrereqCheckService.js
```diff
@@ -20,16 +20,30 @@
...
       // Remove trailing white space at end of string.
       var RIGHT_TRIM_PATTERN = /\s+$/;
       return str.replace(RIGHT_TRIM_PATTERN, '');
     };
 
+    var checkGlobalCallsPresent = function(code) {
+      var codeLines = code.split('\n');
+      for (var i = 0; i < codeLines.length; i++) {
+        var line = rightTrim(codeLines[i]);
+        var pattern = /(^\s+)|(^def)|(^import)/;
```
theresa-lee: Can we rename pattern to something like VALID_LINE_PATTERN or something so that it's more clear what the regex matches?

dchen97: Done!

PR #325, opened 2017-06-10
Feedback is now stored in localStorage 

theresa-lee:
- Added a reset feedback button that will reset the feedback window only 
- Feedback is JSON-ified and then reconstructed 
- Changed the CodeStorageService to ServiceLocalStorage 

Things that related to this that I will fix after this:
- Checkmarks/status for completed questions aren't saved properly
- Styling/Positioning of the reset buttons
- Reinforcement saving? (Am unsure if we want to save this)


client/services/LocalStorageService.js
```diff
@@ -0,0 +1,122 @@
...
+    } catch (e) {
+      localStorageIsAvailable = false;
+    }
+
+    var getLocalStorageKeyForCode = function(questionId, language) {
+      return questionId + ":" + language;
+    };
+
+    var getLocalStorageKeyForFeedback = function(questionId, language) {
+      return questionId + ":feedback:" + language;
```
seanlip: Do we actually need the language here? If the user changes the language, does the existing feedback disappear? If not, what happens to it?

Also, the discrepancy between the key for code and the key for feedback is a bit odd. I think we may want to backtrack a bit and make a short doc that outlines the desired namespace structure for localStorage (for storing question completion, code, feedback). What conventions would we use if we were coming up with this from scratch, now that we have a better idea of all the data we need to persist?

client/services/LocalStorageService.js
```diff
@@ -0,0 +1,122 @@
...
+        if (storedFeedback === null) {
+          return null;
+        }
+        var storedFeedbackStorage = [];
+        var feedbackParagraphs = [];
+
+        var rawFeedback = JSON.parse(storedFeedback);
+        for (var i = 0; i < rawFeedback.length; i++) {
+          for (var j = 0; j < rawFeedback[i].feedbackParagraphs.length; j++) {
+            if (rawFeedback[i].feedbackParagraphs[j]._type === "text") {
```
seanlip: All this should be replaced with a single FeedbackParagraphObjectFactory.createFromJson(...) call. The detailed reconstitution is not something that should be a part of this service.

client/services/LocalStorageService.js
```diff
@@ -0,0 +1,122 @@
...
+        if (!localStorageIsAvailable) {
+          return null;
+        }
+        var localStorageKey = getLocalStorageKeyForFeedback(
+          questionId, language);
+        var storedFeedback = localStorage.getItem(localStorageKey);
+        if (storedFeedback === null) {
+          return null;
+        }
+        var storedFeedbackStorage = [];
```
seanlip: I don't think I understand the semantics of this naming choice. `reconsistutedFeedback`, perhaps?

seanlip:
@theresa-lee -- is this obsolete?

theresa-lee:
Yes. This is obsolete now that #381 is in.

PR #328, opened 2017-06-13
Code submission tests

eledavi:
Corrected variable IMPORT_PATTERN so line 86 of CodeSubmissionObjectFactory.js evaluates to true and lines 87-93 are run

### DIFF
--- a/client/domain/CodeSubmissionObjectFactory.js
+++ b/client/domain/CodeSubmissionObjectFactory.js
@@ -164,7 +164,7 @@ tie.factory('CodeSubmissionObjectFactory', [
       var insertPos = 0;
       for (var i = 0; i < this._preprocessedCodeLines.length; i++) {
         if (this.IMPORT_PATTERN.test(this._preprocessedCodeLines[i])) {
-          var importLine = this._preprocessedCodeLines[i].slice(0);
+          var importLine = this._preprocessedCodeLines[i];
           this._preprocessedCodeLines.splice(i, 1);
           this._preprocessedCodeLines.splice(insertPos, 0, importLine);
           this._rawCodeLineIndexes.splice(i, 1);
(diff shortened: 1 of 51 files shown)


client/domain/CodeSubmissionObjectFactory.js
```diff
@@ -164,8 +164,7 @@ tie.factory('CodeSubmissionObjectFactory', [
       var insertPos = 0;
       for (var i = 0; i < this._preprocessedCodeLines.length; i++) {
         if (this.IMPORT_PATTERN.test(this._preprocessedCodeLines[i])) {
-          var importLine = this._preprocessedCodeLines[i].slice(
-              this.TAB_LENGTH);
+          var importLine = this._preprocessedCodeLines[i].slice(0);
```
seanlip: Are you sure this change is correct? Did you test manually?

eledavi: I retested it and corrected it. New commit should be good.

client/domain/CodeSubmissionObjectFactory.js
```diff
@@ -68,7 +68,7 @@ tie.factory('CodeSubmissionObjectFactory', [
      * @type {RegExp}
      */
     CodeSubmission.prototype.IMPORT_PATTERN = new RegExp(
-      '^\\ {4}import\\ \\w+$');
+      '^\\ {0}import\\ \\w+$');
```
theresa-lee: Per our conversation, {0} is a no op.

eledavi:
I corrected the regex variable. @theresa-lee

theresa-lee:
Turns out that part of the code doesn't actually need to be tested anymore because that function is not necessary (being deleted in #358).

PR #331, opened 2017-06-14
Fix #306 Add Privacy Notice

### LINKED ISSUE #306: Add "we don't store any of your code" notice to the browser-only version

dchen97:
Addresses #306 

* Added a privacy notice that we will not store code if the user is on the browser only version.
    * Will appear on the bottom of the question window and below the feedback (if there are any). (**Note:** I wasn't sure where exactly to place it so I'm open to suggestions.)
    * Will appear as: "**Privacy Notice:** Your code is only stored on your browser's local storage. In other words, we _do not_ store your code on any of our servers or databases." (**Note:** Am also open to suggestions for wording.)

### DIFF
--- a/client/components/LearnerViewDirective.js
+++ b/client/components/LearnerViewDirective.js
@@ -92,13 +92,6 @@ tie.directive('learnerView', [function() {
                   </div>
                   <br>
                 </div>
-                <div class="tie-privacy-notice" ng-show="isBrowserOnly">
-                  <p>
-                    <b>Privacy Notice:</b> Your code is only stored on your 
-                    browser's local storage. In other words, we <em>do not</em> 
-                    store your code on any of our servers or databases.
-                  </p>
-                </div>
               </div>
               <select class="tie-select-menu" name="question-set-select"
                       ng-change="changeQuestionSet(currentQuestionSetId)" ng-model="currentQuestionSetId"
@@ -149,6 +142,15 @@ tie.directive('learnerView', [function() {
             </div>
           </div>
         </div>
+        <div class="tie-footer">
+          <ul>
+            <li class="about-tie button">
+              <a href="https://github.com/google/tie">About TIE</a>
+            </li>
+            <li class="privacy button" ng-click="onPrivacyClick()">
+              <a href="#">Privacy</a>
+            </li>
+          </ul>
       </div>
       <style>
         html {
@@ -486,6 +488,35 @@ tie.directive('learnerView', [function() {
         .tie-wrapper.night-mode {
           background-color: #212121;
         }
+        
+        .tie-footer {
+          background-color: #d5d5d5;
+          position: absolute;
+          bottom: 0;
+          left: 0;
+          width: 100%;
+        }
+        
+        .tie-footer ul {
+          list-style-type: none;
+          margin: 0;
+        }
+        
+        .tie-footer li {
+          float: left;
+          padding: 0;
+        }
+        
+        .tie-footer a {
+          text-decoration: none;
+          color: #5A5A5A;
+          display: block;
+          padding: 10px;
+        }
+        
+        .tie-footer li:hover {
+          background-color: #969696;
+        }
       </style>
     `,
     controller: [
@@ -532,8 +563,6 @@ tie.directive('learnerView', [function() {
           code: ''
         };
 
-        $scope.isBrowserOnly = !SERVER_URL;
-
         var autosaveCancelPromise;
         var cachedCode;
         var congratulatoryFeedback = FeedbackObjectFactory.create();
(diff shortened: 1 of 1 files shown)


client/components/LearnerViewDirective.js
```diff
@@ -524,6 +531,9 @@ tie.directive('learnerView', [function() {
         $scope.editorContents = {
           code: ''
         };
+
+        $scope.isBrowserOnly = SERVER_URL !== null;
```
seanlip: `!SERVER_URL` probably suffices.

seanlip:
Thanks @dchen97! I just took a look at the code; adding @rabidbit for UI review/approval.

client/components/LearnerViewDirective.js
dchen97: Hmm when I seem to do that, it shows up in the non-browser version, although I'm not totally sure why. :confused:

dchen97: Sorry misunderstood what "isBrowserOnly" should mean but after going over with @seanlip it works now.

rabidbit:
Thanks for working on this. We may need to work on the wording, but displaying a privacy statement would be good.

I don't think the feedback window is the place for a privacy statement. I think we can have two links on the bottom of the page below the row of buttons. One "Privacy" link in small font (which pops up a privacy statement) and one "About TIE" (which points to TIE's GitHub repo). See bottom of https://www.google.com/ as an example.

Note that here are two UIs requiring privacy statements (server version and client-only version). The text will be different for each.

dchen97:
So I've pushed up changes to add a footer to the bottom of the page with 2 links - one "About TIE" that directs to the GitHub and a "Privacy" button that simply has a popup alert with the correct privacy notice. This is how the UI looks now:

New TIE UI screenshot

However, now I'm getting E2E build errors and it seems to have to do with the footer messing with the test's ability to click on the `Reset Code` button? Would anyone happen to know how I can fix this because I don't totally understand how this testing framework works.

rabidbit:
Could you make the font smaller, remove the background color and border, and position them right under the button controls, fully-justified, and scaling with the application size? Like this:

screen shot 2017-06-15 at 10 20 03 am

Thanks!

dchen97:
I've made adjustments you suggested @rabidbit . Here's a screenshot of the updated version:

image

Does this look good?

rabidbit:
Also, I just realized that https://github.com/google/tie/blob/master/README.md might be a better place to point for "About TIE" (as opposed to https://github.com/google/tie). Thanks again!

rabidbit:
One last request. For "About TIE", could you open in new tab rather than redirect the current tab? This will maintain the appearance of the division between the local client app and anything else on the Internet.

As for the pop-up, given future potential work in creating modals and such, we can go with what you have, if you open an issue to fix that later.

Thanks!

dchen97:
The `About TIE` button should now open a new tab!

rabidbit:
Thanks dchen97!

This looks good to me from a UX/functional standpoint.

I did notice that the CSS rules appears to be appended at the end. The convention is to order all style rules and attributes, unless it affects style cascading. Deferring to theresa-lee for any comments from a FE perspective.

client/components/LearnerViewDirective.js
```diff
@@ -140,9 +140,18 @@ tie.directive('learnerView', [function() {
...
+                  <a target="_blank" href="https://github.com/google/tie/blob/master/README.md">About TIE</a>
+                </li>
+                <li class="privacy button" ng-click="onPrivacyClick()">
+                  <a href="#">Privacy</a>
+                </li>
+              </ul>
+            </div>
           </div>
         </div>
-      </div>
```
theresa-lee: Did we have an extra closing div before?

dchen97: Sorry it was hard to tell which divs aligned with which. Fixed it now.

rabidbit:
We'll need to add a Terms page later. I have a slight preference for fully justified, as it appears more symmetric and there is a distinction between privacy and legal docs verses the general about doc, but I don't have strong feelings against things on the left or middle. I do have strong feelings against having everything in the center or the right.

Slight preference for:

screen shot 2017-06-16 at 10 50 39 am

OK with:

screen shot 2017-06-16 at 10 51 03 am

OK with:

screen shot 2017-06-16 at 11 02 15 am


Really dislike:

screen shot 2017-06-16 at 10 41 28 am

theresa-lee:
Okay, I don't feel too strongly about having it be on the left side or centered, so I think keeping it on both sides is fine then.

client/components/LearnerViewDirective.js
```diff
@@ -159,9 +169,15 @@ tie.directive('learnerView', [function() {
         div.CodeMirror span.CodeMirror-matchingbracket {
           color: rgb(75, 206, 75);
         }
-        .tie-arrow-highlighter {
-          background-color: white;
-          border-radius: 100px;
+        .about-tie.button {
```
seanlip: .tie-about-button

client/components/LearnerViewDirective.js
```diff
@@ -159,9 +169,15 @@ tie.directive('learnerView', [function() {
         div.CodeMirror span.CodeMirror-matchingbracket {
           color: rgb(75, 206, 75);
         }
-        .tie-arrow-highlighter {
-          background-color: white;
-          border-radius: 100px;
+        .about-tie.button {   
+          float: left;        
+        }                     
+        .privacy.button {
```
seanlip: .tie-privacy-button

client/components/LearnerViewDirective.js
dchen97: Done!

seanlip:
LGTM. Thanks!

PR #336, opened 2017-06-16
Various content fixes.

seanlip:
- mostCommonCharacter: drop unicode portion of question.
- reverseWords: detect case when entire string is reversed and provide specific feedback for that case.
- i18n: reword feedback message for "corner" case to sound less like a gotcha

### DIFF
--- a/assets/questions/findMostCommonCharacter.js
+++ b/assets/questions/findMostCommonCharacter.js
@@ -102,31 +102,5 @@ class AuxiliaryCode(object):
       ]
     }],
     performanceTests: []
-  }, {
-    instructions: [{
-      content: [
-        'Now, make sure that your code works for ',
-        'different string encodings. ',
-        'What if the provided string is unicode, rather than ASCII?'
-      ].join(''),
-      type: 'text'
-    }],
-    prerequisiteSkills: ['Arrays', 'Strings', 'Hash Maps'],
-    acquiredSkills: ['String Manipulation'],
-    inputFunctionName: null,
-    outputFunctionName: null,
-    mainFunctionName: 'findMostCommonCharacter',
-    correctnessTests: [{
-      input: '\u0041\u0042\u0043\u0041',
-      allowedOutputs: ['A'],
-      tag: 'the general case'
-    }],
-    buggyOutputTests: [],
-    performanceTests: [{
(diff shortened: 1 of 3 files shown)


seanlip:
Hi @eyurko @rabidbit -- friendly ping! Can I get a review for this please?

assets/questions/reverseWords.js
```diff
@@ -93,6 +97,14 @@ globalData.questions['reverseWords'] = {  // eslint-disable-line dot-notation
           "last reversed word."
         ].join('')
       ]
+    }, {
+      buggyFunctionName: 'AuxiliaryCode.reverseString',
+      messages: [
+        [
+          "Please re-read the question. You need to reverse each word ",
```
eyurko: I would drop the "please re-read the question" -- it feels a bit aggressive. Maybe "You're reversing the entire string, but you should only be reversing the words, not their ordering. Check the example and try again." or something?

assets/questions/findMostCommonCharacter.js
```diff
@@ -102,31 +102,5 @@ class AuxiliaryCode(object):
...
-    inputFunctionName: null,
-    outputFunctionName: null,
-    mainFunctionName: 'findMostCommonCharacter',
-    correctnessTests: [{
-      input: '\u0041\u0042\u0043\u0041',
-      allowedOutputs: ['A'],
-      tag: 'the general case'
-    }],
-    buggyOutputTests: [],
-    performanceTests: [{
```
eyurko: I think dropping this will drop performance tests from the entire problem. Can you move this further down in the problem?

seanlip: Good catch, thanks! Further up, you mean? Done.

assets/questions/reverseWords.js
seanlip: Good catch, thanks -- I wanted to highlight that, in general, reading the question correctly is an important thing to do, but I agree that the way it's worded here comes off wrong. I've changed it to the following; is it OK?

> Whoops! It looks like you're reversing the entire string, but you should only be reversing each word in turn, not their ordering. Please check the given example, and try again.

seanlip:
@eyurko ping?

Issue #333, opened 2017-06-15
Explicit styling for buttons


rabidbit:
With the exception of the "Run" button, the styling of button controls are not consistent when page is zoomed. We should probably use explicit styling to keep the control appearance consistent.

![screen shot 2017-06-15 at 10 25 13 am](https://user-images.githubusercontent.com/25493966/27193845-42536ef2-51b5-11e7-9d20-754c12087b88.png)

![screen shot 2017-06-15 at 10 25 27 am](https://user-images.githubusercontent.com/25493966/27193848-45bb8b4c-51b5-11e7-8114-ee4ae3bd6da1.png)

dchen97:
Just checking but for this issue do we just want to address the "Reset Code" button or should we also style the dropdowns?

rabidbit:
Both (explicit styling for all controls). The Run button is OK since it already has explicit styling.

seanlip:
Fixed in #337. Thanks @dchen97!

PR #337, opened 2017-06-16
Fix Issue 333 - Consistent Button Styling

dchen97:
Fix for Issue #333 
* Added explicit styling for select menus and buttons to make UI more consistent.

This is what the UI looks like right now:
image

### DIFF
--- a/client/components/LearnerViewDirective.js
+++ b/client/components/LearnerViewDirective.js
@@ -132,7 +132,7 @@ tie.directive('learnerView', [function() {
                 <div class="tie-code-auto-save" ng-class="{'night-mode': isInDarkMode}" ng-show="autosaveTextIsDisplayed">
                   Saving code...
                 </div>
-                <button class="tie-run-button tie-button blue protractor-test-run-code-btn"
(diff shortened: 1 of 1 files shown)


rabidbit:
Ooo... Nice! Like the overall look! In terms of color, however, would not use gray since it often implies a disabled feature/button. Can you make them white or a light, off-white color?

rabidbit:
Adding @theresa-lee in case she has additional comments.

dchen97:
OK I've changed the background colors of the buttons to be white. This is what the UI looks like now:

image

rabidbit:
Let's not do a gray background color at all for controls (reserve that for disabled features). Instead, you could use a 1px gray border on hover.

In terms of the code, it looks like you are using a class called "blue". Convention is to prepend "tie-" to all classnames to avoid generic names. I see that you are using multi-class selectors for blue, but I'm inclined to stick with convention. What do you think @theresa-lee ?

dchen97:
OK I've gotten rid of the gray background colors altogether and changed the `.blue` class to be `.tie-blue`. Below is a GIF showing what happens to each button on hover.

recording

rabidbit:
LGTM! Thanks dchen97!

PR #341, opened 2017-06-20
Fix #334 - Add Modal Component

### LINKED ISSUE #334: Create Custom Modal Component
**Steps for Reproduction:** 
* Pull down `add-privacy-notice` branch (if not already merged into `master` branch).
* Open the `app.html` file in Chrome
* Click on link labeled `Privacy` in the bottom right corner.
* Notice that a pop-us is displayed with a "This page says..." title at the top.

The issue is that we don't want this title to appear, but since it's a browser-specific behavior, we need to make our own custom modals that can ideally be further extended to other use cases (i.e. Clippy pop ups).

dchen97:
Fix to Issue #334 
* Added a styled Modal directive that should be appended to the Learner View
* Added 2 methods `showModal()` and `showExampleModal()` so that modal's contents can be changed as needed
* Added click handlers to allow user to close modal when they click `OK` button or by clicking outside of the Modal window

screenshot 1

# To Test
* Add the following to the bottom of the LearnerView template HTML:           
`<a href="#" ng-click="showExampleModal()">Show Example Modal</a>`
* Then click on the link and observe that a modal appears and closes as expected.

# EDIT:
I've merged in the changes to the master so that the `Privacy` button will now present a modal with the appropriate message depending on if it's browser-only or not. It should look like this:

screenshot 2

Behavior should otherwise be the same.


theresa-lee:
I think this would be better as a directive in itself. You can look at CodeSnippetDirective or SyntaxErrorDirective works to get an idea! Also, best not to check in code that has testing dummy data in it.

dchen97:
OK I have made the modal it's own directive now under `ModalDirective.js`. Hopefully it should still work as expected!

dchen97:
I've written an update above but just in case, I've merged in the changes to master that added in the Privacy notice and made that a modal, so you can use that button to test the directive now.

seanlip:
@theresa-lee is this supposed to go in?

seanlip:
@theresa-lee -- friendly ping

theresa-lee:
Closed as #406 is the updated version of this.

PR #357, opened 2017-06-23
Add Print Feedback Directive

dchen97:
* Created the `PrintFeedbackDirective` to display the log output toggle and related warning whenever a user submits code that has a `print` statement in it.
* Added `isPrintFeedback` and `createPrintFeedbackParagraph` to `FeedbackParagraph`
* Added related lines to Learner View to allow for display of Print feedback paragraphs before all other feedback paragraphs for a given submission

Example of what feedback would look like with the log output expanded: 
image

Example of feedback with log output not expanded:
image


dchen97:
We decided not to use this UI for print feedback. Instead, reference #385

PR #358, opened 2017-06-28
Removing global imports from student code is not necessary

theresa-lee:
Deleting the function that removes the global imports.


seanlip:
Any idea what's up with the CLA?

theresa-lee:
Ah, I submitted from my laptop and I don't think I set it up correctly. Let me fix that!

PR #356, opened 2017-06-22
Improve Run-time Error Feedback

dchen97:
* Added series of user-friendly feedback responses to common run-time errors in `FeedbackGeneratorService`
* Added relevant tests for each runtime error response
* Changed affected test in `SolutionHandlerService`

PR #360, opened 2017-07-07
Wrong Language Syntax Detection

dchen97:
* Created new Python Prereq Check Failure types (and subsequently, new TIE constants) to accommodate new syntax error detection that has to do with using common syntax from other languages (i.e. Java, C, C++)
* Added `getWrongLanguageType` method to return the appropriate Prereq Check Failure Type if wrong language syntax are detected
* Add new responses to new Python Prerequisite Failures in `FeedbackGeneratorService`
* Added appropriate unit tests


client/services/code_evaluators/PythonPrereqCheckService.js
```diff
@@ -213,11 +236,18 @@ tie.factory('PythonPrereqCheckService', [
             PREREQ_CHECK_TYPE_BAD_IMPORT, unsupportedImports, null);
         }
 
+        var wrongLangName = detectAndGetWrongLanguageType(code);
```
theresa-lee: Can we rename this to wrongLangError or wrongLangErrorName perhaps? I feel like wrongLangName sounds like the name of the language.

dchen97: Done!

client/services/code_evaluators/PythonPrereqCheckService.js
```diff
@@ -20,16 +20,39 @@
...
+     * @returns {string | null} If there is wrong language syntax detected,
+     *    will return the name of the error as referenced in
+     *    WRONG_LANGUAGE_ERRORS.
+     */
+    var detectAndGetWrongLanguageType = function(code) {
+      var errorName = null;
+      if (code !== '') {
+        WRONG_LANGUAGE_ERRORS.python.forEach(function(error) {
+          if (code.search(new RegExp(error.regExString)) !== -1) {
+            errorName = error.errorName;
```
dchen97: Yeah, that's a good idea. Done!

client/app.js
```diff
@@ -146,6 +146,196 @@ tie.constant('PREREQ_CHECK_TYPE_BAD_IMPORT', 'badImport');
...
+    {
+      // Used a switch statement
+      errorName: 'switch',
+      regExString: '\\bswitch\\b\\s*\\((\\w|\\s)*\\)\\s*[{|:]?\\s*((\\bcase' +
+        '\\b)|(\\bdefault\\b))',
+      feedbackParagraphs: [
+        {
+          _type: 'text',
+          _content: [
+            "Sad to say but Python doesn't support switch statements. We ",
```
dchen97: Done!

client/domain/PrereqCheckFailureObjectFactory.js
```diff
@@ -197,10 +241,14 @@ tie.factory('PrereqCheckFailureObjectFactory', [
...
      *    headers the user is given before they begin adding their own code.
+     * @param {string} wrongLangName Name of the error that corresponds to the
+     *    error information in WRONG_LANGUAGE_ERRORS
      * @returns {PrereqCheckFailure}
      */
-    PrereqCheckFailure.create = function(type, badImports, starterCode) {
-      return new PrereqCheckFailure(type, badImports, starterCode);
+    PrereqCheckFailure.create = function(
+        type, badImports, starterCode, wrongLangName) {
+      var name = (typeof wrongLangName === 'undefined') ? null : wrongLangName;
```
seanlip: A common "fallback to default value" syntax in JS is: `var blah = wrongLangName || null;`

client/services/FeedbackGeneratorService.js
```diff
@@ -415,6 +415,20 @@ tie.factory('FeedbackGeneratorService', [
             'Please keep your code within the existing predefined functions',
             '-- we cannot process code in the global scope.'
           ].join(' '));
+        } else if (prereqCheckFailure.hasWrongLanguage()) {
+          WRONG_LANGUAGE_ERRORS.python.forEach(function(error) {
+            if (error.errorName === prereqCheckFailure.getWrongLangKey()) {
+              error.feedbackParagraphs.forEach(function(json) {
+                if (json._type === 'text') {
```
seanlip: Shouldn't be referencing private stuff here. Declare methods on the object if you need to.

Also, "json" is not very descriptive as a variable name. Maybe `paragraph` (if that's what this is)?

client/services/code_evaluators/PythonPrereqCheckService.js
```diff
@@ -20,16 +20,40 @@
...
+     * @param {string}
+     * @returns {string | null} If there is wrong language syntax detected,
+     *    will return the name of the error as referenced in
+     *    WRONG_LANGUAGE_ERRORS.
+     */
+    var detectAndGetWrongLanguageType = function(code) {
+      var errorName = null;
+      if (code !== '') {
+        WRONG_LANGUAGE_ERRORS.python.forEach(function(error) {
+          if (code.search(new RegExp(error.regExString)) !== -1 &&
```
seanlip: Consider doing: if (!errorName && code.search(...) !== -1). That way, once errorName is defined, the condition check shortcuts and doesn't do the more expensive search.

client/domain/PrereqCheckFailureObjectFactory.js
seanlip: I'm also seeing wrongLangErrorName, wrongLangName, and name. Let's be consistent and just use one of these throughout if they're all referring to the same concept. In particular "wrongLangName" conveys something quite different from "wrongLangErrorName" -- I thought it was referring to stuff like e.g. "c++".

client/app.js
```diff
@@ -148,6 +148,196 @@ tie.constant('PREREQ_CHECK_TYPE_BAD_IMPORT', 'badImport');
...
 tie.constant('PREREQ_CHECK_TYPE_GLOBAL_CODE', 'globalCode');
+tie.constant('PREREQ_CHECK_TYPE_WRONG_LANG', 'wrongLang');
+
+/**
+ * Dictionary of wrong language detection errors and their related information
+ *
+ * @type {{}}
+ */
+tie.constant('WRONG_LANGUAGE_ERRORS', {
+  python: [
```
seanlip: Consider compactifying a bit, similar to what we did [here](https://github.com/google/tie/pull/356/files#diff-b2454bae671216e787350ab0d8bc2380R158).

client/domain/PrereqCheckFailureObjectFactory.js
dchen97: Done!

dchen97: Done!

client/services/FeedbackGeneratorService.js
dchen97: I didn't really think about it much when I was writing this, but would it just be better to write everything as public properties? these aren't supposed to be fleshed out objects per se but just a simplified representation of a FeedbackParagraph object.

client/services/code_evaluators/PythonPrereqCheckService.js
dchen97: Done!

client/app.js
dchen97: Done!

client/services/FeedbackGeneratorService.js
seanlip: We should really be working with actual domain objects if you have anything more complicated than, say, a string. It's a bit hard to maintain plain js objects because they are so mutable (eg keys can be added to them at any time, etc)

So, when you say "simplified representation of a FeedbackParagraph object" what exactly do you mean? IE what exactly is "json"?

seanlip: Oh wait, I think I get it now -- this is actually reading from a constant? If so, I think referring to .type is fine, sorry.

client/app.js
```diff
@@ -148,6 +148,183 @@ tie.constant('PREREQ_CHECK_TYPE_BAD_IMPORT', 'badImport');
...
  */
 tie.constant('PREREQ_CHECK_TYPE_GLOBAL_CODE', 'globalCode');
+tie.constant('PREREQ_CHECK_TYPE_WRONG_LANG', 'wrongLang');
+
+/**
+ * Dictionary of wrong language detection errors and their related information
+ *
+ * @type {{}}
+ */
+tie.constant('WRONG_LANGUAGE_ERRORS', {
```
seanlip: Add Karma test that checks that the internal structure of this is valid. It will be useful when folks add new categories of errors.

client/services/FeedbackGeneratorService.js
```diff
@@ -415,6 +415,20 @@ tie.factory('FeedbackGeneratorService', [
             'Please keep your code within the existing predefined functions',
             '-- we cannot process code in the global scope.'
           ].join(' '));
+        } else if (prereqCheckFailure.hasWrongLanguage()) {
+          WRONG_LANGUAGE_ERRORS.python.forEach(function(error) {
+            if (error.errorName === prereqCheckFailure.getWrongLangKey()) {
+              error.feedbackParagraphs.forEach(function(paragraph) {
+                if (paragraph.type === 'text') {
```
seanlip: Can we declare constants somewhere like PARAGRAPH_TYPE_TEXT and PARAGRAPH_TYPE_CODE? The validation I mentioned earlier should check that the value in the relevant location is one of the allowed paragraph types.

client/app.js
dchen97: Which file would I put the test under? I'm not sure where it would be most appropriate since it doesn't directly relate specifically to any one object.

seanlip: I'd suggest creating a new appSpec.js file.

(Also, if you have time, it would be nice to do something similar for the other constant.)

dchen97: Done!

client/services/FeedbackGeneratorService.js

PR #363, opened 2017-07-10
Fix #36: Prevent Invalid System/AuxiliaryCode Calls

### LINKED ISSUE #36: Prevent users from making calls to system-level classes.
Students should not be able to access AuxiliaryCode or System classes.

Repro steps (at least, the easiest ones):

Navigate to Problem 2
Paste in the following code:
`def isBalanced(s):
    return AuxiliaryCode.countNumberOfParentheses(s)
`
Run code. You should get an error message (rather than a SyntaxError or something) since it called one of the AuxiliaryCode methods. We should block students from calling any of those classes, potentially via throwing a SyntaxError or a ForbiddenNamespaceError of our own making. Something to tell them yo you can't run our code.

dchen97:
Fix to Issue #36 
* Added 2 PrereqCheckFailure types to account for invalid calls to the `System` and `AuxiliaryCode` classes
* Added new feedback for these new PrereqCheckFailure types in `FeedbackGenerator`
* Added relevant tests

### DIFF
--- a/client/app.js
+++ b/client/app.js
@@ -122,6 +122,20 @@ tie.constant('SYSTEM_CODE', {
   ].join('\n')
 });
 
+/**
+ * Provides class names for system and auxiliary code according to the language
+ * that is being used.
+ *
+ * @type {{}}
+ * @constant
+ */
+tie.constant('SYSTEM_AUXILIARY_CLASSES', {
+  python: {
+    systemClass: 'System',
+    auxiliaryClass: 'AuxiliaryCode'
+  }
+});
+
 // Pre-requisite check error types
 /**
  * Pre-requisite check error type for when starter code is missing or altered.
--- a/client/app.js
+++ b/client/app.js
@@ -124,20 +124,6 @@ tie.constant('SYSTEM_CODE', {
   ].join('\n')
 });
 
-/**
- * Provides class names for system and auxiliary code according to the language
- * that is being used.
- *
- * @type {{}}
- * @constant
- */
-tie.constant('SYSTEM_AUXILIARY_CLASSES', {
-  python: {
-    systemClass: 'System',
-    auxiliaryClass: 'AuxiliaryCode'
-  }
-});
-
 // Pre-requisite check error types
 /**
  * Pre-requisite check error type for when starter code is missing or altered.
--- a/client/app.js
+++ b/client/app.js
@@ -148,6 +148,25 @@ tie.constant('PREREQ_CHECK_TYPE_BAD_IMPORT', 'badImport');
 tie.constant('PREREQ_CHECK_TYPE_GLOBAL_CODE', 'globalCode');
 
 /**
+ * Pre-requisite check error type to see if the user tried to use any methods
+ * from System in their code submission.
+ *
+ * @type {string}
+ * @constant
+ */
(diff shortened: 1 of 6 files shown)


dchen97:
Had an idea about making this more sustainable. Will work on it more so it's not ready for review yet.

dchen97:
OK I've made some adjustments but I'm not sure they're actually all that effective or helpful in terms of scalability for how things are right now. I'm open to suggestions! Otherwise, this should be ready for review now.

client/app.js
```diff
@@ -122,6 +122,20 @@ tie.constant('SYSTEM_CODE', {
...
 });
 
+/**
+ * Provides class names for system and auxiliary code according to the language
+ * that is being used.
+ *
+ * @type {{}}
+ * @constant
+ */
+tie.constant('SYSTEM_AUXILIARY_CLASSES', {
```
seanlip: I thought we already had constants for these (cf one of Elena's recent PRs). I think it's also fine for them to be consistent across all languages?

client/services/FeedbackGeneratorService.js
```diff
@@ -413,6 +413,26 @@ tie.factory('FeedbackGeneratorService', [
...
             '-- we cannot process code in the global scope.'
           ].join(' '));
+        } else if (prereqCheckFailure.hasInvalidAuxiliaryCodeCall()) {
+          feedback.appendTextParagraph([
+            'Looks like your code had a runtime error. Here is the error ',
+            'message: '
+          ].join(''));
+          feedback.appendCodeParagraph([
+            'ForbiddenNamespaceError: It looks like you\'re trying to call ',
+            'the AuxiliaryCode class or its methods, which is forbidden. ',
```
seanlip: Error message should make use of the constant in app.js.

client/services/FeedbackGeneratorService.js
```diff
@@ -413,6 +413,26 @@ tie.factory('FeedbackGeneratorService', [
...
+            'Please resubmit without using this class.'
+          ].join(''));
+        } else if (prereqCheckFailure.hasInvalidSystemCall()) {
+          feedback.appendTextParagraph([
+            'Looks like your code had a runtime error. Here is the error ',
+            'message: '
+          ].join(''));
+          feedback.appendCodeParagraph([
+            'ForbiddenNamespaceError: It looks you\'re using the System class ',
+            'or its methods, which is forbidden. Please resubmit without ',
```
seanlip: Ditto.

client/services/code_evaluators/PythonPrereqCheckService.js
```diff
@@ -164,6 +192,14 @@ tie.factory('PythonPrereqCheckService', [
             PREREQ_CHECK_TYPE_GLOBAL_CODE, null, starterCode);
         }
 
+        if (checkInvalidSystemClassCalls(code) === SYSTEM_CALL) {
```
seanlip: Would expect something more like: `if (hasInvalidSystemClassCalls(code))`, and `if (hasInvalidAuxiliaryClassCalls(code))`. Then you don't need to pass those strings around.

client/app.js
dchen97: Done!

client/services/FeedbackGeneratorService.js

client/services/FeedbackGeneratorService.js

PR #364, opened 2017-07-10
Task tests

eledavi:
Created spec file to test TaskObjectFactory 
Removed CLASS_NAME_SYSTEM_CODE and its use in the if-else statement on lines 185-188 bc it was a constant (defined in client/app.js) with the same string value as CLASS_NAME_AUXILIARY_CODE. Based on my tests there's never a case where the second half of the if-else statement runs. I'm not sure if the value of CLASS_NAME_SYSTEM_CODE is buggy or if CLASS_NAME_SYSTEM_CODE isn't necessary in this file. PTAL.

### DIFF
--- a/client/domain/TaskObjectFactorySpec.js
+++ b/client/domain/TaskObjectFactorySpec.js
@@ -1,0 +1,65 @@
+// Copyright 2017 The TIE Authors. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS-IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+/**
+ * @fileoverview Unit tests for TaskObject domain objects.
+ */
+
+describe('TaskObjectFactory', function() {
+  var QuestionObjectFactory;
+  var question;
+
+  var TITLE = 'title';
+  var STARTER_CODE = 'starterCode';
+  var AUXILIARY_CODE = 'auxiliaryCode';
+  var OUTPUT_FUNCTION_NAME_1 =
+  'AuxiliaryCode_AuxiliaryCodeOutputFunctionNameWithoutClass';
+  var OUTPUT_FUNCTION_NAME_2 = 'System_SystemOutputFunctionNameWithoutClass';
+
+  beforeEach(module('tie'));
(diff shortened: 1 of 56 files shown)


seanlip:
Great catch! The value in app.js is an error. It should be `tie.constant('CLASS_NAME_SYSTEM_CODE', 'System');` and the corresponding code is defined [here](https://github.com/google/tie/blob/master/client/app.js#L101). So the other stuff in TaskObjectFactory.js should be left as is. In order to test it, you could probably make a dummy question data object for tests that uses System.extendString().

eledavi:
@seanlip Done! PTAL

client/domain/TaskObjectFactorySpec.js
```diff
@@ -0,0 +1,65 @@
...
+
+/**
+ * @fileoverview Unit tests for TaskObject domain objects.
+ */
+
+describe('TaskObjectFactory', function() {
+  var QuestionObjectFactory;
+  var question;
+
+  var TITLE = 'title';
```
seanlip: You don't need to pull these out as constants. Just include them within the definition of `question` below.

client/domain/TaskObjectFactorySpec.js
```diff
@@ -0,0 +1,65 @@
...
+  var STARTER_CODE = 'starterCode';
+  var AUXILIARY_CODE = 'auxiliaryCode';
+  var OUTPUT_FUNCTION_NAME_1 =
```
seanlip: Actually, the expected value is more like "AuxiliaryCode.splitString" or similar. The idea is that there are System and AuxiliaryCode classes, each of which have some custom methods. The question is providing a reference to this method so that it can be invoked on the test cases during execution.

client/domain/TaskObjectFactorySpec.js
```diff
@@ -0,0 +1,65 @@
...
+  'AuxiliaryCode_AuxiliaryCodeOutputFunctionNameWithoutClass';
+  var OUTPUT_FUNCTION_NAME_2 = 'System_SystemOutputFunctionNameWithoutClass';
```
seanlip: Ditto.

client/domain/TaskObjectFactorySpec.js
eledavi: Done

seanlip:
Also FYI, there are merge conflicts.

client/domain/TaskObjectFactorySpec.js
eledavi: Done!

client/domain/TaskObjectFactorySpec.js
eledavi: Done!

seanlip:
@eledavi - could you please resolve the merge conflicts? Then this can be merged.

eledavi:
@seanlip Done!

seanlip:
Thanks @eledavi! We can merge this after the Travis checks pass.

PR #365, opened 2017-07-10
Button for resetting feedback

theresa-lee:
Added a button to reset the feedback window (it needs to be updated when the localstorage stuff is implemented).

### DIFF
--- a/client/components/LearnerViewDirective.js
+++ b/client/components/LearnerViewDirective.js
@@ -734,6 +734,7 @@ tie.directive('learnerView', [function() {
          *     question
          */
         var loadQuestion = function(questionId, introParagraphs) {
+          clearFeedback();
           question = QuestionDataService.getQuestion(questionId);
           tasks = question.getTasks();
           currentTaskIndex = 0;
--- a/client/components/LearnerViewDirective.js
+++ b/client/components/LearnerViewDirective.js
@@ -104,8 +104,8 @@ tie.directive('learnerView', [function() {
                 <option style="display: none" value="">Theme</option>
               </select>
               <button class="tie-feedback-reset tie-button" name="feedback-reset"
-                      ng-click="resetFeedback()">
-                  Reset Feedback
+                  ng-click="resetFeedback()">
+                Reset Feedback
               </button>
             </div>
             <div class="tie-coding-ui">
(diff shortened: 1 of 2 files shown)


seanlip:
Hm, I'm slightly confused by this. I thought we were planning to have one button that resets (only) code, and nothing else? The issue is that the current button resets both.

theresa-lee:
So I believe the decision from the feedback from users is that we would have two reset buttons, one for feedback and one for code. This PR adds in the one for feedback, and the current reset code button does indeed reset both code and feedeback because we don't have storing the feedback mechanisms in place.

So this isn't a PR to fix the current reset code button (which is admittedly buggy) but that would be something that would be fixed with the localStorage stuff being added.

Should we wait until after that has been submitted and integrated before adding in this button?

seanlip:
Ah -- actually, I had thought the decision was to have one button that resets only the code, and to hold off on adding the "reset feedback" button until users find that they need it. I can't find the notes on this, though, so just to double check -- @rabidbit, do you remember which one it is?

rabidbit:
Separating clearing code and clearing feedback functionality - definite yes.
Ability to clear code - definite yes.
Ability to clear feedback - slight yes; so, we can go either way and rely on future feedback to make adjustments. I'd say do what's best from two angles (a) easiest direction to revert in (b) least confusing from a UX standpoint (since enabling feedback would result in another button.
Ability to specify if feedback should be cleared before moving on to the next part of the question / prompt - definite yes.

seanlip:
Hm, I'm slightly confused about the last two points:

> Ability to clear feedback - slight yes; so, we can go either way and rely on future feedback to make adjustments. I'd say do what's best from two angles (a) easiest direction to revert in (b) least confusing from a UX standpoint (since enabling feedback would result in another button).

This last part suggests having one button, I think (at least in the interim).

> Ability to specify if feedback should be cleared before moving on to the next part of the question / prompt - definite yes.

But this suggests having two buttons? Or do you mean: definite yes to actually clearing the feedback when moving between tasks/questions?

rabidbit:
Whether there is a button to clear the feedback before moving on to the next task and how such a button is implemented is an open UI/UX question for the person implementing it (is it a checkmark in the "proceed to next question / question library" overlay? Is it a button within the overlay?). This is something users asked for; so, I think it should be done.

I'm making a distinction between this and the feature where the user can clear the feedback history at any time for any reason (which, if implemented, I would assume almost requires a separate button to clear). I think this is a separate feature where there was a slight preference to having this feature. Since it was slight in a small population set, I think it's up to us to decide (hence some of my thoughts on how to decide if we should do this or not), and we can adjust later as needed.

theresa-lee:
Closing this because we decided that we would not have a reset feedback button unless user studies result in users asking for them.

Issue #335, opened 2017-06-15
Question Numbers go off page


dchen97:
**Steps to Reproduction:**
* Open `app.html` in browser
* Select the `all` question set.
* Notice that the circles with question numbers now extend beyond the viewport to the right.

I'm not sure what the ideal fix is, but just thought we should keep this issue for a later fix.

seanlip:
No longer relevant since we no longer have question numbers. Closing.

PR #373, opened 2017-07-14
Fix 335: Question Numbers Going Off Page

dchen97:
Fix to Issue #335 
* Added styling so that question numbers will all be equally spaced on page to allow them all to be presented without going off the page

Example:
image


rabidbit:
Thanks for doing this. Looks cool. We may not use the stepper for the next version (Beta) given our goals for question quantity, but we may re-introduce it again later as we add more content.

seanlip:
@theresa-lee is this supposed to go in?

theresa-lee:
Closing because updated UI will not make use of the stepper.

PR #376, opened 2017-07-18
Fix #355: Resolve Code Window Flashing

### LINKED ISSUE #355: Code window flashes white when navigating between questions
Expected behavior: code window does not flash white for a split second when switching between questions.

Observed behavior: code window flashes white for a split second when switching between questions.

Steps to reproduce: navigate between questions and note code window flashes white for a split second.

dchen97:
Fix to Issue #355 
* Added a "placeholder" codemirror window that would look exactly like the original code window when it is being removed and replaced in the DOM. Resolves the code window flashing problem.

### DIFF
--- a/client/components/LearnerViewDirective.js
+++ b/client/components/LearnerViewDirective.js
@@ -121,7 +121,7 @@ tie.directive('learnerView', [function() {
                                    class="protractor-test-code-input">
                     </ui-codemirror>
                   </div>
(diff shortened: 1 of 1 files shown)


seanlip:
@theresa-lee is this PR supposed to go in?

theresa-lee:
Closing because 1) planning not to be able to switch around questions like this for the updated UI 2) Unless users complain about it, it isn't something I actually noticed that was a problem.

PR #378, opened 2017-07-19
Fix AttributeError Regex Bug

dchen97:
* Adjusted AttributeError regex for when error names an attribute that includes non-word characters

## To test:
* Before importing changes, try using a `length` method or property for a string
* Note that an error is thrown because `found` variable is null
* Import changes and now no error should be thrown


theresa-lee:
Sorry this is so late. If you can't resolve merge conflicts (because of the file re-organization), then maybe best to create a new PR (but link to this one).

dchen97:
Done with fixing merge conflicts!

PR #379, opened 2017-07-19
Fix #370: Reinforcement Input Formatting

### LINKED ISSUE #370: Show reinforcement inputs in code font.
In the following screenshot, the last line reads `Fails on "lots of spaces"`. However, the input given here was actually meant to be "lots &nbsp;&nbsp;&nbsp;of&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;spaces".

screenshot from 2017-07-13 20 03 41

The input should be shown in code font so that the space characters are visible.

dchen97:
Fix to Issue #370 
* Added helper function and adjusted test case input formatting to ensure spaces don't get collapsed when rendering reinforcement test case input

Example of old Reinforcement bullet rendering:
image

Example of Reinforcement with proposed changes:
image

### DIFF
--- a/assets/questions/findMostCommonCharacter.js
+++ b/assets/questions/findMostCommonCharacter.js
@@ -92,7 +92,7 @@ class AuxiliaryCode(object):
       allowedOutputs: ['s'],
       tag: 'strings with many spaces'
     }, {
-      input: 'a b c d e f f',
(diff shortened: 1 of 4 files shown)


dchen97:
OK I've added an E2E test that should now check that, when rendered, the input will appear as expected.

PR #380, opened 2017-07-24
Syntax error underlining

eledavi:
* Added function to highlight/underline syntax errors
* Updated syntax error highlighting feature to highlight/un-highlight syntax errors as the user's code changes
* Updated line number that is highlighted


client/components/LearnerViewDirective.js
```diff
@@ -531,6 +531,10 @@ tie.directive('learnerView', [function() {
         .tie-step-unlocked {
           background-color: rgb(0, 128, 0);
         }
+        .tie-syntax-error-line {
+          background: #FBC2C4 !important;
```
seanlip: !important is a bit of an antipattern. Is it possible to do without it?

client/components/LearnerViewDirective.js
```diff
@@ -541,14 +545,14 @@ tie.directive('learnerView', [function() {
...
+      'QuestionDataService', 'LANGUAGE_PYTHON', 'ErrorTracebackObjectFactory',
+      'FeedbackObjectFactory', 'ReinforcementObjectFactory',
+      'CodeStorageService', 'SECONDS_TO_MILLISECONDS',
+      'DEFAULT_AUTOSAVE_SECONDS', 'DISPLAY_AUTOSAVE_TEXT_SECONDS', 'SERVER_URL',
       function(
           $scope, $interval, $timeout, SolutionHandlerService,
-          QuestionDataService, LANGUAGE_PYTHON, FeedbackObjectFactory,
-          ReinforcementObjectFactory, CodeStorageService,
+          QuestionDataService, LANGUAGE_PYTHON, ErrorTracebackObjectFactory,
+          FeedbackObjectFactory, ReinforcementObjectFactory, CodeStorageService,
```
seanlip: Nit: try to have the line breaks here match the above, so that it's easier to compare the two.

client/components/LearnerViewDirective.js
```diff
@@ -733,6 +737,30 @@ tie.directive('learnerView', [function() {
...
 
         /**
+          * Highlights syntax errors in the coding UI
+          *
+          * @param {number} lineNumber
+          */
+        var highlightLine = function(lineNumber) {
+          var actualLineNumber = lineNumber - 1;
+          var codeLines = document.querySelectorAll('.CodeMirror-line');
+          codeLines[actualLineNumber].className += ' tie-syntax-error-line';
```
seanlip: Just to check, why do you need to apply the new class to two places? Would just one not work?

client/components/LearnerViewDirective.js
```diff
@@ -773,7 +801,15 @@ tie.directive('learnerView', [function() {
             for (var i = 0; i < feedbackParagraphs.length; i++) {
               if (feedbackParagraphs[i].isSyntaxErrorParagraph()) {
                 hasSyntaxError = true;
+                clearHighlight();
```
seanlip: There's a bit more refactoring that's needed here, I think. Syntax errors are no longer feedback paragraphs, so they should really be a separate construct altogether. Maybe we should chat about this...

client/components/LearnerViewDirective.js
eledavi: Done!

client/components/LearnerViewDirective.js
eledavi: Done!

client/components/LearnerViewDirective.js
```diff
@@ -733,6 +739,27 @@ tie.directive('learnerView', [function() {
...
+          codeLines[actualLineNumber].className += ' tie-syntax-error-line';
+        };
+
+        /**
+          * Clears highlight from syntax errors in the coding UI
+          */
+        var clearHighlight = function() {
+          var codeLines = document.querySelectorAll('.tie-syntax-error-line');
+          for (var i = 0; i < codeLines.length; i++) {
+            codeLines[i].className = 'CodeMirror-line';
```
seanlip: Shouldn't you just remove the class you added (in case something else adds stuff to that line as well in the future)?

client/components/LearnerViewDirective.js
```diff
@@ -733,6 +739,27 @@ tie.directive('learnerView', [function() {
...
 
         /**
+          * Highlights syntax errors in the coding UI
+          *
+          * @param {number} lineNumber
+          */
+        var highlightLine = function(lineNumber) {
+          var actualLineNumber = lineNumber - 1;
+          var codeLines = document.querySelectorAll('.CodeMirror-line');
+          codeLines[actualLineNumber].className += ' tie-syntax-error-line';
```
seanlip: Pull class name out into a named variable if you're going to use it in multiple places (see comment in clearHighlight() below).

client/components/LearnerViewDirective.js
seanlip: Retracting this comment after discussion with @eledavi. There's still a feedback message.

client/components/LearnerViewDirective.js
```diff
@@ -773,7 +800,15 @@ tie.directive('learnerView', [function() {
...
                 hasSyntaxError = true;
+                clearHighlight();
+                var errorContentArray = feedbackParagraphs[i].getContent()
+                  .split(' ');
+                var syntaxErrorLineNumber = errorContentArray[errorContentArray
+                  .length - 1];
+                highlightLine(parseInt(syntaxErrorLineNumber, 10));
                 break;
+              } else {
+                clearHighlight();
```
seanlip: Instead of using clearHighlight() in both the if and else case, why not just clearHighlight() once before the for loop?

eledavi: Done!

client/components/LearnerViewDirective.js
eledavi: Done!

eledavi:
Here are screenshots of what the syntax error highlighting looks like right now. LMK if I need to make any adjustments. 
screenshot from 2017-07-26 17 42 45
screenshot from 2017-07-26 17 43 21

client/components/LearnerViewDirective.js
```diff
@@ -733,6 +739,33 @@ tie.directive('learnerView', [function() {
...
 
         /**
+          * Highlights syntax errors in the coding UI
+          *
+          * @param {number} lineNumber
+          */
+        var highlightLine = function(lineNumber) {
+          var actualLineNumber = lineNumber - 1;
+          var codeLines = document.querySelectorAll('.CodeMirror-line');
+          if ($scope.isInDarkMode === true) {
```
seanlip: You can omit the "=== true".

client/components/LearnerViewDirective.js
```diff
@@ -733,6 +739,33 @@ tie.directive('learnerView', [function() {
...
+            codeLines[actualLineNumber].className +=
+              ' tie-syntax-error-line night-mode';
```
seanlip: Hm. So if I change the mode midway, what happens?

I think this function shouldn't reference night-mode at all. Instead, there should be a CSS rule whose selector is something like `.parent-element.night-mode .tie-syntax-error-line`.

client/components/LearnerViewDirective.js
```diff
@@ -733,6 +739,33 @@ tie.directive('learnerView', [function() {
...
+          }
+        };
+
+        /**
+          * Clears highlight from syntax errors in the coding UI
+          */
+        var clearHighlight = function() {
+          var codeLines = document.querySelectorAll('.tie-syntax-error-line');
+          for (var i = 0; i < codeLines.length; i++) {
+            codeLines[i].className = codeLines[i].className.slice(0,
```
seanlip: This seems too fiddly. Isn't there something like classList.remove()?

client/components/LearnerViewDirective.js
```diff
@@ -733,6 +739,33 @@ tie.directive('learnerView', [function() {
...
+          * @param {number} lineNumber
+          */
+        var highlightLine = function(lineNumber) {
+          var actualLineNumber = lineNumber - 1;
+          var codeLines = document.querySelectorAll('.CodeMirror-line');
+          if ($scope.isInDarkMode === true) {
+            codeLines[actualLineNumber].className +=
+              ' tie-syntax-error-line night-mode';
+          } else {
+            codeLines[actualLineNumber].className += ' tie-syntax-error-line';
```
seanlip: I think, instead of manipulating strings, doing something like classList.add() would be better -- since that's what you're trying to do conceptually. See https://developer.mozilla.org/en-US/docs/Web/API/Element/classList

client/components/LearnerViewDirective.js
```diff
@@ -771,8 +804,14 @@ tie.directive('learnerView', [function() {
             var feedbackParagraphs = feedback.getParagraphs();
             var hasSyntaxError = false;
             for (var i = 0; i < feedbackParagraphs.length; i++) {
+              clearHighlight();
               if (feedbackParagraphs[i].isSyntaxErrorParagraph()) {
                 hasSyntaxError = true;
+                var errorContentArray = feedbackParagraphs[i].getContent()
+                  .split(' ');
+                var syntaxErrorLineNumber = errorContentArray[errorContentArray
```
seanlip: Break line after '['. Odd to split up errorContentArray and .length.

seanlip: Also I think this is better done by calling something like feedbackParagraphs[i].getErrorLineNumber() -- it's something that the paragraph object itself should take care of.

seanlip:
The UI looks nice -- thanks @eledavi! I have a few more code comments.

client/components/LearnerViewDirective.js
eledavi: Done!

client/components/LearnerViewDirective.js
eledavi: Done!

client/components/LearnerViewDirective.js
eledavi: Done!

client/components/LearnerViewDirective.js
eledavi: Done!

client/components/LearnerViewDirective.js
```diff
@@ -733,6 +739,32 @@ tie.directive('learnerView', [function() {
...
+          * Highlights syntax errors in the coding UI
+          *
+          * @param {number} lineNumber
+          */
+        var highlightLine = function(lineNumber) {
+          var actualLineNumber = lineNumber - 1;
+          var codeLines = document.querySelectorAll('.CodeMirror-line');
+          if ($scope.isInDarkMode) {
+            codeLines[actualLineNumber].classList.add('tie-syntax-error-line',
+              'night-mode');
```
seanlip: Ah, no, this isn't quite what I meant.

Try having your CSS rule be something like `.tie-wrapper.night-mode .CodeMirror-line.tie-syntax-error-line`. Does that work? We currently have too many "night-mode"s scattered all over the place and I'm leery of adding another one indirectly via JS.

seanlip: (Oh, sorry, my bad. I just realized you hadn't actually addressed that comment yet.)

client/components/LearnerViewDirective.js
eledavi: Unfortunately the only method like that for feedback paragraphs is the getContent() method I used. Would you like me to add a method to get just the error line number?

seanlip: It is fine to add new methods, yes; make sure you test them too. The method could start by checking that the paragraph is of the appropriate type. and throwing an error if not.

client/domain/FeedbackParagraphObjectFactory.js
```diff
@@ -109,6 +109,16 @@ tie.factory('FeedbackParagraphObjectFactory', [
       return this._content;
     };
 
+    /**
+     * A getter for the line number where the error occurs.
+     *
+     * @returns {number}
+     */
+    FeedbackParagraph.prototype.getErrorLineNumber = function() {
+      var errorContentArray = this.getContent().split(' ');
```
seanlip: Should check that the paragraph is of the correct type first, and if not, raise error

Also, add dedicated tests for this method.

client/components/LearnerViewDirective.js
```diff
@@ -733,6 +739,27 @@ tie.directive('learnerView', [function() {
...
+        var highlightLine = function(lineNumber) {
+          var actualLineNumber = lineNumber - 1;
+          var codeLines = document.querySelectorAll('.CodeMirror-line');
+          codeLines[actualLineNumber].classList.add('tie-syntax-error-line');
+        };
+
+        /**
+          * Clears highlight from syntax errors in the coding UI
+          */
+        var clearHighlight = function() {
```
seanlip: Just realized that clearAllHighlights() is probably a better name, since that's what this is doing (as opposed to clearing a single highlight).

client/components/LearnerViewDirective.js
```diff
@@ -733,6 +739,27 @@ tie.directive('learnerView', [function() {
...
 
         /**
+          * Highlights syntax errors in the coding UI
+          *
+          * @param {number} lineNumber
+          */
+        var highlightLine = function(lineNumber) {
+          var actualLineNumber = lineNumber - 1;
+          var codeLines = document.querySelectorAll('.CodeMirror-line');
+          codeLines[actualLineNumber].classList.add('tie-syntax-error-line');
```
seanlip: Pull .tie-syntax-error-line out into a named constant since you use it multiple times

client/components/LearnerViewDirective.js
eledavi: Done!

client/components/LearnerViewDirective.js
eledavi: Done!

eledavi:
@theresa-lee Done! LMK what you think of this highlight color.
screenshot from 2017-07-31 15 19 56

theresa-lee:
@eledavi Looks good to me!

client/domain/FeedbackParagraphObjectFactory.js
eledavi: Done!

eledavi:
This PR is now irrelevant. All of these changes can be found on PR #384.

PR #382, opened 2017-07-26
Add require-jsdoc to linter

dchen97:
* Added new `require-jsdoc` rule to linter so that we don't regress in adding JSDoc comments

*Note*: There's another related rule called [valid-jsdoc](http://eslint.org/docs/rules/valid-jsdoc) that enforces style and consistency within JSDoc comments. I wasn't sure whether we wanted to be that strict, and there are also a lot of ways we can specify it, which I'd love to get team input on if we do decide to go with adding this rule.

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -148,6 +148,14 @@
         "quote-props": ["error", "as-needed"],
         "radix": "error",
         "require-await": "error",
+        "require-jsdoc": ["error", {
+          "require": {
+            "FunctionDeclaration": true,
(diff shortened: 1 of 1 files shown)


seanlip:
Thanks, @dchen97! I do lean towards being as strict as possible, since the more of these checks we can automate, the better. @theresa-lee @rabidbit @eyurko @eledavi any thoughts?

theresa-lee:
I'm okay with having just 'require-jsdoc' enabled for now.

PR #383, opened 2017-07-26
Language unfamiliarity

eledavi:
Add preliminary methods/stuff to FeedbackGeneratorService for language unfamiliarity feature


seanlip:
Missing docstrings?

I suggest that @dchen97 and @theresa-lee take first pass on this (to verify the spec and UI/UX). I'm happy to do a code pass once that's confirmed.

Also, removing the other two reviewers from the "attention list" (but @eyurko and @rabidbit feel free to chime in if you'd like to!)

client/services/FeedbackGeneratorServiceSpec.js
```diff
@@ -248,6 +248,17 @@ describe('FeedbackGeneratorService', function() {
...
+    it('should return the string for the text feedback', function() {
+      var feedbackString = FeedbackGeneratorService
+        ._getUnfamiliarLanguageFeedback();
+      expect(feedbackString).toEqual([
+        "Seems like you're having some trouble with this language. Why ",
+        "don't you take a look at the primer?"
+      ].join(''));
+    });
+  });
+
```
dchen97: We'll have to add some tests for `resetCounters` and also to double check that the correct feedback is appended when language unfamiliarity is detected. We can pair program that part together if you want.

client/services/FeedbackGeneratorService.js
```diff
@@ -30,6 +30,33 @@ tie.factory('FeedbackGeneratorService', [
...
+    var consecutiveSameErrorCounter = 0;
+    var consecutiveWrongLanguageCounter = 0;
+    var previousErrorString = '';
+
+    /**
+     *
+     */
+    var _resetCounters = function(currentCtr) {
+      var notResetCtr = currentCtr || '';
+      if (notResetCtr !== 'consecutiveSyntaxErrorCounter') {
```
dchen97: I get the feeling there is a better way to do this rather than comparing direct strings that can be more precise and foolproof. Perhaps we can use constants to represent these strings?

client/services/FeedbackGeneratorService.js
```diff
@@ -280,6 +307,16 @@ tie.factory('FeedbackGeneratorService', [
     };
 
     /**
+     *
+     */
+    var _getUnfamiliarLanguageFeedback = function() {
+      return [
+        "Seems like you're having some trouble with this language. Why ",
+        "don't you take a look at the primer?"
```
dchen97: Perhaps we can add a small note to look at the bottom of the screen or to look at the `New to Python` button (or whatever we decide to call the primer button) so the user knows where to go.

client/services/FeedbackGeneratorService.js
```diff
@@ -373,21 +410,46 @@ tie.factory('FeedbackGeneratorService', [
...
               tasks[tasks.length - 1], codeEvalResult));
         }
+        if (consecutiveSameErrorCounter === UNFAMILIARITY_THRESHOLD) {
+          feedback.appendTextParagraph(_getUnfamiliarLanguageFeedback());
+          consecutiveSameErrorCounter = 0;
+        }
+        previousErrorString = codeEvalResult.getErrorString();
         return feedback;
       },
       getSyntaxErrorFeedback: function(errorString) {
```
dchen97: This isn't a big deal but we should probably add a JSDoc comment for this function and in it we can explain some of the reasoning for incrementing and resetting the variables.

client/services/FeedbackGeneratorService.js
```diff
@@ -30,6 +30,33 @@ tie.factory('FeedbackGeneratorService', [
     RUNTIME_ERROR_FEEDBACK_MESSAGES, WRONG_LANGUAGE_ERRORS, LANGUAGE_PYTHON,
     CLASS_NAME_AUXILIARY_CODE, CLASS_NAME_SYSTEM_CODE, PARAGRAPH_TYPE_TEXT,
     PARAGRAPH_TYPE_CODE, PARAGRAPH_TYPE_SYNTAX_ERROR) {
+    /**
+     *
+     */
+    var UNFAMILIARITY_THRESHOLD = 5;
```
dchen97: Add JSDoc Comment

client/services/FeedbackGeneratorService.js
```diff
@@ -30,6 +30,33 @@ tie.factory('FeedbackGeneratorService', [
...
+
+    var consecutiveSyntaxErrorCounter = 0;
```
dchen97: Add JSDoc Comment

client/services/FeedbackGeneratorService.js
```diff
@@ -30,6 +30,33 @@ tie.factory('FeedbackGeneratorService', [
...
+    var consecutiveSameErrorCounter = 0;
```
dchen97: Add JSDoc Comment

client/services/FeedbackGeneratorService.js
```diff
@@ -30,6 +30,33 @@ tie.factory('FeedbackGeneratorService', [
...
+    var consecutiveWrongLanguageCounter = 0;
```
dchen97: Add JSDoc Comment

client/services/FeedbackGeneratorService.js
```diff
@@ -30,6 +30,33 @@ tie.factory('FeedbackGeneratorService', [
...
+    var previousErrorString = '';
```
dchen97: Add JSDoc Comment

client/services/FeedbackGeneratorService.js
```diff
@@ -30,6 +30,33 @@ tie.factory('FeedbackGeneratorService', [
...
+
+    /**
+     *
+     */
+    var _resetCounters = function(currentCtr) {
```
dchen97: Add JSDoc Comment

client/services/FeedbackGeneratorService.js
```diff
@@ -30,6 +30,33 @@ tie.factory('FeedbackGeneratorService', [
...
+      var notResetCtr = currentCtr || '';
```
dchen97: Perhaps a short comment explaining the purpose for checking if currentCtr is null?

client/services/FeedbackGeneratorService.js
```diff
@@ -280,6 +307,16 @@ tie.factory('FeedbackGeneratorService', [
     };
 
     /**
+     *
+     */
+    var _getUnfamiliarLanguageFeedback = function() {
```
dchen97: Add JSDoc Comment

client/services/FeedbackGeneratorService.js
```diff
@@ -373,21 +410,46 @@ tie.factory('FeedbackGeneratorService', [
...
+        consecutiveSyntaxErrorCounter++;
         var feedback = FeedbackObjectFactory.create(false);
         feedback.appendSyntaxErrorParagraph(errorString);
+        if (consecutiveSyntaxErrorCounter === UNFAMILIARITY_THRESHOLD) {
+          feedback.appendTextParagraph(_getUnfamiliarLanguageFeedback());
+          consecutiveSyntaxErrorCounter = 0;
+        }
         return feedback;
       },
       getPrereqFailureFeedback: function(prereqCheckFailure) {
```
dchen97: We should also probably add a JSDoc comment here as well.

client/services/FeedbackGeneratorService.js
eledavi: Done!

client/services/FeedbackGeneratorService.js
eledavi: Done!

client/services/FeedbackGeneratorService.js
eledavi: Done!

client/services/FeedbackGeneratorService.js
eledavi: Done!

client/services/FeedbackGeneratorService.js
eledavi: Done!

client/services/FeedbackGeneratorService.js
eledavi: Done!

client/services/FeedbackGeneratorService.js
eledavi: Done!

client/services/FeedbackGeneratorService.js
eledavi: Done! PTAL

client/services/FeedbackGeneratorService.js
eledavi: Done!

client/services/FeedbackGeneratorService.js
eledavi: Done! PTAL and let me know if we should rephrase the message

client/services/FeedbackGeneratorService.js
eledavi: Done!

client/services/FeedbackGeneratorService.js
eledavi: Done!

client/services/FeedbackGeneratorService.js
```diff
@@ -373,21 +410,46 @@ tie.factory('FeedbackGeneratorService', [
        * @returns {Feedback}
        */
       getFeedback: function(tasks, codeEvalResult, rawCodeLineIndexes) {
+        _resetCounters('consecutiveSameErrorCounter');
+
+        if (previousErrorString === codeEvalResult.getErrorString()) {
```
theresa-lee: Can previousErrorString be null here?

client/services/FeedbackGeneratorService.js
theresa-lee: I'm not sure I understand why this is called the notResetCtr?

client/services/FeedbackGeneratorServiceSpec.js
```diff
@@ -248,6 +248,18 @@ describe('FeedbackGeneratorService', function() {
     });
   });
 
+  describe('_getUnfamiliarLanguageFeedback', function() {
+    it('should return the string for the text feedback', function() {
+      var feedbackString = FeedbackGeneratorService
+        ._getUnfamiliarLanguageFeedback();
+      expect(feedbackString).toEqual([
+        "Seems like you're having some trouble with this language. Why ",
+        "don't you take a look at the New to Python button at the bottom ",
```
dchen97: Small nit with wording but rephrase to clarify looking at the "primer linked through the `New to Python` button" rather than just the button itself

dchen97: Sorry didn't realize I put this in the test, but this is more related to the actual method with the message above in `FeedbackGeneratorService`

client/services/FeedbackGeneratorService.js
```diff
@@ -373,21 +447,58 @@ tie.factory('FeedbackGeneratorService', [
        * @returns {Feedback}
        */
       getFeedback: function(tasks, codeEvalResult, rawCodeLineIndexes) {
+        // Reset all counters but consecutiveSameErrorCounter.
+        _resetCounters('consecutiveSameError');
+
+        // If the user recieves the same error message increment the same error
+        // counter.
+        if (previousErrorString === codeEvalResult.getErrorString()) {
```
theresa-lee: Is it possible that previousErrorString will be null here?

client/services/FeedbackGeneratorService.js
```diff
@@ -30,6 +30,65 @@ tie.factory('FeedbackGeneratorService', [
...
+    /**
+     * Resets all but one specific counter or all counters, depending on
+     * the parameter.
+     *
+     * @param {string} currentCtr
+     * @private
+     */
+    var _resetCounters = function(currentCtr) {
+      // If the parameter is null, reset all counters.
+      var notResetCtr = currentCtr || '';
```
theresa-lee: I feel like there's probably a better name for this as notResetCtr isn't very clear as to what it is.

dchen97: counterToNotReset maybe?

dchen97:
When we proposed the algorithm for detecting the language unfamiliarity, we only proposed tracking consecutive errors, not alternating ones. I mean, theoretically someone could have alternating errors, but I feel like it's very difficult to say that is due to language unfamiliarity per se since they could just be correcting logical errors and forgetting some of the syntax, correcting it, and then correcting logical errors. However, I think this can be open for discussion to everyone else too, but that's just not what we were aiming to track with our algorithm to begin with.

client/services/FeedbackGeneratorService.js
```diff
@@ -30,6 +30,65 @@ tie.factory('FeedbackGeneratorService', [
...
+      if (notResetCtr !== 'consecutiveSyntaxError') {
```
theresa-lee: I think Diana's comment might have been 'outdated' since recent commits, but yes, if we do end up using strings, these should probably be string constants.

client/services/FeedbackGeneratorService.js
```diff
@@ -30,6 +30,65 @@ tie.factory('FeedbackGeneratorService', [
...
+    /**
+     * Global constant to act as the threshhold before prompting the user
+     * to look at the language primer.
+     *
+     * @type {number}
+     */
+    var UNFAMILIARITY_THRESHOLD = 5;
+
+    /**
+     * Global number to keep track of consecutive syntax errors.
```
seanlip: In what way is this global? It looks like it's scoped to this service.

Maybe s/Global number/Counter. Ditto elsewhere.

client/services/FeedbackGeneratorService.js
```diff
@@ -30,6 +30,65 @@ tie.factory('FeedbackGeneratorService', [
...
+     * @type {number}
+     */
+    var consecutiveSameErrorCounter = 0;
+
+    /**
+     * Global number to keep track of consecutive uses of the wrong language.
+     *
+     * @type {number}
+     */
+    var consecutiveWrongLanguageCounter = 0;
```
seanlip: consecutiveWrongLanguageErrorCounter? Otherwise you seem to be counting languages.

client/services/FeedbackGeneratorService.js
```diff
@@ -30,6 +30,65 @@ tie.factory('FeedbackGeneratorService', [
...
+
+    /**
+     * Global string to store the previous error string.
```
seanlip: I am not sure what this means. Could you elaborate/rephrase?

client/services/FeedbackGeneratorService.js
```diff
@@ -30,6 +30,65 @@ tie.factory('FeedbackGeneratorService', [
...
+     *
+     * @type {string}
+     */
+    var previousErrorString = '';
+
+    /**
+     * Resets all but one specific counter or all counters, depending on
+     * the parameter.
+     *
+     * @param {string} currentCtr
```
seanlip: Eschew abbreviations. Use currentCounter.

client/services/FeedbackGeneratorService.js
seanlip: counterToExclude?

client/services/FeedbackGeneratorService.js
seanlip: Yes, agreed that these should be string constants.

client/services/FeedbackGeneratorService.js
```diff
@@ -280,6 +339,21 @@ tie.factory('FeedbackGeneratorService', [
...
+     * Returns a string when it is detected that the user is unfamiliar
+     * with the programming language.
+     *
+     * @returns {String}
+     * @private
+     */
+    var _getUnfamiliarLanguageFeedback = function() {
+      return [
+        "Seems like you're having some trouble with this language. Why ",
+        "don't you take a look at the New to Python button at the bottom ",
```
seanlip: Should 'New to Python' be in quotes?

Should that string be pulled out into a constant and also used in the button text, so that we can ensure the two of them are always in sync?

client/services/FeedbackGeneratorService.js
```diff
@@ -30,6 +30,65 @@ tie.factory('FeedbackGeneratorService', [
...
+     * @type {number}
+     */
+    var consecutiveSyntaxErrorCounter = 0;
+
+    /**
+     * Global number to keep track of consecutive same errors.
+     *
+     * @type {number}
+     */
+    var consecutiveSameErrorCounter = 0;
```
seanlip: Perhaps consecutiveSameRuntimeErrorCounter, to make clear that you're not including syntax etc. errors here.

client/services/FeedbackGeneratorService.js
```diff
@@ -373,21 +447,58 @@ tie.factory('FeedbackGeneratorService', [
...
           feedback.setReinforcement(
             ReinforcementGeneratorService.getReinforcement(
               tasks[tasks.length - 1], codeEvalResult));
         }
+
+        // If the same error counter reaches the threshold, prompt the user to
+        // look at the primer.
+        if (consecutiveSameErrorCounter === UNFAMILIARITY_THRESHOLD) {
+          feedback.appendTextParagraph(_getUnfamiliarLanguageFeedback());
+          consecutiveSameErrorCounter = 0;
```
seanlip: Hm, why are we doing this here? Could the user be getting the same error because their code works fine, but keeps printing out the wrong solution? In such a case, is looking at the primer actually helpful?

client/services/FeedbackGeneratorService.js
seanlip: I agree with @dchen97, but I don't see the JSDoc.

client/services/FeedbackGeneratorService.js
seanlip: Not done.

client/services/FeedbackGeneratorService.js
```diff
@@ -373,21 +447,58 @@ tie.factory('FeedbackGeneratorService', [
...
+        // error counter.
+        consecutiveSyntaxErrorCounter++;
         var feedback = FeedbackObjectFactory.create(false);
         feedback.appendSyntaxErrorParagraph(errorString);
+
+        // If the syntax error reaches the threshold, prompt the user to
+        // look at the primer.
+        if (consecutiveSyntaxErrorCounter === UNFAMILIARITY_THRESHOLD) {
+          feedback.appendTextParagraph(_getUnfamiliarLanguageFeedback());
+          consecutiveSyntaxErrorCounter = 0;
```
seanlip: Should this only be reset when they look at the primer?

client/services/FeedbackGeneratorService.js
```diff
@@ -457,16 +568,25 @@ tie.factory('FeedbackGeneratorService', [
             'in getPrereqFailureFeedback().'].join());
         }
 
+        // If the wrong language counter reaches the threshold, prompt the user
+        // to look at the primer.
+        if (consecutiveWrongLanguageCounter === UNFAMILIARITY_THRESHOLD) {
+          feedback.appendTextParagraph(_getUnfamiliarLanguageFeedback());
+          consecutiveWrongLanguageCounter = 0;
```
seanlip: Ditto.

client/services/FeedbackGeneratorService.js
eledavi: I changed it to counterNotToReset. PTAK and LMK if you think of a better name.

eledavi: **PTAL

client/services/FeedbackGeneratorService.js
dchen97: Elena and I talked about this, and we thought it might be best not to keep bugging them to look at the primer if they elect not to. The message will still be present in the feedback history so they can look at the primer whenever they so choose, but we didn't want the constant reminders to be annoying.

client/services/FeedbackGeneratorService.js
eledavi: Oops, I thought I added that but I didn't. My apologies

client/services/FeedbackGeneratorService.js
dchen97: Well theoretically this could happen. We were thinking, though, if the user is getting the exact same error string (i.e. same error type with the same traceback) 5 times in a row, then odds are that either their code isn't changing or they don't understand what is the problem with the code to begin with. This is an extrapolation we made, though, and is definitely open for discussion. This is just what we came up with this when we were thinking of factors to indicate language unfamiliarity.

seanlip: I agree with "either their code isn't changing or they don't understand what is the problem with the code to begin with". But does that imply that the underlying issue is a language syntax/construct issue?

client/services/FeedbackGeneratorService.js
seanlip: Makes sense, thanks for explaining. Consider adding a comment in the code to explain the rationale behind the design decision here, so that it provides context for the existing behaviour in case someone wants to tweak it in the future.

client/services/FeedbackGeneratorService.js
dchen97: Elena and I both agree with what you're saying, and without real data, it'll be hard to tell how accurate any of our language unfamiliarity triggers are. For now, our belief is that if someone is familiar with the language and its syntax, then they probably would have tried something logically different with their code after receiving the same error 2 or 3 times in a row, which would likely throw either other types of errors or at least different error tracebacks. After 5 times, it seems like the user might be grasping at straws. Nonetheless, we could also just leave this trigger out for now if we feel it's unnecessary.

seanlip: OK. I suggest polling the rest of the team for opinions, and then you and Elena can make a decision based on the inputs and we'll test it in our next set of user tests.

dchen97:
@seanlip We figured we should use different counters because the way we thought that the algorithm would work is that we would track consecutive syntax and runtime errors, rather than just syntax and runtime errors, because it's plausible that a user could be dealing with back-to-back and alternating syntax and runtime errors in the process of logically working through the problem. However, it's much more likely when a user has 5 syntax errors in a row or 5 of the same runtime errors in a row that they are really just not understanding the language or errors being given. As said above, though, this is just an assumption that Elena and I have made when we were trying to figure out the algorithm for language unfamiliarity detection.

client/services/FeedbackGeneratorService.js
```diff
@@ -24,12 +24,76 @@ tie.factory('FeedbackGeneratorService', [
...
+     * Resets all but one specific counter or all counters, depending on
+     * the parameter.
+     *
+     * @param {string} currentCounter
+     * @private
+     */
+    var _resetCounters = function(currentCounter) {
+      // If the parameter is null, reset all counters.
+      var counterNotToReset = currentCounter || '';
+      if (counterNotToReset !== 'consecutiveSyntaxError') {
```
theresa-lee: I think it might have gotten lost in the outdated comments, but can we make these constants? Since we use them here and also in the other files when we call this method.

seanlip:
> @seanlip  We figured we should use different counters because the way we thought  that the algorithm would work is that we would track consecutive syntax  and runtime errors, rather than just syntax and runtime errors, because  it's plausible that a user could be dealing with back-to-back and  alternating syntax and runtime errors in the process of logically  working through the problem. However, it's much more likely when a user  has 5 syntax errors in a row or 5 of the same runtime errors in a row  that they are really just not understanding the language or errors being  given. As said above, though, this is just an assumption that Elena and  I have made when we were trying to figure out the algorithm for  language unfamiliarity detection.

But this isn't just a "runtime error" -- this is a "using a different language's construct" error, right? Which seems to indicate "unfamiliar with the language" to me... (On the other hand, if it were a generic runtime error, then I would agree with you.)

Perhaps I'm misunderstanding the characterization here, and I'm happy to chat about this if it would be helpful.

dchen97:
@seanlip Ahh OK I misread your initial comment and assumed it was discussing the same runtime error counter. Elena and I didn't really discuss having it as one counter, but we can talk about it once we get into the office today.

client/services/FeedbackGeneratorService.js

dchen97:
@seanlip 
@eledavi and I have talked and we agree that the suggestion to combine both the syntax error and wrong language counters makes sense. Elena is working on adjusting the files now and should send updates soon.

client/services/FeedbackGeneratorService.js
```diff
@@ -47,11 +45,11 @@ tie.factory('FeedbackGeneratorService', [
     var UNFAMILIARITY_THRESHOLD = 5;
 
     /**
-     * Counter to keep track of consecutive syntax errors.
+     * Counter to keep track of language unfamiliarity errors.
```
dchen97: We should probably add a note to this JSDoc comment that a "language unfamiliarity" error includes syntax errors and wrong language errors.

eledavi: Done!

client/services/FeedbackGeneratorService.js
```diff
@@ -30,6 +30,65 @@ tie.factory('FeedbackGeneratorService', [
     RUNTIME_ERROR_FEEDBACK_MESSAGES, WRONG_LANGUAGE_ERRORS, LANGUAGE_PYTHON,
     CLASS_NAME_AUXILIARY_CODE, CLASS_NAME_SYSTEM_CODE, PARAGRAPH_TYPE_TEXT,
     PARAGRAPH_TYPE_CODE, PARAGRAPH_TYPE_SYNTAX_ERROR) {
+
+    /**
+     * Global constant to act as the threshhold before prompting the user
```

client/services/FeedbackGeneratorService.js

client/services/FeedbackGeneratorService.js

client/services/FeedbackGeneratorService.js

client/services/FeedbackGeneratorService.js

client/services/FeedbackGeneratorService.js

client/services/FeedbackGeneratorService.js

client/services/FeedbackGeneratorService.js

client/services/FeedbackGeneratorService.js

client/services/FeedbackGeneratorService.js
dchen97: Done! Added a null-check to the if statement

client/services/FeedbackGeneratorService.js
dchen97: Now it's done!

client/services/FeedbackGeneratorService.js

dchen97:
@seanlip and @theresa-lee 

This should be ready for another pass.

client/question/question.js
```diff
@@ -93,6 +93,38 @@ tie.constant('CLASS_NAME_AUXILIARY_CODE', 'AuxiliaryCode');
...
+ * @type {string}
+ * @constant
+ */
+tie.constant('PYTHON_PRIMER_BUTTON_NAME', 'New to Python?');
+
+/**
+ * Constant to represent the consecutiveLanguageUnfamiliarityCounter option
+ * for the
+ *
+ * resetCounters function in FeedbackGeneratorService. Is to be passed into the
```
seanlip: Why the line breaks?

client/question/question.js
```diff
@@ -185,6 +217,15 @@ tie.constant('PARAGRAPH_TYPE_CODE', 'code');
 tie.constant('PARAGRAPH_TYPE_SYNTAX_ERROR', 'error');
 
 /**
+ * Constant for the number of times that a user can make a mistake (i.e.
+ * same error, syntax error, etc.) until we prompt them to look at the
+ * primer.
+ *
+ * @type {number}
```
seanlip: Should have `@constant` annotation?

client/question/services/FeedbackGeneratorService.js
```diff
@@ -24,12 +24,60 @@ tie.factory('FeedbackGeneratorService', [
...
+     *
+     * @type {string}
+     */
+    var previousErrorString = '';
+
+    /**
+     * Resets all but one specific counter or all counters, depending on
+     * the parameter.
+     *
+     * @param {string} currentCounter
```
seanlip: Why not call this "counterNotToReset"? I think that clarifies its purpose within this function better.

EDIT: this comment may be irrelevant -- see comment below.

client/question/services/FeedbackGeneratorService.js
```diff
@@ -24,12 +24,60 @@ tie.factory('FeedbackGeneratorService', [
...
+     */
+    var _resetCounters = function(currentCounter) {
+      // If the parameter is null, reset all counters.
+      var counterNotToReset = currentCounter || '';
+      if (counterNotToReset !== EXCLUDE_LANGUAGE_UNFAMILARITY_ERROR) {
+        consecutiveLanguageUnfamiliarityCounter = 0;
+      }
+      if (counterNotToReset !== EXCLUDE_CONSECUTIVE_SAME_RUNTIME_ERROR) {
+        consecutiveSameRuntimeErrorCounter = 0;
+        previousErrorString = '';
```
seanlip: Should this be previousRuntimeErrorString (since you're dealing with a specific class of error here)?

That said, this probably shouldn't even be here -- it's outside what this function's supposed to do. Could the caller do it instead?

client/question/services/FeedbackGeneratorService.js
```diff
@@ -280,6 +328,26 @@ tie.factory('FeedbackGeneratorService', [
...
+     * the user is unfamiliar with the given programming language.
+     *
+     * @param {string} language
+     * @returns {String}
+     * @private
+     */
+    var _getUnfamiliarLanguageFeedback = function(language) {
+      if (language === LANGUAGE_PYTHON) {
+        return [
+          "Seems like you're having some trouble with this language. Why ",
```
seanlip: Perhaps "this language" --> "Python"?

client/question/question.js
```diff
@@ -93,6 +93,38 @@ tie.constant('CLASS_NAME_AUXILIARY_CODE', 'AuxiliaryCode');
...
+ * for the
+ *
+ * resetCounters function in FeedbackGeneratorService. Is to be passed into the
+ * resetCounters function as a parameter when we want to reset all counters
+ * except the consecutiveLanguageUnfamiliarityCounter.
+ *
+ * @type {string}
+ * @constant
+ */
+tie.constant('EXCLUDE_LANGUAGE_UNFAMILARITY_ERROR',
```
seanlip: ERROR_COUNTER_LANGUAGE_UNFAMILIARITY
ERROR_COUNTER_SAME_RUNTIME

("error counter" is what it is, and the suffix is what type of error counter it is. Starting with 'exclude' makes it sound like a verb)

client/question/services/FeedbackGeneratorService.js
```diff
@@ -24,12 +24,60 @@ tie.factory('FeedbackGeneratorService', [
...
+    var previousErrorString = '';
+
+    /**
+     * Resets all but one specific counter or all counters, depending on
+     * the parameter.
+     *
+     * @param {string} currentCounter
+     * @private
+     */
+    var _resetCounters = function(currentCounter) {
```
seanlip: Actually, on reading more of the file I think the design here isn't quite right. We're sort of spreading the info about which counter to increment all over the place (since the caller takes care of incrementing one counter and this one takes care of resetting all the others). That's a bit confusing.

Can we instead have a helper function along the lines of:

    updateCounters(countersToIncrement)

that increments each counter in the given list, and resets all other counters to 0? That way this function can handles all counter management and you aren't modifying the counters outside of this function, which would be cleaner.

client/question/services/FeedbackGeneratorService.js
```diff
@@ -417,21 +485,82 @@ tie.factory('FeedbackGeneratorService', [
...
+        // Reset all counters but consecutiveLanguageUnfamiliarityCounter.
+        _resetCounters(EXCLUDE_LANGUAGE_UNFAMILARITY_ERROR);
+
+        // If the user receives another syntax error, increment the
+        // language unfamiliarity error counter.
+        consecutiveLanguageUnfamiliarityCounter++;
         var feedback = FeedbackObjectFactory.create(false);
         feedback.appendSyntaxErrorParagraph(errorString);
+
+        // If the syntax error reaches the threshold, prompt the user to
```
seanlip: This logic keeps repeating. I suggest pulling it out into a separate function that checks the counters and appends feedback / resets the counters as needed. The feedback should probably only be appended once even if both counters hit the threshold (I realize this is not possible with the current configuration but it provides some robustness against future changes, e.g. we may end up having a type of error that increments both counters at once).

client/question/question.js
eledavi: Done!

client/question/question.js
eledavi: Done!

client/question/services/FeedbackGeneratorService.js
eledavi: Done!

client/question/services/FeedbackGeneratorService.js
eledavi: I created the _updateCounters() and _checkThreshold() methods based off of your comments. PTAL and LMK your thoughts.

client/question/services/FeedbackGeneratorService.js
```diff
@@ -24,12 +24,85 @@ tie.factory('FeedbackGeneratorService', [
...
+     */
+    var _updateCounters = function(counterToIncrement) {
+      // If the parameters are empty clear all counters.
+      if (counterToIncrement === undefined) {
+        consecutiveLanguageUnfamiliarityCounter = 0;
+        consecutiveSameRuntimeErrorCounter = 0;
+      } else if (counterToIncrement === ERROR_COUNTER_LANGUAGE_UNFAMILIARITY) {
+        consecutiveLanguageUnfamiliarityCounter++;
+        consecutiveSameRuntimeErrorCounter = 0;
+        previousErrorString = '';
```
seanlip: This should not be touched here. This method should just update counters, and nothing else. Let the caller handle the error string.

client/question/services/FeedbackGeneratorService.js
```diff
@@ -24,12 +24,85 @@ tie.factory('FeedbackGeneratorService', [
...
+      } else if (counterToIncrement === ERROR_COUNTER_SAME_RUNTIME) {
+        consecutiveSameRuntimeErrorCounter++;
+        consecutiveLanguageUnfamiliarityCounter = 0;
```
seanlip: Optional: Consider swapping this and the line above, so that we always handle the two counters in the same order. Easier to pattern-match that way.

client/question/services/FeedbackGeneratorService.js
```diff
@@ -24,12 +24,85 @@ tie.factory('FeedbackGeneratorService', [
...
+
+    /**
+     * Checks the error counters to if any have reached the threshold. If they
+     * have, then it appends to the feedback to prompt the user to look at the
+     * primer.
+     *
+     * @param {Feedback} feedback
+     * @private
+     */
+    var _checkThreshold = function(feedback) {
```
seanlip: To me, _checkThreshold sounds like something that returns a boolean or does some kind of check, but this function seems to have a lot of side-effects and this should be indicated in the name somehow. Finding a good name that indicates this is tricky, though. I'm thinking of, e.g., _applyThresholdUpdates or something more "active" like that, but you might have a better idea.

client/question/services/FeedbackGeneratorService.js
```diff
@@ -24,12 +24,85 @@ tie.factory('FeedbackGeneratorService', [
...
+    var previousErrorString = '';
+
+    /**
+     * Updates the counter specified in the paramter to track language
+     * unfamiliarity and resets all other counters.
+     *
+     * @param {string} counterToIncrement
+     * @private
+     */
+    var _updateCounters = function(counterToIncrement) {
```
seanlip: You are passing in a string here, but in some cases you call this method with `undefined` as an arg. I suggest that, in those cases, you call a separate _resetCounters() method instead that takes no args, and have this function assume in all cases that its input is a string.

eledavi: Done!

client/question/services/FeedbackGeneratorService.js
eledavi: Done!

client/question/services/FeedbackGeneratorService.js
eledavi: Done!

client/question/services/FeedbackGeneratorService.js
eledavi: Done!

client/question/services/FeedbackGeneratorService.js
```diff
@@ -417,21 +515,56 @@ tie.factory('FeedbackGeneratorService', [
        * @returns {Feedback}
        */
       getFeedback: function(tasks, codeEvalResult, rawCodeLineIndexes) {
+        // If the user receives the same error message increment the same error
+        // counter.
+        if (previousErrorString &&
+            previousErrorString === codeEvalResult.getErrorString()) {
+          _updateCounters(ERROR_COUNTER_SAME_RUNTIME);
+        }
```
seanlip: Just to check: is the intended behavior that, if the above condition does not hold, the counters etc. remain as they are (and don't get reset)? If so, that's fine. I just wanted to confirm.

client/question/services/FeedbackGeneratorService.js
```diff
@@ -417,21 +515,56 @@ tie.factory('FeedbackGeneratorService', [
...
+       * Returns the Feedback object for the given syntax error string.
+       *
+       * @param {string} errorString
+       * @returns {Feedback}
+       */
       getSyntaxErrorFeedback: function(errorString) {
+        // If the user receives another syntax error, increment the
+        // language unfamiliarity error counter.
+        _updateCounters(ERROR_COUNTER_LANGUAGE_UNFAMILIARITY);
+        previousErrorString = '';
```
seanlip: Is this "error string" supposed to be for general errors or just the specific case of runtime errors? If the latter, perhaps it should be renamed to previousRuntimeErrorString.

client/question/services/FeedbackGeneratorService.js
```diff
@@ -24,12 +24,90 @@ tie.factory('FeedbackGeneratorService', [
...
+     * @private
+     */
+    var _updateCounters = function(counterToIncrement) {
+      if (counterToIncrement === ERROR_COUNTER_SAME_RUNTIME) {
+        consecutiveSameRuntimeErrorCounter++;
+        consecutiveLanguageUnfamiliarityCounter = 0;
+      } else if (counterToIncrement === ERROR_COUNTER_LANGUAGE_UNFAMILIARITY) {
+        consecutiveSameRuntimeErrorCounter = 0;
+        consecutiveLanguageUnfamiliarityCounter++;
+      }
```
seanlip: (Forgot one more thing.) Should we also do "else, throw Error" here?

client/question/services/FeedbackGeneratorService.js
eledavi: Yes, y'all are right. I just went back and corrected that!

client/question/services/FeedbackGeneratorService.js
```diff
@@ -24,12 +24,90 @@ tie.factory('FeedbackGeneratorService', [
...
+     *
+     * @private
+     */
+    var _resetCounters = function() {
+      consecutiveSameRuntimeErrorCounter = 0;
+      consecutiveLanguageUnfamiliarityCounter = 0;
+    };
+
+    /**
+     * Checks the error counters to if any have reached the threshold. If they
```
eledavi: Done!

client/question/services/FeedbackGeneratorService.js
eledavi: Done!

client/question/services/FeedbackGeneratorService.js
eledavi: Done!

seanlip:
OK, LGTM. Thanks @eledavi!

@dchen97 -- looks like you have an outstanding hold on this PR. Could you either approve / request further changes, please?

PR #385, opened 2017-08-02
Add Print Statement Feedback

dchen97:
* Added methods in `FeedbackGeneratorService` to detect print statements and append feedback to explain print statements are not supported
* Added related tests
* Adjusted feedback methods so that RuntimeError, InfiniteLoop, and all other types of feedback (except Syntax and PrereqFailure) can be appended to a previously created feedback object.
* Adjusted the `.eslintrc.json` to fix linter errors from new tests.


client/services/FeedbackGeneratorService.js
```diff
@@ -129,13 +132,15 @@ tie.factory('FeedbackGeneratorService', [
      *    name for the function needed to process/format the user's output.
      * @param {CorrectnessTest} correctnessTest
      * @param {*} observedOutput Actual output for running user's code.
+     * @param {Feedback} oldFeedback Feedback object to append feedback to. If
```
seanlip: Instead of adding this everywhere, would it make sense to add it in the caller instead? It keeps the contract of the individual functions clearer, and perhaps you can add a helper method to the Feedback object class along the lines of feedback.appendFeedback(newFeedback).

dchen97: Yeah that makes sense. I've added the appendFeedback helper method to the `Feedback` object class now and have used that instead in my implementation. It has definitely helped with simplifying much of the code.

client/domain/FeedbackObjectFactory.js
```diff
@@ -167,6 +167,19 @@ tie.factory('FeedbackObjectFactory', [
...
+     * Appends the feedback paragraphs in the given Feedback object to the
+     * current feedback.
+     *
+     * @param {Feedback} feedbackToAppend
+     */
+    Feedback.prototype.appendFeedback = function(feedbackToAppend) {
+      var paragraphsToAppend = feedbackToAppend.getParagraphs();
+      paragraphsToAppend.forEach(function(paragraph) {
+        this._paragraphs.push(paragraph);
+      }.bind(this));
```
seanlip: Interesting -- I've never actually seen this syntax before, but I like it! (I always used to use `that = this` and reference `that`.)

theresa-lee:
Looks good! Only thing is if we should change the color of the text at all to make it more obvious. It sort of blends in as like explanation text (since same color as intro paragraphs).

dchen97:
Hmmm so looking at the code right now, I feel like there are 2 possible ways that I can try to change the color of the text:

1. Add another FeedbackParagraph type (i.e. 'warning' type) that would change any feedback paragraph of that type to have a different font color.
2. Add another property to FeedbackParagraph called `styleClasses` or `additionalStyleClasses` so that for any FeedbackParagraph of any type, we can just add any additional CSS classes that we'd specify.

1 would probably be easier, and I think with 2, we'd have to make some design decisions (i.e. on which element do we apply the class for Syntax Error paragraphs) but it might be easier in the long run if we want to specially style any other FeedbackParagraphs that - for the most part - would be styled the same way as pre-existing FeedbackParagraph type.

theresa-lee:
On second thought, we could just leave it as is, and then in the next round of user tests, see if that's actually an issue. My main point of concern is just that they would be trying to print out things and glossing over/missing the feedback that tells them that, but that might not happen at all. We can always change it if it is an issue. But yes, I would probably go with approach 1 if we were going to add color.

PR #386, opened 2017-08-03
Frontend (client/...) restructure directory

theresa-lee:
- under client, there are 3 sub folders: menu, question and a shared folder
- app.html becomes question.html, which is now located in the question directory


theresa-lee:
Am unsure if I edited the config files correctly, but karma seems to be running fine, as well as e2e (though it could be that it's running okay because the config file isn't correct..).

karma.conf.js
```diff
@@ -20,8 +20,8 @@ module.exports = function(config) {
       'third_party/skulpt-c14015/skulpt.min.js',
       'third_party/skulpt-c14015/skulpt-stdlib.js',
       'third_party/ui-codemirror-0.3.0/ui-codemirror.min.js',
-      'client/app.js',
-      'client/**/*.js',
+      'client/question/question.js',
+      'client/**/**/*.js',
```
seanlip: I think you can leave this as it was, and in fact you don't need the line above either. That's because, in the config file `**/` represents multiple dirs, I think -- could you please check? If so we can just replace this with `client/**/*.js`.

karma.conf.js
```diff
@@ -33,8 +33,8 @@ module.exports = function(config) {
     exclude: [],
     // Pre-process matching files before serving them to the browser.
     preprocessors: {
-      'client/*.js': preprocessorOptions,
       'client/**/*.js': preprocessorOptions,
+      'client/**/**/*.js': preprocessorOptions,
```
seanlip: I think you don't need this line, for the same reason. Could you please verify that, and remove this if so?

client/menu/menu.css
```diff
@@ -0,0 +1,47 @@
+body {
```
seanlip: Use spaces, not tabs. Include license info at top.

client/menu/menu.html
```diff
@@ -0,0 +1,104 @@
...
+<html>
+  <head>
+    <meta charset="UTF-8">
+    <meta name="viewport" content="width=device-width, initial-scale=1.0, user-scalable=yes">
+    <link rel="stylesheet" href="../../third_party/codemirror-5.19.0/lib/material.css">
+    <link rel="stylesheet" href="menu.css">
+    <script src="../../third_party/angular-1.6.1/angular.min.js"></script>
+  </head>
+  <body>
+    <div class="tie-menu-question-list-wrapper">
```
seanlip: Ah, so, when I talked about a dummy menu page, I just meant something that has a text/pointer that redirects to our current question page -- but the whole thing needs to still be fully usable; this is no longer just a development prototype. People are downloading and using the current application, so every change to it needs to result in a coherent UX.

You could perhaps have a single card that advertises a question set on strings, or something similar. But the data needs to be real and the UX resulting from this change needs to make sense.

theresa-lee: So can I do it without a JS file or do you want me to actually pull the data from the assets?

karma.conf.js
seanlip: Ah ok. Could you leave a comment in that explains this clearly? Thanks!

client/menu/menu.html
seanlip: I'm agnostic to that; it's fine to not use a JS file or the assets for now. It's fine to do something simple now as a placeholder, as long as it makes sense and points to the question page (in fact just one link overall is probably better given that we're still showing the same question page regardless of which is clicked).

karma.conf.js
theresa-lee: Done.

client/menu/menu.html
theresa-lee: Alright. I kept one of the cards since I feel like we might still be able to use the structure that I put in place for the future cards, and then added a greeting text. It's fairly simple/not stylized (I would add a logo but I don't think we have a PNG of the file) so let me know if you think I should add a bit more CSS to style it up.

theresa-lee:
Okay, PTAL!

PR #387, opened 2017-08-09
Fix Java Comment RegEx Bug

dchen97:
* Fixed bug where the ZeroDivisionError would get blocked by the Java Comment Regex for Wrong Language Syntax Detection

### DIFF
--- a/client/question/question.js
+++ b/client/question/question.js
@@ -248,7 +248,7 @@ tie.constant('WRONG_LANGUAGE_ERRORS', {
   }, {
     // Used the Java comment syntax
     errorName: 'javaComment',
-    regExString: '\\/(\\s*|\\w*)*\\n|\\/\\*(\\*)?(\\s*|\\w*)*\\*\\/',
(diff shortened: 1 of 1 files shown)


client/question/question.js
```diff
@@ -248,7 +248,7 @@ tie.constant('WRONG_LANGUAGE_ERRORS', {
   }, {
     // Used the Java comment syntax
     errorName: 'javaComment',
-    regExString: '\\/(\\s*|\\w*)*\\n|\\/\\*(\\*)?(\\s*|\\w*)*\\*\\/',
+    regExString: '\\/\\/(\\s*|\\w*)*\\n|\\/\\*(\\*)?(\\s*|\\w*)*\\*\\/',
```
seanlip: Btw, are the two inner asterisks in `(\\s*|\\w*)*` (both occurrences) necessary? (Wouldn't the asterisk outside the parens take care of that?)

dchen97:
I think in the confusion of all the double backslashes I just forgot one ><

client/question/question.js
dchen97: No, they don't appear to be necessary. I'll delete them.

dchen97: Done!

PR #390, opened 2017-08-12
Pulling Question Data into it's own module, tieData

theresa-lee:


theresa-lee:
@seanlip PTAL

client/question/question.html
```diff
@@ -10,9 +10,10 @@
     <script src="../../third_party/codemirror-5.19.0/mode/python/python.js"></script>
     <script src="../../third_party/codemirror-5.19.0/addon/edit/matchbrackets.js"></script>
     <script src="../../third_party/ui-codemirror-0.3.0/ui-codemirror.min.js"></script>
-    <script src="../shared/modules/config.js"></script>
+    <script src="../config/config.js"></script>
```
seanlip: Do you think we could group these imports by module? I.e. all the ../config stuff first, then one blank line, then all the ../data stuff, then one blank line, then all the other stuff? Slightly easier to grok that way, I think.

seanlip:
@eyurko please also note the change in file structure. We got rid of shared/modules and decided to make all the modules top-level (so client/question, client/menu, client/config, client/data) otherwise it starts to get a bit messy because e.g. the QuestionPage module depends on the data module which depends on the config module.

client/question/question.html
theresa-lee: Done.

seanlip:
Oh, wait, not quite yet. Tests are failing.

theresa-lee:
Okay, sorry about that. @eyurko I restructured the server_URL constant into tieConfig since that seemed like a more appropriate place for that. I think we'll want to move the server handler stuff in there as well but perhaps in a different PR.

seanlip:
NB: as discussed, we might want  to name that file something more general than serverUrl.js. Perhaps deploymentSpecificConfig.js (or similar).

PR #393, opened 2017-08-18
Added menu.js + controller to retrieve question data

theresa-lee:
- Renamed data file to be consistent (all other files did not have the 'tie' in front)
- Created a menu.js file with a controller to grab the data from the other module services
- Moved shared constants

### DIFF
--- a/client/menu/menu.js
+++ b/client/menu/menu.js
@@ -13,7 +13,17 @@
 // limitations under the License.
 
 /**
- * @fileoverview Module for storing question and user data
+ * @fileoverview Adds an importable configuration file for the TIE menu page.
  */
 
-window.tieData = angular.module('tieData', []);
+window.tieMenu = angular.module('tieMenu', ['tieData']);
+
+tieMenu.controller('MenuPageController',
+  ['$scope', 'QuestionDataService', function($scope, QuestionDataService) {
+    var questionSetId = 'strings';
+    QuestionDataService.initCurrentQuestionSet(questionSetId);
+    $scope.questionSet = QuestionDataService.getCurrentQuestionSet(
+      questionSetId);
+    $scope.questionIds = $scope.questionSet.getQuestionIds();
+  }]
+);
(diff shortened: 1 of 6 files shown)


client/menu/menu.js
```diff
@@ -13,7 +13,17 @@
...
  */
 
-window.tieData = angular.module('tieData', []);
+window.tieMenu = angular.module('tieMenu', ['tieData']);
+
+tieMenu.controller('MenuPageController',
+  ['$scope', 'QuestionDataService', function($scope, QuestionDataService) {
+    var questionSetId = 'strings';
+    QuestionDataService.initCurrentQuestionSet(questionSetId);
+    $scope.questionSet = QuestionDataService.getCurrentQuestionSet(
```
theresa-lee: You can do a console.log for the questionSet or the questionIds here to see it grabbing correct values. I had to remove because linter (and we wouldn't want console logs in our production code anyway).

client/menu/menu.js
```diff
@@ -13,7 +13,17 @@
...
 
 /**
- * @fileoverview Module for storing question and user data
+ * @fileoverview Adds an importable configuration file for the TIE menu page.
  */
 
-window.tieData = angular.module('tieData', []);
+window.tieMenu = angular.module('tieMenu', ['tieData']);
+
+tieMenu.controller('MenuPageController',
```
seanlip: This seems reasonable -- but, in this or a future PR, could you please pull this and the menu page HTML out into a menu-view directive in order to mimic what is done in the question page?

client/menu/menu.js
seanlip: Yep, removing console.logs is absolutely the right thing to do here :)

When you extract into a directive, please pull this initialization logic into a function and write Karma tests for it.

theresa-lee: Okay, will do!

client/menu/menu.js
theresa-lee: Oh wait, sorry, rereading over the final comments below, I realize you were saying to pull this into a view (like Learner's View). I'll try and do that in a following PR.

theresa-lee:
Yes, so the next step is for me to make the controller logic into a directive as a menu view (+ tests) and then pull the data from the service to populate the cards!

PR #397, opened 2017-08-25
First pass at improving reverseWords question.

seanlip:
This is a first pass at adding more feedback to the reverseWords question. There's yet more feedback that can be added, but I'm blocked by a lack of functionality, so I figured I'd submit this first, add the missing functionality, and then do another pass at adding more feedback.

I also found and fixed a couple of existing bugs (@theresa-lee PTAL):
- hintIndex was not being carried over when the code received "don't use print statements" feedback, so only the first message of a feedback set was ever being shown. This resulted in a frustrating UX since the later feedback was never reached.
- The loading dots were overlapping the reinforcements, which looked ugly. I gave them a top margin.

The additional functionality I plan to add in future PRs is the following:
- After giving a sequence of hints, show "you output X but we expected Y", so at least it's clear exactly which input is failing.
- Allow custom feedback to be provided, based on which _test suites_ passed/failed.
- Allow buggy output tests to be run on only a subset of the test suites. This is because the code could fail in one way for some of the "easier" test suites, which allows appropriate feedback to be given -- but then fail in multiple ways for the "corner case" tests depending on how buggy the implementation is. This makes it hard to write custom buggy functions to catch such errors, since currently all buggy functions run on all test suites.

### DIFF
--- a/assets/questions/reverseWords.js
+++ b/assets/questions/reverseWords.js
@@ -28,124 +28,388 @@ globalData.questions['reverseWords'] = {  // eslint-disable-line dot-notation
     python:
 `class AuxiliaryCode(object):
     @classmethod
-    def forgetLastWord(cls, s):
-        result = ""
-        reversed_word = []
-        for c in s:
-            if c.isspace():
-                result += "".join(reversed(reversed_word))
-                result += c
-                reversed_word = []
-            else:
-                reversed_word.append(c)
-        return result
+    def _reverseCorrectly(cls, s, isCharValid):
+      result = ''
+      reversed_word = ''
+      for c in s:
+        if isCharValid(c):
+          result += reversed_word[::-1]
+          result += c
+          reversed_word = ''
+        else:
+          reversed_word += c
+      if reversed_word:
+        result += reversed_word[::-1]
+      return result
+
+    @classmethod
+    def _reverseCorrectlyWithGivenDelimiters(cls, s, delimiters):
+      return cls._reverseCorrectly(s, lambda x: x not in delimiters)
+
+    @classmethod
+    def returnNone(cls, s):
+      return None
+
+    @classmethod
+    def returnSampleOutput(cls, s):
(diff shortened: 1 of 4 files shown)


assets/questions/reverseWords.js
```diff
@@ -28,124 +28,388 @@ globalData.questions['reverseWords'] = {  // eslint-disable-line dot-notation
...
+          'Adding spaces at ends of words to keep things simple is a nice ',
+          'idea, but be careful about how you handle the extra space at the ',
+          'end.'
+        ].join('')
+      ]
+    }, {
+      buggyFunctionName: 'AuxiliaryCode.treatAsciiCharsContiguously',
+      messages: [
+        [
+          'Check the ASCII tables -- are you sure that "A"-"z" be treated as ',
```
theresa-lee: Should 'be' -> is/are?

assets/questions/reverseWords.js
```diff
@@ -28,124 +28,388 @@ globalData.questions['reverseWords'] = {  // eslint-disable-line dot-notation
...
+          'a contiguous block of characters?'
+        ].join('')
+      ]
+    }, {
+      buggyFunctionName: 'AuxiliaryCode.useWeakInequalitiesForCharValidity',
+      messages: [
+        [
+          'Are you correctly handling the check for whether a character is a ',
```
theresa-lee: Does there need to be an "or not" in here? I can't tell, it just feels like it's missing for me

assets/questions/reverseWords.js
seanlip: Thanks, good catch! Fixed.

assets/questions/reverseWords.js
seanlip: I checked -- consensus [here](https://english.stackexchange.com/questions/3382/whether-or-not-vs-whether) seems to be to just use "whether", so I'm going to go for that.

PR #405, opened 2017-09-20
Add Python Primer doc (2 themes)

rabidbit:


client/docs/py-primer-dark.html
```diff
@@ -0,0 +1,178 @@
...
+      }
+      .transparent {
+        font-size: 0px;
+        height: 0px;
+        opacity: 0;
+      }
+    </style>
+  </head>
+  <body>
+    <div class="h1">
```
seanlip: Why not just use h1, h2 tags (and style those appropriately if needed)? These are (semantically) headers, and one should use the right tags for accessibility reasons.

client/docs/py-primer-dark.html
```diff
@@ -0,0 +1,178 @@
...
+      <span class="transparent">"""</span>
+      PYTHON IN 2 MINUTES (OR SO)
+      <span class="transparent">"""</span>
+    </div>
+    <p>
+      <span class="transparent">"""</span>The following is a very short primer
```
seanlip: Not sure I understand the class="transparent" thing. Can a similar effect be achieved with CSS rather than using """ characters?

rabidbit: This page supports full page copy/paste onto Python interpreter. The section headings, therefore, need to be commented out with """, but stylistically, I don't want the section headers or the footer to show """. If you do a select all on the page and paste onto a test file, you'll see the headers and footer are commented out. It's one less thing for the user to get tripped up on if they decide to copy/paste from this page.

client/docs/py-primer-dark.html
rabidbit: Changed as requested.

client/docs/py-primer-dark.html
seanlip: Ah. If that's the motivation, I recommend not optimizing for that. There's pedagogical value in typing stuff out, rather than copy/pasting.

If you want to keep it, though, consider adding an inline comment above the first use of the class (in both files) to explain the rationale, so that others understand why it's there too.

rabidbit: Added comment below style section.

client/docs/py-primer-dark.html
```diff
@@ -0,0 +1,179 @@
...
+      .string {
+        color: #a0b97c;
+      }
+      .transparent {
+        font-size: 0px;
+        height: 0px;
+        opacity: 0;
+      }
+    </style>
+    <--! Note: transparent style used to hide three doublequotes used to comment out doc headers, in case user decides to copy/paste that part of the page -->
```
seanlip: HTML comments start with `<!––`.

rabidbit: Woops. Sorry. Fixed.

rabidbit:
Yes, I realize it's not ideal, but we needed these pages on GitHub for our current user studies, and it's blocking adding a link to the primer from the stand-alone version. Agree we can optimize later.

seanlip:
Ack.

There also seems to be an issue with travis. I'll look into it.

seanlip:
@rabidbit could you please merge master into this branch? I think that will fix the travis issues.

PR #406, opened 2017-09-21
Privacy Modal

theresa-lee:
This modal has different behavior for the client and server version:

For the client version, this will not show up on page load, but will give a popup if you click on the 'Privacy' text. It's just text and an okay button. 

For the server version, this will show up on page load (unless you have accepted beforehand), and if the user accepts and it's not a shared computer, the cookie will last for 1 year. If it is a shared computer, it will last for the session. If the user rejects, it will bring them back to the menu page.

Cookies don't work for local files so you won't be able to test that locally I think. If you toggle the 2 instances of "SERVER_URL" in LearnerView, then you should be able to see the modal change (however, it still won't remember your decision).


client/question/components/LearnerViewDirective.js
```diff
@@ -137,6 +137,8 @@ tie.directive('learnerView', [function() {
           </div>
         </div>
       </div>
+      <privacy-modal is-displayed='privacyModalIsDisplayed' is-client-version="!SERVER_URL">
```
seanlip: Please use double quotes around attr values, for consistency.

client/question/components/LearnerViewDirective.js
```diff
@@ -489,19 +499,19 @@ tie.directive('learnerView', [function() {
       </style>
     `,
     controller: [
-      '$scope', '$interval', '$timeout', '$location', 'SolutionHandlerService',
-      'QuestionDataService', 'LANGUAGE_PYTHON', 'FeedbackObjectFactory',
-      'ReinforcementObjectFactory', 'LocalStorageService',
-      'ServerHandlerService', 'SECONDS_TO_MILLISECONDS',
+      '$scope', '$interval', '$timeout', '$location', '$cookies',
```
seanlip: Move dependency on $cookies to a separate service, as discussed.

client/question/components/PrivacyModalDirective.js
```diff
@@ -0,0 +1,159 @@
...
+ * @fileoverview Directive for the privacy agreement modal.
+ */
+
+tie.directive('privacyModal', [function() {
+  return {
+    restrict: 'E',
+    scope: {
+      isDisplayed: '=',
+      title: '@',
+      isClientVersion: '&'
```
seanlip: Can we get rid of this binding and just depend on [ServerHandlerService](https://github.com/google/tie/blob/master/client/question/services/ServerHandlerService.js)? (Same thing in the learner view.)

client/question/components/PrivacyModalDirective.js
```diff
@@ -0,0 +1,159 @@
...
+              using our tool, you may generate computer code. Metadata about the code,
+              such as how long it took to write the code and whether it performs certain
+              tasks or meet certain criteria may also be created. Generated metadata and
+              computer code (collectively “User Generated Code”) will be collected and
+              analyzed as part of product improvement. By using TIE, you grant Google a
+              worldwide license to use, host, store, and create derivative works of
+              your User Generated code.
+            </p>
+          <div class="tie-privacy-modal-button-container">
+          <button ng-show="isClientVersion()" class="tie-button-blue tie-privacy-modal-button" ng-click="closeModal()">
```
seanlip: Indentation wrong. Should be indented by 2 from previous line.

client/question/components/PrivacyModalDirective.js
```diff
@@ -0,0 +1,159 @@
...
+        // Need to create this to be able to use ng-model, see below
+        // https://stackoverflow.com/questions/12618342/ng-model-does-not-update-controller-value/22768720#22768720
+        $scope.modalData = {};
+
+        /**
+         * Stores the user's acceptance of the privacy policy
+         * and closes the modal.
+         */
+        $scope.accept = function() {
+          // If this is not a shared computer
```
seanlip: Don't need this comment; it just repeats the next line of code in words. Comments are better used to explain _why_ something is the case.

client/question/components/PrivacyModalDirective.js
```diff
@@ -0,0 +1,159 @@
...
+          if (!$scope.modalData.isUsingSharedComputer) {
+            var expiryDate = new Date();
+            expiryDate.setDate(
+              expiryDate.getDate() + PRIVACY_COOKIE_LIFETIME);
+            $cookies.put(PRIVACY_COOKIE, true, {expires: expiryDate});
+          // If this is a shared computer
```
seanlip: Ditto re comment redundancy.

client/question/components/PrivacyModalDirective.js
```diff
@@ -0,0 +1,159 @@
...
+
+        /**
+         * Stores the user's acceptance of the privacy policy
+         * and closes the modal.
+         */
+        $scope.accept = function() {
+          // If this is not a shared computer
+          if (!$scope.modalData.isUsingSharedComputer) {
+            var expiryDate = new Date();
+            expiryDate.setDate(
```
seanlip: Can we just directly compute the date and set it?

client/question/components/PrivacyModalDirective.js
```diff
@@ -0,0 +1,159 @@
...
+              expiryDate.getDate() + PRIVACY_COOKIE_LIFETIME);
+            $cookies.put(PRIVACY_COOKIE, true, {expires: expiryDate});
+          // If this is a shared computer
+          } else if ($scope.modalData.isUsingSharedComputer) {
```
seanlip: `else` suffices here; no need `else if`

client/question/question.js
```diff
@@ -569,3 +570,25 @@ tie.constant('DISPLAY_AUTOSAVE_TEXT_SECONDS', 1);
  * @constant
  */
 tie.constant('DEFAULT_QUESTION_ID', 'reverseWords');
+/**
+ * Default cookie lifetime for the privacy policy in days.
+ *
+ * @type {number}
+ * @constant
+ */
+var ONE_YEAR_IN_DAYS = 365;
```
seanlip: Change to 0.5 year -- cc @rabidbit

client/question/question.js
```diff
@@ -569,3 +570,25 @@ tie.constant('DISPLAY_AUTOSAVE_TEXT_SECONDS', 1);
...
+tie.constant('PRIVACY_COOKIE_LIFETIME', ONE_YEAR_IN_DAYS);
+/**
+ * Key to use to store the privacy policy agreement.
+ *
+ * @type {string}
+ * @constant
+ */
+tie.constant('PRIVACY_COOKIE', 'privacyPolicyAccepted');
```
seanlip: Please confirm adherence to cookie naming guidelines.

client/question/question.js
```diff
@@ -569,3 +570,25 @@ tie.constant('DISPLAY_AUTOSAVE_TEXT_SECONDS', 1);
...
+/**
+ * Menu page url relative to the question page.
+ *
+ * @type {string}
+ * @constant
+ */
+tie.constant('MENU_PAGE_URL_FROM_QUESTION', '../menu/menu.html');
```
seanlip: MENU_PAGE_URL_FROM_QUESTION_PAGE

seanlip:
P.S. Please also merge master into this branch when you get a chance -- there was an issue with Travis that should be fixed by doing so. Thanks!

client/question/components/PrivacyModalDirective.js
seanlip: Actually, I think I was mistaken here -- what you have is fine. I was misremembering the JS Date API. Sorry!

client/question/components/LearnerViewDirective.js
```diff
@@ -983,6 +985,15 @@ tie.directive('learnerView', [function() {
         };
 
         $scope.initQuestionSet(questionSetId);
+
+        // If server version, and the user has not accepted the privacy policy,
+        // show them the privacy modal.
+        if (SERVER_URL) {
+          var privacyPolicyAccepted = CookieStorageService.getPrivacyCookie();
```
seanlip: Do we need to check expiry, or is this manifested by the disappearance of the cookie? (You can test using a cookie of short duration to see what happens, but it might be prudent to just check non-expiry anyway.)

client/question/components/PrivacyModalDirective.js
```diff
@@ -0,0 +1,154 @@
...
+
+        $scope.isClientVersion = !ServerHandlerService.doesServerExist();
+
+        /**
+         * Stores the user's acceptance of the privacy policy
+         * and closes the modal.
+         */
+        $scope.accept = function() {
+          if ($scope.modalData.isUsingSharedComputer) {
+            CookieStorageService.setPrivacyCookieToBeTrue();
```
seanlip: Not sure what "set to be true" means --setPrivacyCookieWithExpiryDate() is also set with true. Perhaps just setTransientPrivacyCookie() or similar?

client/question/components/PrivacyModalDirective.js
```diff
@@ -0,0 +1,154 @@
...
+          }
+          $scope.closeModal();
+        };
+
+        /**
+         * Stores the user's rejection of the privacy policy
+         * and redirects to the menu page.
+         */
+        $scope.reject = function() {
+          CookieStorageService.setPrivacyCookieToBeFalse();
```
seanlip: Hm, in this case, why set a cookie at all? I can understand deleting the existing cookie if one is present, but if we're going to take the cookie's existence as indicative of consent then we shouldn't set it in this case, right?

client/question/question.js
```diff
@@ -569,3 +570,25 @@ tie.constant('DISPLAY_AUTOSAVE_TEXT_SECONDS', 1);
...
  */
 tie.constant('DEFAULT_QUESTION_ID', 'reverseWords');
+/**
+ * Default cookie lifetime for the privacy policy in days.
+ *
+ * @type {number}
+ * @constant
+ */
+var SIX_MONTHS_IN_DAYS = 180;
+tie.constant('PRIVACY_COOKIE_LIFETIME', SIX_MONTHS_IN_DAYS);
```
seanlip: PRIVACY_COOKIE_LIFETIME_DAYS (include units)

client/question/services/CookieStorageService.js
```diff
@@ -0,0 +1,55 @@
...
+       * expiration ends when the browser session ends.
+       */
+      setPrivacyCookieToBeFalse: function() {
+        $cookies.put(PRIVACY_COOKIE_NAME, 'false');
+      },
+      /**
+       * Retrieves the privacy policy cookie value.
+       */
+      getPrivacyCookie: function() {
+        return ($cookies.get(PRIVACY_COOKIE_NAME) === 'true');
```
seanlip: Should we check non-expiry?

Maybe we should also call this hasPrivacyCookie(), since it returns a boolean?

client/question/question.js
theresa-lee: Done.

client/question/components/LearnerViewDirective.js
theresa-lee: Well I think the cookie should expire and return undefined (which getPrivacyCookie will return as false), and I can test on the server soon. (Locally cookies don't work).

client/question/services/CookieStorageService.js
theresa-lee: Oops, missed this. Changed to hasPrivacyCookie, and are you asking if we should check the expiration date of the cookie? I'm not quite sure if we actually have access to that. I think you can only get the value/object that you stored.

seanlip: Yeah, I was. But I don't think there's a way to check that, so please ignore that comment.

theresa-lee:
Okay, the cookies are working for both expiring and session, tested using locally deployed server.

The only caveat is that when the user has the option for their browser to remember their tabs, then it doesn't count as a session when they close out the window. So I think if someone checks that option and the browser has that option, then the cookie will live on?

seanlip:
Great, thanks @theresa-lee! Do you want me to review this as the PR for submission, or do you have further changes coming?

Also, what do you mean that "it doesn't count as a session when they close out the window"?

theresa-lee:
@seanlip You can review this for submission! 

Well, a session is finished usually by when the user closes out their window. But if that option is selected, the browser does not count the closing of the window as the end of a session because it is remembering where the user left off and still has the cookies stored.

seanlip:
Looks good, thanks! Will wait for @rabidbit's approval prior to merge.

PR #408, opened 2017-09-26
Add EventHandler Service.

eyurko:
Also update PythonCodeRunnerService and have some extra commits that still lingered despite me changing to a new branch.

 Annoying!

### DIFF
--- a/client/question/services/EventHandlerService.js
+++ b/client/question/services/EventHandlerService.js
@@ -1,0 +1,137 @@
+// Copyright 2017 The TIE Authors. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS-IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+/**
+ * @fileoverview Service that provides methods to create events on TIE's
+ * backend, if it exists.
+ */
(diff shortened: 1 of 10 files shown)


client/question/services/EventHandlerService.js
```diff
@@ -0,0 +1,137 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+/**
+ * @fileoverview Service that provides methods to create events on TIE's
+ * backend, if it exists.
+ */
+tie.factory('EventHandlerService', [
+  '$http', 'ServerHandlerService', function($http, ServerHandlerService) {
+    var FEEDBACK_CATEGORIES = [
```
seanlip: Declare externally as tie.constant, and then import it?

client/question/services/EventHandlerService.js
```diff
@@ -0,0 +1,137 @@
...
+      'SYNTAX', 'MATCHED BUGGY OUTPUT', 'UNMATCHED BUGGY OUTPUT', 'RUNTIME',
+      'PERFORMANCE', 'NO ERROR'];
+
+    return {
+      FEEDBACK_CATEGORIES: FEEDBACK_CATEGORIES,
+      /**
+       * Submits data to TIE's backend to create a SessionPauseEvent.
+       *
```
seanlip: Document args? Ditto elsewhere.

client/question/services/EventHandlerService.js
```diff
@@ -0,0 +1,137 @@
...
+       */
+      createSessionPauseEvent: function(sessionId) {
+        if (ServerHandlerService.doesServerExist()) {
+          var data = {
+            sessionId: sessionId
+          };
+          $http.post('/ajax/admin/create_session_pause_event', data);
```
seanlip: Should be /ajax/event/create_session_pause_event, right? Ditto elsewhere.

eyurko: It's admin on the backend, which is probably incorrect, so I'll make a CL to fix that as well.

PR #409, opened 2017-09-27
Introduce test suites to the question schema.

seanlip:
This PR introduces a testSuites property to the question data schema, replacing the old correctnessTests property.

Sorry it's a bit large -- updating the 16 question data files comprises probably about half of the PR.


client/question/domain/CodeEvalResultObjectFactory.js
```diff
@@ -134,8 +136,8 @@ tie.factory('CodeEvalResultObjectFactory', [
 
     /**
      * Returns the results for the last task that is run
-     * from the correctnessTests. The function should return the last array in
-     * _correctnessTestResults.
+     * from the correctnessTests. The function should return the last subarray
+     * in _correctnessTestResults.
```
eyurko: should this become _testCaseResults?

eyurko: ditto for others

client/question/domain/CodeEvalResultObjectFactorySpec.js
```diff
@@ -21,7 +21,7 @@ describe('CodeEvalResultObjectFactory', function() {
   var codeEvalResult;
   var CODE = 'code';
   var OUTPUT = '';
-  var CORRECTNESS_TEST_RESULTS = [[true, true], [false, false]];
+  var CORRECTNESS_TEST_RESULTS = [[[true, true]], [[false, false]]];
```
eyurko: same here?

seanlip:
Thanks for the review!

I want to do a follow-up change to make it observedOutputs or something like that, since the outputs are not restricted to true/false. But that would have made the PR even bigger, so I'd like to do it separately. Does that sound OK?

seanlip:
Discussed with @eyurko; decided to make follow-up changes in separate PR. Merging this one.

PR #410, opened 2017-09-28
Added ability to extract all feedback paragraphs as plain text.

eyurko:
i need this


client/question/domain/FeedbackObjectFactory.js
```diff
@@ -186,6 +186,24 @@ tie.factory('FeedbackObjectFactory', [
...
 
+
+    /**
+     * Returns all of the feedback provided as a JSON dict, preserving type
+     * and content information.
+     *
+     * @returns [{string, string}] The feedback paragraphs list of dicts.
+     */
+    Feedback.prototype.getParagraphsAsListOfDicts = function() {
+      var feedbackParagraphs = [];
```
seanlip: (Optional) A bit cleaner would be:

    return this._paragraphs.map(function(paragraph) {
      return paragraph.toDict();
    });

toDict() should be defined on the paragraph object prototype, btw, rather than on the class. Might be worth changing it if it's not the case.

client/question/domain/FeedbackObjectFactorySpec.js
```diff
@@ -69,6 +76,22 @@ describe('FeedbackObjectFactory', function() {
     );
   });
 
+  describe('getParagraphsAsListOfDicts', function() {
+    it('should return an array of paragraphs as dictionaries', function() {
+      feedback.appendTextParagraph("This");
```
seanlip: Please use single quotes here and below.

client/question/domain/FeedbackObjectFactory.js
```diff
@@ -186,6 +186,19 @@ tie.factory('FeedbackObjectFactory', [
       }
     };
 
+
+    /**
+     * Returns all of the feedback provided as a JSON dict, preserving type
+     * and content information.
+     *
+     * @returns [{string, string}] The feedback paragraphs list of dicts.
```
seanlip: Should the second arg of the type be Object?

eyurko: Whoops, yup!

PR #417, opened 2017-10-03
Add event handlers to TIE's frontend.

eyurko:


client/question/components/LearnerViewDirective.js
```diff
@@ -661,6 +661,48 @@ tie.directive('learnerView', [function() {
         var questionWindowDiv =
             document.getElementsByClassName('tie-question-window')[0];
 
+        // Set the name of the hidden property and the change event for
+        // visibility
```
seanlip: Nit: add period at end.

client/question/components/LearnerViewDirective.js
```diff
@@ -661,6 +661,48 @@ tie.directive('learnerView', [function() {
...
+        var hidden;
```
seanlip: This could have a better name -- it's not clear what it represents, and at first I thought it was a boolean specifying whether the page was hidden or not. That said, I think it would be clearer to get rid of this and just have an isDocumentHidden() function, and handle the browser-specific part there.

client/question/components/LearnerViewDirective.js
```diff
@@ -661,6 +661,48 @@ tie.directive('learnerView', [function() {
...
+        } else if (typeof document.webkitHidden !== 'undefined') {
+          hidden = 'webkitHidden';
+          visibilityChange = 'webkitvisibilitychange';
+        }
+
+        // Warn if the browser doesn't support addEventListener
+        // or the Page Visibility API
+        if (typeof document.addEventListener === 'undefined' ||
+          typeof document[hidden] === 'undefined') {
+          alert('Action not supported by current browser.');
```
seanlip: The user is going to have no idea what's going on when this pops up. We need a more graceful fallback, and we also need to understand the consequences it'll have on our statistics if we can't record certain events -- maybe just don't record pageleave in that case.

client/question/components/LearnerViewDirective.js
```diff
@@ -906,6 +964,9 @@ tie.directive('learnerView', [function() {
         $scope.resetCode = function() {
           LocalStorageService.clearLocalStorageCode(
             $scope.currentQuestionId, language);
+          EventHandlerService.createCodeResetEvent(
+            SessionIdService.getSessionId());
+          // Do we want to trigger another QuestionStartEvent?
```
seanlip: Yes, I guess? But that's already being done in loadQuestion right?

client/question/services/SolutionHandlerServiceSpec.js
```diff
@@ -134,11 +137,14 @@ describe('SolutionHandlerService', function() {
           '    return False'
         ].join('\n');
 
+        spyOn(EventHandlerService, 'createCodeSubmitEvent');
```
seanlip: I don't think you need to add this to every single case -- can't we just have one Karma test dedicated to checking that events are handled correctly?

client/question/components/LearnerViewDirective.js
eyurko: I think we just won't record them because the EventListener won't be added.

client/question/components/LearnerViewDirective.js
eyurko: Fair enough.

client/question/services/SolutionHandlerServiceSpec.js
eyurko: Got it.

client/question/components/LearnerViewDirective.js
eyurko: Fixed! Called it "determineHiddenAttributeNameForBrowser".

client/question/components/LearnerViewDirective.js
eyurko: Aight

client/question/components/LearnerViewDirective.js
```diff
@@ -661,6 +661,63 @@ tie.directive('learnerView', [function() {
...
+            return 'msHidden';
+          } else if (typeof document.webkitHidden !== 'undefined') {
+            return 'webkitHidden';
+          }
+          return null;
+        };
+
+        var setEventListenerForVisibilityChange = function() {
+          var hiddenAttributeName = determineHiddenAttributeNameForBrowser();
+          var visibilityChange;
```
seanlip: Nit: define this on line 705, so that its scope is limited. The rest of the function does not need to know about it.

client/question/components/LearnerViewDirective.js
```diff
@@ -661,6 +661,63 @@ tie.directive('learnerView', [function() {
...
+          if (typeof document.addEventListener === 'undefined' ||
+            hiddenAttributeName === null ||
+            typeof document[hiddenAttributeName] === 'undefined') {
```
seanlip: Why would this trigger? Wouldn't it be caught by the previous line?

client/question/components/LearnerViewDirective.js
```diff
@@ -701,6 +759,12 @@ tie.directive('learnerView', [function() {
           $scope.nextButtonIsShown = false;
           var feedback = FeedbackObjectFactory.create();
           $scope.greetingParagraphs = feedback.getParagraphs();
+          EventHandlerService.createQuestionStartEvent(
+            SessionIdService.getSessionId(), $scope.currentQuestionId,
+            'QUESTION_VERSION');
```
seanlip: No -- declare QuestionDataService.getQuestionVersion(questionId) and return null from there for now (if we're working client-side), but add a TODO to it to make it actually return the correct question version once we surface this information from the server. Ditto elsewhere. Want to defend against any chance of these (invalid) strings getting stored.

client/question/components/LearnerViewDirective.js
```diff
@@ -734,25 +798,41 @@ tie.directive('learnerView', [function() {
...
          * to the feedback passed into the function.
          *
          * @param {Feedback} feedback
+         * @param {string} code
          */
-        var setFeedback = function(feedback) {
+        $scope.setFeedback = function(feedback, code) {
+          EventHandlerService.createCodeSubmitEvent(
+            SessionIdService.getSessionId(),
+            feedback.getParagraphsAsListOfDicts(), 'ERROR CATEGORY',
```
seanlip: For now, omit the 'ERROR CATEGORY' argument altogether. We'll add it back later if needed. (Same reason as above.)

client/question/components/LearnerViewDirective.js
eyurko: k

client/question/components/LearnerViewDirective.js
eyurko: True; I'll drop this.

client/question/components/LearnerViewDirective.js
eyurko: why not just create feedback.getErrorCategory() and return null, for consistency with the solution above?

I'd rather do that than rearchitect EventHandlerService twice.

client/question/components/LearnerViewDirective.js
eyurko: Why not QuestionObjectFactory.getQuestionVersion? Shouldn't that be stored on the object, rather than the service?

eyurko: SGTM.

eyurko: Had to make a version with no args since I couldn't avoid unused variables.

client/question/components/LearnerViewDirective.js
seanlip: That's fine too, yes.

client/data/services/QuestionDataService.js
```diff
@@ -76,6 +76,16 @@ tieData.factory('QuestionDataService', [
         return question.getTitle();
       },
       /**
+       * Returns the question's version.
+       * Currently returns null, as question versioning isn't implemented yet.
+       * TODO(eyurko): Return correct question version, once implemented.
+       *
+       * @returns {null}
+       */
+      getQuestionVersion() {
```
seanlip: Btw, I just noticed the style for this isn't quite right -- we typically do: `functionName: function(args) {`.  It looks like a couple of the functions above/below this one also need this small fix; would you mind doing that too in this PR?

eyurko: sure

client/question/domain/FeedbackObjectFactory.js
```diff
@@ -148,6 +156,26 @@ tie.factory('FeedbackObjectFactory', [
     };
 
     /**
+     * A getter for the _errorCategory property.
+     * This function should return a number that indicates the index of the
```
seanlip: ?

eyurko:
Yeah PTAL -- I added some tests.

200 lines of tests, to be precise.

client/question/components/LearnerViewDirectiveSpec.js
```diff
@@ -219,6 +219,146 @@ describe('LearnerViewDirective', function() {
     });
   });
 
+  describe("isDocumentHidden", function() {
```
seanlip: Nit: single quotes throughout please

client/question/components/LearnerViewDirectiveSpec.js
```diff
@@ -219,6 +219,146 @@ describe('LearnerViewDirective', function() {
...
+    it('should return document[hiddenAttributeName]', function() {
+      expect($scope.isDocumentHidden('sure')).toEqual(document.sure);
```
seanlip: Maybe just use 'hidden' or 'attrname'? Hard to tell from context what 'sure' represents.

client/question/components/LearnerViewDirective.js
```diff
@@ -661,61 +661,93 @@ tie.directive('learnerView', [function() {
...
             return 'webkitHidden';
           }
           return null;
         };
 
-        var setEventListenerForVisibilityChange = function() {
-          var hiddenAttributeName = determineHiddenAttributeNameForBrowser();
+        $scope.determineVisibilityChangeAttributeNameForBrowser = function() {
+          // Handle page visibility change
+          if (typeof $scope.getHiddenAttribute() !== 'undefined') {
```
seanlip: There's duplication here and above -- probably cleaner to have a function that gets the browser type and then just read off the variable name (or in fact the attr prefix) from an object mapping browser type to attr prefix.

PR #423, opened 2017-10-06
Complete schema changes for suite-level tests and ignored suite IDs for buggy-output tests.

seanlip:
@eyurko PTAL

Note: Suite-level tests are now part of the task schema, but aren't integrated into the feedback flow yet. I'll do that in a separate follow-on PR after this gets merged, since the current one is getting long.


assets/questions/reverseWords.js
```diff
@@ -395,10 +427,33 @@ globalData.questions['reverseWords'] = {  // eslint-disable-line dot-notation
...
       messages: [
         'Note that numbers should not be treated as English words.'
       ]
     }],
+    suiteLevelTests: [{
+      testSuiteIdsThatMustPass: ['SAMPLE_INPUT', 'GENERAL_CASE', 'WHITESPACE'],
+      testSuiteIdsThatMustFail: ['CORNER_CASES'],
+      messages: [
+        [
+          'Are you happy with the code? It must give the correct answer in ',
```
eyurko: I don't think the first sentence is appropriate. Maybe "It seems like your code is failing on some test cases. Try stepping through some sample inputs on paper and see if your output is consistent with your expected output in all cases."

client/question/services/SolutionHandlerServiceSpec.js
```diff
@@ -368,5 +366,79 @@ describe('SolutionHandlerService', function() {
...
+            input: 'task_1_suite_2_test_1',
+            allowedOutputs: [false]
+          }, {
+            input: 'task_1_suite_2_test_2',
+            allowedOutputs: [false]
+          }]
+        }];
+      }));
+
+      it('should check all buggy output if nothing is ignored', function(done) {
```
eyurko: Nit: buggy outputs

eyurko: and others below, where needed.

seanlip: Done.

assets/questions/reverseWords.js
seanlip: Done. Thank you for the catch!

seanlip:
Thanks, @eyurko! All comments addressed. PTAL.

PR #424, opened 2017-10-11
Add event categories.

eyurko:
Ugh I keep forgetting to drop the commit history.

(Sorry)

Just ignore the LearnerViewDirective changes?

### DIFF
--- a/client/question/domain/FeedbackObjectFactorySpec.js
+++ b/client/question/domain/FeedbackObjectFactorySpec.js
@@ -40,6 +40,19 @@ describe('FeedbackObjectFactory', function() {
     });
   });
 
+  describe('getErrorCategory', function() {
+    it('should return null', function() {
+      expect(feedback.getErrorCategory()).toBe(null);
+    });
+  });
+
+  describe('setErrorCategory', function() {
+    it('should set the _errorCategory property', function() {
+      feedback.setErrorCategory('ERROR OPERATOR');
+      expect(feedback._errorCategory).toBe('ERROR OPERATOR');
+    });
+  });
+
   describe('firstParagraphText', function() {
     it('should throw an error when adding paragraph one as code', function() {
       var errorFunction = function() {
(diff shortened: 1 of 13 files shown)


client/question/services/FeedbackGeneratorService.js
```diff
@@ -512,6 +521,7 @@ tie.factory('FeedbackGeneratorService', [
           'You\'ve completed all the tasks for this question! Click the ',
           '"Next" button to move on to the next question.'
         ].join(''));
+        feedback.setErrorCategory('SUCCESSFUL');
```
seanlip: This reads oddly. Should we rename the method to feedback.setCategory()?

client/question/domain/FeedbackObjectFactorySpec.js
```diff
@@ -40,6 +40,19 @@ describe('FeedbackObjectFactory', function() {
...
 
+  describe('getErrorCategory', function() {
+    it('should return null', function() {
+      expect(feedback.getErrorCategory()).toBe(null);
+    });
+  });
+
+  describe('setErrorCategory', function() {
+    it('should set the _errorCategory property', function() {
+      feedback.setErrorCategory('ERROR OPERATOR');
```
seanlip: I think there should be a check on-set to ensure only categories we recognize are allowed, and that we should have a tie.constant('FEEDBACK_CATEGORIES', ...) whose value is an object that looks like:
```
   {
      SYNTAX_ERROR: 'SYNTAX_ERROR',
      ...
   }
```
so that this becomes

    feedback.setErrorCategory(FEEDBACK_CATEGORIES.ERROR_OPERATOR);

Otherwise we're repeating too many raw strings all over the codebase; it's easy for typos to come in.

seanlip: I've done this in #425 btw -- I needed this functionality.

seanlip:
Integrated in #425; closing.

PR #425, opened 2017-10-12
Add functionality for suite-level tests

seanlip:
This PR introduces functionality for suite-level tests. There was a bit of yak-shaving needed since there was significant overlap with buggy-output tests, and I needed to get the type of feedback given in the previous run in order to figure out whether to display a next hint or not. This required introducing a FEEDBACK_CATEGORIES object, updating errorCategory to feedbackCategory, and updating the way the "print" keyword feedback is handled.

I noticed the following while working through the codebase, but will defer them to future PRs:
- storing an answerIsCorrect property is redundant, since this is now indicated by FEEDBACK_CATEGORIES.SUCCESSFUL
- the feedbackCategory variable in EventHandlerService.createCodeSubmitEvent is left unchanged for now, because a similar change needs to happen on the backend
- correctnessTestResults needs to be renamed to observedOutputs, with consequent changes on the backend.

@eyurko PTAL


eyurko:
> the feedbackCategory variable in EventHandlerService.createCodeSubmitEvent is left unchanged for now, because a similar change needs to happen on the backend

You should just make the frontend change now and fix it on the backend when you sync -- pushing it to a future PR means that it could stall / block backend changes.

seanlip:
> You should just make the frontend change now and fix it on the backend when you sync -- pushing it to a future PR means that it could stall / block backend changes.

Thanks for the note -- done.

Yeah, good point about #424, I just had another look and I didn't realize this PR covered so much of it. Let me make another commit bringing in the extra stuff from that (e.g. question version) and then I'll close it.

seanlip:
OK, everything from #424 is added. PTAL. I'll close the other one.

Thanks!

client/question/domain/FeedbackObjectFactorySpec.js
```diff
@@ -23,33 +23,38 @@ describe('FeedbackObjectFactory', function() {
...
+          FEEDBACK_CATEGORIES.INVALID_CATEGORY, false);
+      }).toThrowError('Invalid feedback category: undefined');
     });
   });
 
-  describe('getErrorCategory', function() {
-    it('should return null', function() {
-      expect(feedback.getErrorCategory()).toBe(null);
+  describe('isAnswerCorrect', function() {
+    it('should return whether or not answer is correct', function() {
```
eyurko: whether or not the* answer is correct

client/question/questionSpec.js
```diff
@@ -63,4 +65,12 @@ describe('question.js', function() {
       });
     });
   });
+
+  describe('FEEDBACK_CATEGORIES', function() {
+    it('should have keys being equal to values', function() {
```
eyurko: should have identical keys and values

client/question/domain/FeedbackObjectFactorySpec.js
seanlip: Done.

client/question/questionSpec.js
seanlip: Done.

seanlip:
Thanks! Resolved comments; will merge once checks pass.

PR #426, opened 2017-10-12
Fix #412: just import everything at the outset.

### LINKED ISSUE #412: User study - Library import check appears broken
When users try to import a library, they get an "undeclared variable" error message. In one case, it appears there was no reaction to the import statement.

seanlip:
Hi @rabidbit -- I took a quick look at this. Turns out what was happening was that the imports were being placed at class-level:

    class StudentCode(object):
        import X
        import Y

        def run():
             # student's code goes here

and for some reason Skulpt wasn't finding them. On investigation, just importing everything that's supported at the top of the file seems to be the simplest way to handle this, so I went ahead and modified the starter code instead so that collections, math, etc. are automatically available.

The one thing I'm not sure about is whether we want to put a comment at the top of the coding area to say something like this:

```
    # This is a Python environment that you can type code into. The following
    # libraries have already been pre-imported and can be used if you need them:
    # collections, copy, image, math, operator, ...
```

but I'm not sure. On the one hand it gives info that might be useful. On the other hand I don't want to create the suggestion that the user should be using something from here, and it's not something an interviewer would say. My current leaning is to not have it for now, but add it later if we detect confusion (hopefully the PR as-is will address most of the issues since there won't be random import errors showing up). Wdyt?

### DIFF
--- a/client/question/question.js
+++ b/client/question/question.js
@@ -113,7 +113,14 @@ tie.constant('FEEDBACK_CATEGORIES', [
  */
 tie.constant('SYSTEM_CODE', {
   python: [
+    'import collections',
     'import copy',
+    'import image',
+    'import math',
+    'import operator',
+    'import random',
+    'import re',
+    'import string',
     'import time',
     '',
     '# A copy of the most-recently processed input item. This is useful',


rabidbit:
I agree that we should not include in starter code, as it's likely users will try to use them even if they don't need them, resulting in distraction and frustration.

Can we detect if users try to import something? If we detect something being imported, we could say something like this in the feedback:

> Looks like you are trying to import libraries. TIE currently supports and auto-imports the following libraries: collections, copy, image, math, and operator. If you think you need to use any of these, you can simply call the method you require.

seanlip:
I think we already do that, but only for unsupported libraries -- see the feedback when you type "import x", for example. For supported libraries, nothing happens, and I think including the note in that case would be a distraction.

So, I think the current behaviour is fine as-is, but let me know if you want me to make further changes?

rabidbit:
Sorry, I should have been more specific in #412 . Users are typing import statements above the template code, like this:

```
import someLibrary

def someCode(x):
    stuff = someLibrary.doStuff()
    return stuff
```

Which results in the following error:

```
It looks like someLibrary isn't a declared variable.
Did you make sure to spell it correctly?
And is it correctly initialized?
```

Which is the same error they get if they didn't have an import statement. That's the issue for #412 

If users can enter import statements above the template code, then I think the existing flow and feedback would work.

seanlip:
Right -- I believe I had tested that case in my PR and it works as expected. Please could you verify?

rabidbit:
I didn't realize that this occurred only for supported but not imported modules. I think I understand the issue and fix. This appears to fix the bug.

seanlip:
Merged. Thanks!

PR #427, opened 2017-10-13
Remove explicit answerIsCorrect variable, since this info is already captured in the feedback category.

seanlip:


client/question/services/EventHandlerService.js
```diff
@@ -153,7 +153,8 @@ tie.factory('EventHandlerService', [
        * @param {string} feedbackCategory The type of feedback shown to the
        *   user.
        * @param {string} code The user's submitted code.
-       * @param {boolean} success Whether or not the question was completed.
+       * @param {boolean} success Whether or not the entire question was
```
seanlip: @eyurko just to confirm, this is indeed supposed to be question, not task, right? It was actually sending task information before, so I just wanted to double-check what you had in mind.

eyurko: I thought it was supposed to be task information, but I don't really have a strong preference. Either way, we can just pull the last CodeSubmitEvent before a TaskCompleteEvent, so it's likely not _explicitly_ necessary.

seanlip: Hm, OK. I think question would be better than task, for two reasons:

- I don't think we are passing any info on which _task_ this corresponds to.
- If we wanted task completion, we can get it from the "successful" feedback category (but see the previous bullet point; we still don't know which task it corresponds to).

Alternatively, maybe we drop it altogether and rely on the other events (question start, question complete, task start, task complete) for progress.

I think I lean towards dropping it altogether so we don't have duplicate info. What do you think?

eyurko:
I'm fine with that, though we should still have feedback category for
determining success

seanlip:
Changes done, PTAL

client/question/services/EventHandlerService.js
eyurko: SGTM

Issue #413, opened 2017-10-03
User study - Not operator regex buggy


rabidbit:
Faulty regex for "Are you making sure to use the right NOT operator?"; specifically, it catches on `!"` in this code:
```
def check(char):
	if char == "!":
```
Also, awkward wording in "Are you making sure to use the right NOT operator?"

seanlip:
@rabidbit -- what do you want the wording here to be?

rabidbit:
"Are you using the right NOT operator?"

seanlip:
Fixed in #431.

PR #429, opened 2017-10-18
Fix #411 and #413: guard against false positives due to problematic keywords in comments and strings.

### LINKED ISSUE #411: User study - Ignore comments
TIE parses comments. E.g., if someone types the word 'print' in a comment like so:
```
# printing is not allowed
# print my_answer
```
TIE notifies user that they cannot print.

seanlip:


seanlip:
Also /cc @eyurko

rabidbit:
IIUC, this captures single line comments but not multi-line comments? From a UX standpoint, I think that's OK. Ideally, would prefer to ignore multi-line as well (e.g., user decides to comment out a whole block of code rather than delete it in order to have it as a reference); would this be challenging to do?

seanlip:
In Python, commenting out multiple lines is done by putting # at the start of each line -- unless you mean turning the comment into a string?

Handling multiline strings is indeed challenging to do, I'm afraid. I made a start on it but decided not to pursue it further at the current time because it is quite complicated to get right, and because I didn't think this case was likely to pop up in our studies. It would be easier if we had a backend parse tree to work with, so we might support that server-side in the future. But I think building our own parser from scratch in the client is not the way to go.

rabidbit:
I was going off the first google hit on "multiline comment python"... https://stackoverflow.com/questions/7696924/way-to-create-multiline-comments-in-python

If it's too technically challenging, then the single line comment (or the technical Python comment) sounds fine with me.

seanlip:
Ah, yes, that is indeed a string. Yes, it will require significant time to do this in full generality.

If you're happy with this, do you want to approve?

PR #430, opened 2017-10-18
Added Python Primer Button 

### LINKED ISSUE #415: User study - There is no link to the Python primer doc
No link to python primer doc (for client-version only)

theresa-lee:
Fixes #415

### DIFF
--- a/client/question/components/LearnerViewDirective.js
+++ b/client/question/components/LearnerViewDirective.js
@@ -109,7 +109,7 @@ tie.directive('learnerView', [function() {
                   <option value="Python" selected>Python</option>
                 </select>
                 <button ng-if="!SERVER_URL" class="tie-python-primer tie-button" ng-class="{'night-mode': isInDarkMode}">
-                  <a class="tie-primer-link" ng-class="{'night-mode': isInDarkMode}" target="_blank" ng-href="../docs/py-primer-{{isInDarkMode === true ? 'dark' : 'light'}}.html">New to python?</a>
+                  <a class="tie-primer-link" ng-class="{'night-mode': isInDarkMode}" target="_blank" ng-href="{{getPythonPrimerUrl()}}">New to python?</a>
                 </button>
                 <button class="tie-code-reset tie-button protractor-test-reset-code-btn" name="code-reset"
                     ng-class="{'night-mode': isInDarkMode}"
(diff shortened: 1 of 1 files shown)


client/question/components/LearnerViewDirective.js
```diff
@@ -105,9 +105,12 @@ tie.directive('learnerView', [function() {
                     </ui-codemirror>
                   </div>
                 </div>
-                <select class="tie-select-menu" name="lang-select-menu" ng-class="{'night-mode': isInDarkMode}">
+                <select ng-if="SERVER_URL" class="tie-select-menu" name="lang-select-menu" ng-class="{'night-mode': isInDarkMode}">
                   <option value="Python" selected>Python</option>
                 </select>
+                <button ng-if="!SERVER_URL" class="tie-python-primer tie-button" ng-class="{'night-mode': isInDarkMode}">
+                  <a class="tie-primer-link" ng-class="{'night-mode': isInDarkMode}" target="_blank" ng-href="../docs/py-primer-{{isInDarkMode === true ? 'dark' : 'light'}}.html">New to python?</a>
```
seanlip: Could you extract this into $scope.getPythonPrimerUrl()?

Also, note that you don't need the `=== true` bit. `isInDarkMode ? 'dark' : 'light'` should suffice.

rabidbit:
I think a link (in the same location and font size) might be better than a button. It will likely see less use that the other buttons; so, I'm thinking a link (using standard web link color) would look softer / less glaring. Also, a link points to a doc / resource as opposed to performing some action on the application or UI.

theresa-lee:
Yeah, I was thinking how the language selector would actually be a button dropdown too so I was trying to keep it consistent with that. The functionality is all here though, I don't mind it being a link. I wanted to submit this before I left so whether you want to change the CSS now on this PR or in another PR after this one is submitted is up to you?

client/question/components/LearnerViewDirective.js
```diff
@@ -916,6 +927,14 @@ tie.directive('learnerView', [function() {
         };
 
         /**
+         * Provides the URL to the appropriately themed python primer file.
+         */
+        $scope.getPythonPrimerUrl = function() {
+          var primerTheme = $scope.isInDarkMode ? 'dark' : 'light';
+          return "../docs/py-primer-" + primerTheme + ".html";
```
seanlip: Nit: use single quotes please.

theresa-lee: Done!

seanlip:
Thanks @theresa-lee! @rabidbit please feel free to merge if you're happy with this.

rabidbit:
We can do the CSS after this PR.

PR #431, opened 2017-10-19
Fix #416 and part of #413: small textual changes.

### LINKED ISSUE #416: User study - Improve "Try reading the question again" feedback
It may be condescending to some users since it implies that they didn’t carefully read it the first time around (feedback tone). Soften the language (e.g., explain that we think they may have misinterpreted the question) and/or point them back example input/output.

seanlip:


assets/questions/reverseWords.js
```diff
@@ -126,8 +126,9 @@ globalData.questions['reverseWords'] = {  // eslint-disable-line dot-notation
         'For this question, you\'ll implement the reverseWords function. ',
         'This function takes a string of words and reverses the individual ',
         'words, but it does not change the ordering of the words within the ',
-        'sentence. All the original whitespace/punctuation should be ',
-        'preserved. Here\'s an example:'
+        'sentence. A word is defined to consist of contiguous letters, and ',
+        'all other characters in the string (such as whitespace and ',
+        'punctuation) should be preserved. Here\'s an example:'
```
rabidbit: Wording appears a bit awkward to me. How about something like...

A word consists of contiguous letters. Characters that are not letters, such as whitespace and punctuation, should be preserved (not moved or reversed).

seanlip: Thanks, I've updated it -- PTAL. Just FYI, I didn't use this text exactly because it doesn't really make sense to me to "reverse" a character.

assets/questions/reverseWords.js
```diff
@@ -303,8 +304,10 @@ globalData.questions['reverseWords'] = {  // eslint-disable-line dot-notation
       ignoredTestSuiteIds: [],
       messages: [
         [
-          "Try reading the question again, and looking at the sample ",
-          "input/output -- it's asking for something different."
+          "I think you might have misinterpreted the question. Try verifying ",
```
rabidbit: I recommend removing "I think". I don't think it's necessary to use pronouns to refer to the app (I think we should rely on the intelligence of the feedback to make it appear more human-like).

rabidbit:
One more minor comment. Otherwise, LGTM.

assets/questions/reverseWords.js
seanlip: Done, thanks -- I was actually trying to make the feedback "softer", but I'm fine either way and have removed it for now.

PR #433, opened 2017-10-24
Add a language-level check for true/false casing.

seanlip:


client/question/question.js
```diff
@@ -334,6 +334,20 @@ tie.constant('WRONG_LANGUAGE_ERRORS', {
...
   }, {
+    // Used the wrong casing for "True" or "False".
+    errorName: 'booleans',
+    regExString: '\\b(true|TRUE|false|FALSE)\\b',
+    feedbackParagraphs: [
+      {
+        type: 'text',
+        content: [
+          "Note that, in Python, we write boolean values as True and False. ",
+          "Triple-check to make sure you've got the right casing!"
```
rabidbit: "Double-check to make sure 'True' or 'False' are written with the letters in the same case."

rabidbit: Basically, please add a little context around "casing" or "case" to make sure people don't get thrown off.

seanlip: Good catch, thanks! Done.

client/question/question.js
```diff
@@ -334,6 +334,20 @@ tie.constant('WRONG_LANGUAGE_ERRORS', {
...
     allowMultiline: false
   }, {
+    // Used the wrong casing for "True" or "False".
+    errorName: 'booleans',
+    regExString: '\\b(true|TRUE|false|FALSE)\\b',
+    feedbackParagraphs: [
+      {
+        type: 'text',
+        content: [
+          "Note that, in Python, we write boolean values as True and False. ",
```
eyurko: "Python uses True and False for boolean values, rather than true and false."

seanlip: Actually, I can't quite write that, since they may have written TRUE/FALSE instead, and it seems a bit clumsy to list all the possibilities. I'd also rather not break this out into more separate tests than necessary.

Do you feel strongly / have alternative wording?

eyurko: I'd just go with "Python only allows True and False for boolean values." or something.

seanlip: Done.

client/question/question.js
```diff
@@ -341,9 +341,9 @@ tie.constant('WRONG_LANGUAGE_ERRORS', {
       {
         type: 'text',
         content: [
-          "Note that, in Python, we write boolean values as True and False. ",
-          "Double-check to make sure 'True' or 'False' are written with the ",
-          "letters in the correct case."
+          "Python only allows True and False for boolean values. Double-check ",
+          "to make sure 'True' or 'False' are written with the letters in the ",
```
eyurko: "with the letters in the correct case" seems a bit stilted to me, but I can't think of anything better. If you can, go for it, otherwise, approved.

seanlip: The more streamlined alternatives I can think of suffer from a consequent reduction in clarity, so on balance I think I'll leave it as is.

PR #434, opened 2017-10-24
Add some more feedback to the 'reverse words' question.

seanlip:
There's a bit more to come here, but I realized while working on this that I might have done the buggy-output-tests refactor incorrectly (it might be better to whitelist a set of test suites to run rather than blacklist a set of test suites not to run). Still thinking through this, but figured I might as well get all the other changes in first.

### DIFF
--- a/assets/questions/reverseWords.js
+++ b/assets/questions/reverseWords.js
@@ -452,14 +452,9 @@ globalData.questions['reverseWords'] = {  // eslint-disable-line dot-notation
       testSuiteIdsThatMustFail: ['SAMPLE_INPUT'],
       messages: [
         [
-          "Hmm, that code doesn't seem to work. Try taking a look at the ",
-          "sample inputs and outputs, then walk through your code to see what ",
-          "might be wrong."
-        ].join(''),
-        [
-          "Have you tried stepping through your code using the sample input ",
-          "and expected output above? It might help you find where the ",
-          "problem is."
+          "That code doesn't seem to work. Try taking a look at the sample ",
+          "inputs and outputs, then walk through your code to see what might ",
+          "be wrong."
         ].join('')
       ]
     }, {
--- a/assets/questions/reverseWords.js
+++ b/assets/questions/reverseWords.js
@@ -164,14 +164,20 @@ globalData.questions['reverseWords'] = {  // eslint-disable-line dot-notation
       id: 'WHITESPACE',
       humanReadableName: 'strings with lots of whitespace',
       testCases: [{
+        input: 'hello    ',
+        allowedOutputs: ['olleh    ']
+      }, {
+        input: '   hello',
+        allowedOutputs: ['   olleh']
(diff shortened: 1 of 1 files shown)


assets/questions/reverseWords.js
```diff
@@ -436,6 +448,52 @@ globalData.questions['reverseWords'] = {  // eslint-disable-line dot-notation
...
+      testSuiteIdsThatMustPass: [],
+      testSuiteIdsThatMustFail: ['SAMPLE_INPUT'],
+      messages: [
+        [
+          "Hmm, that code doesn't seem to work. Try taking a look at the ",
+          "sample inputs and outputs, then walk through your code to see what ",
+          "might be wrong."
+        ].join(''),
+        [
+          "Have you tried stepping through your code using the sample input ",
```
eyurko: Try stepping through your code using the sample input above. Note where and how it differs from the expected output.

assets/questions/reverseWords.js
```diff
@@ -436,6 +448,52 @@ globalData.questions['reverseWords'] = {  // eslint-disable-line dot-notation
...
+      testSuiteIdsThatMustFail: ['PUNCTUATION'],
+      messages: [
+        [
+          'Be careful about what constitues a "word". Punctuation should be ',
+          'left in place, and not reversed along with the words.'
+        ].join(''),
+        [
+          'You might have an incorrect test for what constitutes a "word ',
+          'separator". Can you think of a better way to ensure that only ',
+          'Roman letters are left in place?'
```
eyurko: Are these Roman letters? I guess so. Drop Roman, here -- made me think.

seanlip: Done.

assets/questions/reverseWords.js
seanlip: Hm -- for this one, the original text is from @rabidbit, and I prefer that version. I think, in the version you've proposed, it's not clear what "it" is.

I'm leaving this as-is for now. I'm generally happy to change it to anything, but this might be better as a discussion between you and @rabidbit.

eyurko: Sure, sounds good -- I think I generally am starting to prefer commands rather than rhetorical questions. I generally am leaning towards "clear and firm" rather than "interrogative", as it seems a bit more direct. The questions worry me that they might be interpreted as vaguely condescending.

Thoughts, @rabidbit? Not gonna let this block submit, though.

rabidbit: I agree that questions can come across as condescending (or frustrating, if the user believes they have done what is being asked).  How about just the following:

"That code doesn't seem to work. Try taking a look at the sample inputs and outputs, and walk through your code to see what might be wrong."

(removes "Hmm," on first feedback and drops the inquiring feedback)

seanlip: Thanks for the resolution! Done.

PR #436, opened 2017-10-25
Fix #435: Impose limits on width and height of question window.

### LINKED ISSUE #435: No minimum size for Feedback Window
Because there is no minimize size for the feedback window, the following can happen:

screen shot 2017-10-25 at 11 12 01 am

And even this:

screen shot 2017-10-25 at 11 11 34 am

seanlip:


rabidbit:
nit: alphabetical order?

seanlip:
Whoops, good catch. Fixed!

rabidbit:
Thanks for the fix!

PR #439, opened 2017-11-02
Fix error introduced in commit 200c144 when checking whether code has changed.

seanlip:
This fixes an error I introduced in an earlier PR (oops!) and adds a test to prevent it from happening again.


seanlip:
Closing and reopening because CLA checks seem to be stalled.

PR #440, opened 2017-11-04
Introduce the concept of tips, i.e. non-blocking feedback.

seanlip:
This is the first of a series of PRs dedicated to tips, i.e. non-blocking feedback that is shown above the regular feedback TIE provides.

This PR lays the groundwork for a tips framework, and converts the "you're using a print statement" warning into a system-generated tip. Subsequent PRs will add functionality for question-specific tips (previously known as "wrong-approach errors"), and introduce a graduated system for not showing tips after they have been ignored.

The PR also refactors the functionality for ignoring lines of code containing strings (cf. #413) into a common place, since that functionality is needed by both tips and prerequisite checks.

/cc @rabidbit

### DIFF
--- a/client/question/question.html
+++ b/client/question/question.html
@@ -69,6 +69,7 @@
     <script src="services/ServerHandlerService.js"></script>
     <script src="services/SessionIdService.js"></script>
     <script src="services/SolutionHandlerService.js"></script>
+    <script src="services/TipsGeneratorService.js"></script>
     <script src="services/TranscriptService.js"></script>
     <script src="services/ReinforcementGeneratorService.js"></script>
     <script src="services/code_evaluators/CodeRunnerDispatcherService.js"></script>
--- a/client/question/questionSpec.js
+++ b/client/question/questionSpec.js
@@ -40,7 +40,7 @@ describe('question.js', function() {
   }));
 
   describe('WRONG_LANGUAGE_ERRORS', function() {
-    it('should have a valid and consistent internal structure', function() {
+    it('should have the correct keys and valid values', function() {
       WRONG_LANGUAGE_ERRORS.python.forEach(function(error) {
         expect(typeof error.errorName).toBe('string');
         expect(typeof error.regExString).toBe('string');
@@ -58,7 +58,7 @@ describe('question.js', function() {
   });
 
   describe('SYSTEM_GENERATED_TIPS', function() {
-    it('should have a valid and consistent internal structure', function() {
+    it('should have the correct keys and valid values', function() {
       SYSTEM_GENERATED_TIPS.python.forEach(function(tip) {
         expect(typeof tip.type).toBe('string');
         expect(typeof tip.regExString).toBe('string');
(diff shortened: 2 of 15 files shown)


client/question/question.html
```diff
@@ -69,6 +69,7 @@
     <script src="services/ServerHandlerService.js"></script>
     <script src="services/SessionIdService.js"></script>
     <script src="services/SolutionHandlerService.js"></script>
+    <script src="services/TipsGeneratorService.js"></script>
```
eyurko: Should this be TipGenerator? Any reason for the plural?

client/question/questionSpec.js
```diff
@@ -41,15 +46,28 @@ describe('question.js', function() {
...
+          expect(ALLOWED_PARAGRAPH_TYPES.indexOf(paragraph.type) !== -1).toBe(
+            true);
+          expect(typeof paragraph.content).toEqual('string');
+        });
+      });
+    });
+  });
+
+  describe('SYSTEM_GENERATED_TIPS', function() {
+    it('should have a valid and consistent internal structure', function() {
```
eyurko: What does valid and consistent mean, in this case?

client/question/services/code_evaluators/PythonPrereqCheckService.js
```diff
@@ -31,23 +31,36 @@ tie.factory('PythonPrereqCheckService', [
       WRONG_LANGUAGE_ERRORS, PREREQ_CHECK_TYPE_INVALID_SYSTEM_CALL,
       PREREQ_CHECK_TYPE_INVALID_AUXILIARYCODE_CALL, CLASS_NAME_AUXILIARY_CODE,
       CLASS_NAME_SYSTEM_CODE) {
+
+    /**
+     * Returns a list of lines in the code that do not contain strings, in
+     * order to prevent spurious false positives in code-analysis checks.
+     *
+     * NOTE TO DEVELOPERS: This check assumes that the code does not contain
```
eyurko: Should we add a TODO to fix that in the future? This seems like it could cause some issues down the line otherwise.

client/question/question.html
seanlip: This is deliberate. We generally provide one piece of feedback and one reinforcement object per submission. However, a submission may elicit multiple tips, and the functions in the service definition return a list of tips, not just one.

client/question/questionSpec.js
seanlip: I actually copied this from similar strings in this file. I believe it means that each dict contains the correct set of keys, and that the corresponding values are valid.

I've changed the string for all the tests here to be "should have the correct keys and valid values".

client/question/services/code_evaluators/PythonPrereqCheckService.js
seanlip: Good idea; done.

client/question/question.html
eyurko: Fair enough.

PR #442, opened 2017-11-10
Add support for event batching to the frontend.

eyurko:


client/question/components/LearnerViewDirective.js
```diff
@@ -769,6 +769,16 @@ tie.directive('learnerView', [function() {
         };
 
         /**
+         * Triggers the SendEventBatch method on an interval defined by
+         * DEFAULT_EVENT_BATCH_PERIOD_SECONDS.
+         */
+        if (ServerHandlerService.doesServerExist()) {
+          $interval(function() {
+            EventHandlerService.sendCurrentEventBatch();
```
seanlip: Perhaps send only if it's nonempty?

client/question/question.js
```diff
@@ -633,6 +633,16 @@ tie.constant('SECONDS_TO_MILLISECONDS', SECONDS_TO_MILLISECONDS);
...
+
+/**
+ * Default time in seconds between calls to SendEventBatch.
+ *
+ * @type {number}
+ * @constant
+ */
+var DEFAULT_EVENT_BATCH_PERIOD_SECONDS = 30;
+tie.constant(
+  'DEFAULT_EVENT_BATCH_PERIOD_SECONDS', DEFAULT_EVENT_BATCH_PERIOD_SECONDS);
```
seanlip: Add newline below this one.

client/question/services/EventHandlerService.js
```diff
@@ -19,7 +19,38 @@
 tie.factory('EventHandlerService', [
   '$http', 'ServerHandlerService', function($http, ServerHandlerService) {
 
+
```
seanlip: Drop extra newline.

client/question/services/EventHandlerService.js
```diff
@@ -19,7 +19,38 @@
...
+    var currentEventBatch = [];
+
+    /**
+     * Submits the current batch of event data to TIE's backend.
+     *
+     */
+    var sendCurrentEventBatch = function() {
+      var data = {
+        events: currentEventBatch,
+        createdMsec: (new Date()).getTime()
```
seanlip: I thought this was going to be timeSentToBackendMsec?

client/question/services/EventHandlerService.js
```diff
@@ -19,7 +19,38 @@
...
+      };
+      return $http.post('/ajax/event/send_event_batch', data).then(
+        function() {
+          while (currentEventBatch.length > 0) {
```
seanlip: I think you can just do "currentEventBatch.length = 0" to make it empty again. (Weird, I know.) https://stackoverflow.com/questions/1232040/how-do-i-empty-an-array-in-javascript

client/question/services/EventHandlerService.js
```diff
@@ -19,7 +19,38 @@
...
+            currentEventBatch.pop();
+          }
+        }, function() {
+          // Otherwise, we should do nothing.
+          // Since the call errored, don't dump the current batch.
+      });
```
seanlip: Consider putting } on the previous line so that the brace indentations match.

client/question/services/EventHandlerService.js
```diff
@@ -19,7 +19,38 @@
...
+    };
+
     return {
+      currentEventBatch: currentEventBatch,
```
seanlip: Mm, I think you want currentEventBatch to be private. Expose it via a getter if needed for tests. You can also signal that it's private by prepending an underscore, e.g. https://github.com/google/tie/blob/master/client/question/services/FeedbackGeneratorService.js#L710

client/question/services/EventHandlerService.js
```diff
@@ -28,9 +59,15 @@ tie.factory('EventHandlerService', [
       createSessionPauseEvent: function(sessionId) {
         if (ServerHandlerService.doesServerExist()) {
           var data = {
-            sessionId: sessionId
+            sessionId: sessionId,
+            createdMsec: (new Date()).getTime()
+          };
+          var event = {
```
seanlip: Apparently `event` is a global value in IE (that's why you see it as blue when looking at it on github). It might be better to use evt instead, here and below.

client/question/services/EventHandlerService.js
```diff
@@ -28,9 +59,15 @@ tie.factory('EventHandlerService', [
       createSessionPauseEvent: function(sessionId) {
         if (ServerHandlerService.doesServerExist()) {
           var data = {
```
seanlip: Optional, here and below: you could just directly declare

```
    var evt = {
      type: ...,
      data: {
        sessionId: sessionId,
        ...
      }
    }
```

or even do:

```
    currentEventBatch.push({
      type: ...,
      data: {
        sessionId: sessionId,
        ...
      }
    });
```

client/question/services/EventHandlerServiceSpec.js
```diff
@@ -58,40 +60,35 @@ describe('EventHandlerService', function() {
...
-    it('sends a POST request to the backend to create a SessionResumeEvent',
+    it('creates a SessionResumeEvent and adds it to the current EventBatch',
       function() {
-        $httpBackend.expectPOST(
-          '/ajax/event/create_session_resume_event').respond(
-          HTTP_STATUS_CODE_OK, {});
         spyOn(ServerHandlerService, 'doesServerExist').and.returnValue(true);
         EventHandlerService.createSessionResumeEvent(sessionId);
-        $httpBackend.flush();
+        expect(EventHandlerService.currentEventBatch.length).toEqual(1);
```
seanlip: Here and below: either check that the length is 0 at the outset, or verify the type of the first event at the end, or both.

client/question/components/LearnerViewDirective.js
eyurko: Reasonable.

client/question/question.js
eyurko: Done.

client/question/services/EventHandlerService.js
eyurko: Done.

client/question/services/EventHandlerService.js
eyurko: Oop; you're right, sorry.

client/question/services/EventHandlerService.js
eyurko: That is weird.

client/question/services/EventHandlerService.js
eyurko: Done.

client/question/services/EventHandlerService.js
eyurko: okay.

client/question/services/EventHandlerService.js
eyurko: Done.

client/question/services/EventHandlerService.js
eyurko: Huh. I'll just do the other thing you suggested.

client/question/services/EventHandlerServiceSpec.js
eyurko: Done.

client/question/question.js
```diff
@@ -633,6 +633,17 @@ tie.constant('SECONDS_TO_MILLISECONDS', SECONDS_TO_MILLISECONDS);
...
  */
 tie.constant('DEFAULT_AUTOSAVE_SECONDS', 5);
+
+/**
+ * Default time in seconds between calls to SendEventBatch.
+ *
+ * @type {number}
+ * @constant
+ */
+var DEFAULT_EVENT_BATCH_PERIOD_SECONDS = 30;
```
seanlip: Why declare this? Just put 30 in line 645 -- don't want to add stuff in global scope.

client/question/services/EventHandlerService.js
```diff
@@ -19,18 +19,58 @@
...
+      var data = {
+        events: _currentEventBatch,
+        timeSentToBackendMsec: (new Date()).getTime()
+      };
+      return $http.post('/ajax/event/send_event_batch', data).then(
+        function() {
+          _currentEventBatch.length = 0;
+        }, function() {
+          // Otherwise, we should do nothing.
+          // Since the call errored, don't dump the current batch.
```
seanlip: Just to check, do we have any guarantees that a backend error here implies that no event in the batch was recorded? Need to make sure we don't process a batch partially and then error, and that this constraint/requirement holds going forward. (Might want to make a note to that effect in the corresponding backend changelist.)

client/question/services/EventHandlerService.js
```diff
@@ -19,18 +19,58 @@
...
+      });
+    };
+
+    var getCurrentEventBatchLength = function() {
+      return _currentEventBatch.length;
+    };
+
     return {
+      getCurrentEventBatchLength: getCurrentEventBatchLength,
```
seanlip: In the previous comment, I meant that you could prefix the key with an underscore: _getCurrentEventBatchLength

client/question/question.js
eyurko: Linter.

client/question/services/EventHandlerService.js
eyurko: Hm. That's a good point, but I don't think we have that constraint, currently. What we COULD do is just create all the objects and then put() them in a final push, rather than create and put them at the same time?

seanlip: Yes, that seems reasonable. We'll also need a comment in the relevant place to ensure that that state of affairs is preserved going forward.

client/question/question.js
seanlip: Er, no -- we don't have that issue in the above line (DEFAULT_AUTOSAVE_SECONDS).

eyurko: That's correct -- the linter ignores magic numbers up to and including 5. :)

Somewhere in the file there's another example for 1000.

client/question/services/EventHandlerService.js
eyurko: I'll try to find a place to add that in my current fix.

client/question/question.js
seanlip: Ah, I see, thanks for clarifying! It's fine to add those specific numbers (30, 1000) to [.eslintrc](https://github.com/google/tie/blob/master/.eslintrc.json#L103). I think it's better to do that than to introduce global vars.

eyurko: I see. The problem is that it changes it for the entire codebase, rather than the specific example, which can cause problems down the line. :\ 

Do you know of, like, a happy medium?

seanlip: Yes, you can disable the linter for specific lines, just like python pragmas. See e.g. https://eslint.org/docs/user-guide/configuring#disabling-rules-with-inline-comments (6th example).

seanlip: (I suggest doing the one with the comment on the right of the line, since that's similar to how we'd do it for python.)

client/question/services/EventHandlerService.js
eyurko: Ah, okay. Done.

client/question/question.js
eyurko: I did it above to avoid any line-length issues.

PR #445, opened 2017-11-11
Add task-specific tips to the feedback flow.

seanlip:
This PR adds task-specific tips (e.g. "you don't need regexes for this question") to the feedback flow. I still need to add tests for:

- CodeEvalResult.prototype.getIndexOfFirstFailedTask
- TipsGeneratorService.getTipParagraphs() to account for question-specific tips

and, in a future PR, make changes to show a given tip at most twice in succession (currently we show a tip whenever it applies).

In the meantime, though, @rabidbit -- could you PTAL at the code/UX? I'd like to confirm that the behavior and core ideas are correct. Thanks!

### DIFF
--- a/client/question/domain/CodeEvalResultObjectFactory.js
+++ b/client/question/domain/CodeEvalResultObjectFactory.js
@@ -135,6 +135,45 @@ tie.factory('CodeEvalResultObjectFactory', [
     };
 
     /**
+     * Compares the observed outputs to the expected values of the test cases,
+     * and returns the index of the first task with at least one failing test
+     * case.
+     *
+     * @returns {number|null} The index of the first task that failed, or null
+     * if all tasks passed.
+     */
+    CodeEvalResult.prototype.getIndexOfFirstFailedTask = function(tasks) {
+      if (this._observedOutputs.length === 0) {
+        // This can occur if there is a runtime or infinite-loop error.
+        return 0;
+      }
+
+      for (var i = 0; i < tasks.length; i++) {
+        var taskPassed = true;
+        var testSuites = tasks[i].getTestSuites();
+        for (var j = 0; j < testSuites.length; j++) {
+          var suitePassed = true;
+          var testCases = testSuites[j].getTestCases();
+          for (var k = 0; k < testCases.length; k++) {
+            if (!testCases[k].matchesOutput(this._observedOutputs[i][j][k])) {
+              suitePassed = false;
+              break;
+            }
+          }
+          if (!suitePassed) {
+            taskPassed = false;
+            break;
+          }
+        }
+        if (!taskPassed) {
+          return i;
+        }
+      }
+
+      return null;
+    };
+
+    /**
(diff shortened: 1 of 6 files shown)


rabidbit:
Hi seanlip, would you happen to have sample user code I can test this functionality with?

seanlip:
To trigger the checks, just include something like `import re` or `# Use regexes` in your code for the "reverse words" question. Here is the list of checks for that question that we currently support: https://github.com/google/tie/blob/master/assets/questions/reverseWords.js#L144

rabidbit:
Thanks, I see it now. Do we need "Tip:" prepended to the tip content? I agree with the concept of returning hints for scenarios like this, but I think it would be better to have more flexibility in what the actual feedback text is. In my mind, the key value is the ability to return feedback under certain conditions like this, separate from what actual text is given.

seanlip:
I think what I would say is, in the current context, would you like "Tip:" prepended? The reason I left that in was to convey that the feedback is non-blocking, but I can take it out if you prefer.

If it is a good idea in the current context, and the concern is about applicability to future contexts, then I would recommend leaving "Tip: " in for now and updating it when the future context actually arises. However, if it is not a good idea in the current context, then I'll get rid of it.

rabidbit:
I personally think it's better to leave it out. It will be more conversational that way.

seanlip:
Done!

client/question/domain/CodeEvalResultObjectFactory.js
```diff
@@ -135,6 +135,45 @@ tie.factory('CodeEvalResultObjectFactory', [
     };
 
     /**
+     * Compares the observed outputs to the expected values of the test cases,
+     * and returns the index of the first task with at least one failing test
+     * case.
+     *
+     * @returns {number|null} The index of the first task that failed, or null
```
rabidbit: Insert var name (task)?

client/question/services/TipsGeneratorService.js
```diff
@@ -24,31 +24,56 @@ tie.factory('TipsGeneratorService', [
...
     /**
      * Returns a list of paragraphs indicating system-generated tips for the
-     * learner.
+     * learner. At most one system-generated tip is presented.
      *
      * @param {string} language The language in which the code is written.
-     * @param {string} code The code to analyze.
-     * @returns {Array} A list of FeedbackParagraph objects, representing
-     *   system-generated tips, to prepend to the regular TIE feedback.
+     * @param {string} codeLines The lines of code to analyze.
```
rabidbit: Is this a string or an array? Used as an array below?

seanlip: Done.

client/question/domain/CodeEvalResultObjectFactory.js
seanlip: Done.

seanlip:
@rabidbit PTAL

PR #447, opened 2017-11-15
NOT FOR MERGE. Provide feedback in cases where code is not edited; provide 'correctness feedback' if we run out of more specific feedback.

seanlip:
IMPORTANT: This PR should **not be merged** into master.

The logic here needs to be redone properly at a later stage in a way that's more maintainable. The aim of this PR is to quickly verify some new functionality that is needed for an imminent study.

/cc @rabidbit

### DIFF
--- a/client/question/components/LearnerViewDirective.js
+++ b/client/question/components/LearnerViewDirective.js
@@ -122,7 +122,13 @@ tie.directive('learnerView', [function() {
                     ng-class="{'active': !nextButtonIsShown}"
                     ng-click="submitCode(editorContents.code)"
                     ng-disabled="nextButtonIsShown">
-                  Run
+                  I'm done!
+                </button>
+                <button class="tie-run-button tie-button protractor-test-run-code-btn"
+                    ng-class="{'active': !nextButtonIsShown}"
+                    ng-click="submitCode(editorContents.code)"
+                    ng-disabled="nextButtonIsShown">
+                  I need help
                 </button>
               </div>
             </div>
--- a/client/question/services/FeedbackGeneratorService.js
+++ b/client/question/services/FeedbackGeneratorService.js
@@ -296,6 +296,9 @@ tie.factory('FeedbackGeneratorService', [
       // Provide a new hint if the student gets stuck on the same bug despite
       // having modified their code, since in that case we don't want to give
       // the same message twice in a row.
+      // NOTE(sll): This doesn't work correctly anymore with the interposition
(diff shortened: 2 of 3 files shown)


rabidbit:
Because the suite level tests execute before the correctness tests, indication that a regression has occurred is not seen until after the suite level test feedback are presented (again). See below:

screen shot 2017-11-15 at 10 20 50 am

seanlip:
Update: added an "I need help" button, as discussed. See bottom right area of screenshot below:

screenshot from 2017-11-20 13 17 39

/cc @rabidbit

seanlip:
@rabidbit @eyurko -- per our chat, I changed both buttons to grey and modified the text on the "I'm done" one. What do you think? Also /cc @theresa-lee. 

screenshot from 2017-11-20 16 06 55

I'm a bit concerned now that the two buttons on the right don't stand out enough from the other buttons in the UI, which are kind of optional. I feel like the user would read the two buttons on the left (e.g. "New to Python?") and get the impression that buttons of this type/color are generally ignorable.

I'm not really sure what the correct thing to do here is, though, and would appreciate advice.

rabidbit:
Try rgba(110, 150, 190, 1) or #6e96be for the button color and rgba(50, 133, 190, 1) or #3285be for on-hover. This doesn't stand out as much as the current color scheme but it stands out enough to make it appear like a logical next button to press.

seanlip:
Will do. You mean for both the buttons, right?

rabidbit:
Yes (both buttons), since they're functionally identical anyway.

rabidbit:
For Dark theme, I'd recommend rgba(110, 150, 190, 1) or #465a6e and rgba(85, 115, 150, 1) or #557396 to lessen the contrast.

seanlip:
I just tried this. I like the new colour, but I needed to change the text to white since black-on-blue looked a bit off to me.

However, I think we actually want the buttons to have different semantics. I had another idea, which is to highlight both buttons, but give them different colours, and in particular associate "I think I'm done" with success. I feel like this achieves our aims better, since it emphasizes both buttons whilst still making clear that they're different in nature. What do you think?

screenshot from 2017-11-20 16 32 19

rabidbit:
@theresa-lee for her opinion from an aesthetics perspective.

The green color looks OK to me in general; there are not many primary colors we can use apart from green in this situation anyway.

seanlip:
Superseded by #472.

PR #449, opened 2017-11-22
Introduce two buttons at the bottom right, one for 'Done' and the other for 'Need help'.

seanlip:


client/question/components/LearnerViewDirective.js
```diff
@@ -118,11 +118,17 @@ tie.directive('learnerView', [function() {
                 <div class="tie-code-auto-save" ng-class="{'night-mode': isInDarkMode}" ng-show="autosaveTextIsDisplayed">
                   Saving code...
                 </div>
+                <button class="tie-run-button tie-button tie-button-green"
+                    ng-class="{'active': !nextButtonIsShown}"
+                    ng-click="submitCode(editorContents.code)"
+                    ng-disabled="nextButtonIsShown">
+                  I think I'm done
+                </button>
                 <button class="tie-run-button tie-button tie-button-blue protractor-test-run-code-btn"
```
rabidbit: protractor-test-run-code-btn doesn't apply to the "I think I'm done" button?

client/question/components/LearnerViewDirective.js
```diff
@@ -176,13 +182,35 @@ tie.directive('learnerView', [function() {
...
+          background-color: rgb(50, 133, 90);
+          border: 1px solid rgb(42, 112, 132);
+        }
+        .night-mode .tie-button-green {
+          background-color: rgb(70, 90, 10);
+          color: #ffffff;
+        }
+        .night-mode .tie-button-green:hover {
+          background-color: rgb(85, 115, 50);
+        }
```
rabidbit: Any reason why above does not follow the same convention as night-mode rules below? If there is a reason for the different convention, what about alphabetical order?

rabidbit:
Buttons, labels, and colors LGTM. Just have a couple of questions on the class names and css rules.

client/question/components/LearnerViewDirective.js
seanlip: Yes, this is deliberate. For determinism, I want the e2e test to trigger a specific button, not a random choice between the two.

client/question/components/LearnerViewDirective.js
seanlip: This is also deliberate, but does introduce an inconsistency. I want to avoid having a whole bunch of "night-mode" classes scattered throughout the HTML, and instead depend on a top-level night-mode class; hence this way of doing it. (I'm happy to refactor the other usages in line with this too.)

I also wanted to keep the night-mode classes near the main classes, because the two are strongly related, and I want there to be a visual reminder that, when someone edits the main class, they should also at least take a look at the night-mode class. So I propose leaving the order as-is.

So, in response to this comment, I plan to refactor the stuff mentioned in the first paragraph but leave the code as-is in terms of ordering. Does that sound OK?

client/question/components/LearnerViewDirective.js
rabidbit: OK

client/question/components/LearnerViewDirective.js
rabidbit: Consistency is good (would be great if existing rules are updated to follow new convention).

If we are no longer following strict alphabetical convention, would be good to note that and how one should determine the order (it feels a bit tricky to say it's alphabetical order unless the author or reviewer feels a different order is better based on relationships). I think there is already the potential for other exceptions, such as where cascading rules may be impacted by order; so, adding another exception based on relationships might make any ordering rule more difficult to follow IMO.

seanlip:
Done -- PTAL.

client/question/components/LearnerViewDirective.js
seanlip: Per offline conversation, decided that, although there's a rule for alphabetizing attributes within a CSS rule, there is no guideline for alphabetizing rules -- so, leaving current order as-is.

seanlip: Also, I've updated the other night-mode rules to follow the new convention (and reduce duplication).

theresa-lee:
The removing of night mode on each element makes more sense (and is cleaner), so I am okay with the the changes to the CSS for that! The colors, not too sure about but we can ask Amy about it? I am fine with leaving it as it is for now.

PR #451, opened 2017-12-01
Convert the tip generator service to a service for showing unprompted feedback.

seanlip:
@eyurko PTAL. I know I still need to add tests for unprompted feedback, but that's the only thing left. I think a review of the overall approach in the meantime would be helpful, though.

Also, @rabidbit: is this an acceptable intermediate state for the 'master' branch (i.e., if we were to do a study with this, would it be an improvement over the current version of the master branch, or should I hold off on merging)?

For information, my next steps after this PR is submitted will be:
- Restyle feedback to conversation bubbles
- Show previous code submissions in feedback, allowing expansion if needed

Thanks!


seanlip:
@eyurko -- PTAL when you get a chance. I've confirmed with @rabidbit that this can go in as-is; he will do the animations in a separate PR. Thanks!

PR #453, opened 2017-12-06
Show syntax errors and output printing requests in a modal. Have feedback come in from the top.

seanlip:
Hi @rabidbit -- PTAL and let me know if any changes are needed before this can be merged. For non-essential updates, I think it would be better to do those in a separate PR after getting the other basic stuff sorted out (e.g. stuff coming in from the top, styling of bubbles, etc.).

Thanks!

### DIFF
--- a/client/question/components/MonospaceDisplayModalDirective.js
+++ b/client/question/components/MonospaceDisplayModalDirective.js
@@ -22,15 +22,15 @@ tie.directive('monospaceDisplayModal', [function() {
     restrict: 'E',
     scope: {},
     template: `
-      <div class="container">
+      <div class="tie-monospace-modal-container">
         <header>
-          <div class="table-cell title">
+          <div class="tie-monospace-modal-table-cell tie-monospace-modal-title">
             {{title}}
           </div>
         </header>
-        <div class="body">
-          <div class="table-cell content-container">
-            <div class="content">
+        <div class="tie-monospace-modal-body">
+          <div class="tie-monospace-modal-table-cell tie-monospace-modal-content-container">
+            <div class="tie-monospace-modal-content">
               <div ng-repeat="line in contentLines track by $index">
                 {{line}}
               </div>
@@ -38,8 +38,8 @@ tie.directive('monospaceDisplayModal', [function() {
           </div>
         </div>
         <footer>
-          <div class="table-cell action-button-container">
-            <button class="tie-button-blue action-button"
+          <div class="tie-monospace-modal-table-cell tie-monospace-modal-action-button-container">
+            <button class="tie-button-blue tie-monospace-modal-action-button"
                     ng-click="closeModal()">
               <span>Close</span>
             </button>
@@ -48,31 +48,37 @@ tie.directive('monospaceDisplayModal', [function() {
       </div>
 
       <style>
-        monospace-display-modal .container {
+        monospace-display-modal .tie-monospace-modal-container {
           display: table;
           margin: 0;
           height: 100%;
           width: 100%;
         }
         monospace-display-modal header,
-        monospace-display-modal .body,
+        monospace-display-modal .tie-monospace-modal-body,
         monospace-display-modal footer {
           display: table-row;
         }
         monospace-display-modal header, monospace-display-modal footer {
           background: #eee;
         }
-        monospace-display-modal .table-cell {
+        .night-mode monospace-display-modal header,
+        .night-mode monospace-display-modal footer {
+          background: #444;
+          color: #ccc;
+        }
+
+        monospace-display-modal .tie-monospace-modal-table-cell {
           display: table-cell;
           padding: 10px;
         }
 
-        monospace-display-modal .title {
+        monospace-display-modal .tie-monospace-modal-title {
           font-size: 18px;
           font-weight: bold;
           text-align: center;
         }
-        monospace-display-modal .content-container {
+        monospace-display-modal .tie-monospace-modal-content-container {
           background: #fff;
           color: #111;
           font-family: monospace;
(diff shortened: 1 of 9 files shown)


rabidbit:
Minor UI tweaks per offline chat (most can be in a separate PR).

seanlip:
Listing next steps here for reference:
- Modal should cover Light/Dark mode button.
- Need to add tests for the new modal service.
- Modal size should be equal to that of the left pane (OK to do this in another PR)
- Create dark mode version (OK to do this in another PR)

client/question/components/MonospaceDisplayModalDirective.js
```diff
@@ -0,0 +1,129 @@
...
+ * @fileoverview Directive for the modal displaying monospaced code/error
+ * messages.
+ */
+
+tie.directive('monospaceDisplayModal', [function() {
+  return {
+    restrict: 'E',
+    scope: {},
+    template: `
+      <div class="container">
```
theresa-lee: Shouldn't we use the same "tie-" naming conventions here in this file for the elements? "container", "body" and "content" are generic.

client/question/components/MonospaceDisplayModalDirective.js
```diff
@@ -0,0 +1,129 @@
...
+        <div class="body">
+          <div class="table-cell content-container">
+            <div class="content">
+              <div ng-repeat="line in contentLines track by $index">
+                {{line}}
+              </div>
+            </div>
+          </div>
+        </div>
+        <footer>
```
theresa-lee: Unsure where it's actually coming from (it shows up on the <footer> when inspecting), but there's like lines on the corner that indicate the window is resizeable. If we aren't allowing that, should probably remove it? The one on the feedback works.

seanlip: Oh, good point. I was actually thinking that it's probably OK for the modal window to be resizable but not movable, similar to the feedback window that it replaces. @rabidbit could you confirm?

client/question/components/OutputSnippetDirective.js
```diff
@@ -20,55 +20,22 @@ tie.directive('outputSnippet', [function() {
...
-        <a href class="toggle" ng-click="toggleOutput()">
-          {{outputIsShown ? 'Hide' : 'Display'}} output
-        </a>
-      </div>
-      <div class="output-container" ng-show="outputIsShown">
-        <span ng-repeat="line in snippetLines">
-          <span class="snippet-line">{{line}}</span>
-          <br>
-        </span>
+        You can click on <a href ng-click="openOutputModal()">this link</a>
```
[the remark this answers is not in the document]
seanlip: Yeah, fair point. Do you want it to not change in non-dark mode, too? If so, should it just stay as whatever the original colour of the link is?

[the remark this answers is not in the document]
theresa-lee: Yeah, I'm okay with it never having the 'visited' purpled link in either of the themes. Since it's likely that if you are using the app, you'll want to click through that link and it doesn't matter whether or not you've clicked it before?

client/question/components/MonospaceDisplayModalDirective.js
rabidbit: The modal window should not be resizable. It should be static in position and size (completely cover the feedback window and related controls). Allowing it to be resizable makes it more complicated as the feedback window would then either also have to be resized at the same time (which is not intuitive) or the modal could be resized to expose the feedback window behind it (which is what we want to avoid).

client/question/components/MonospaceDisplayModalDirective.js
seanlip: Done.

client/question/components/MonospaceDisplayModalDirective.js
seanlip: Sounds good, thanks for the confirmation. This is fixed.

client/question/components/OutputSnippetDirective.js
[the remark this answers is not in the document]
seanlip: OK, sounds good -- done!

client/question/components/SyntaxErrorSnippetDirective.js
```diff
@@ -20,15 +20,19 @@ tie.directive('syntaxErrorSnippet', [function() {
...
-        <a href class="toggle" ng-click="toggleSyntaxErrorHint()">
-          {{syntaxErrorIsShown ? 'Hide' : 'Display'}} error details
-        </a>
+        <p>
+          It looks like your code has a syntax error. Try to figure out what the
+          error is.
+        </p>
+        <p>
+          If you cannot figure out the problem, you can click on
+          <a href class="toggle" ng-click="openSyntaxErrorModal()">this link</a>
```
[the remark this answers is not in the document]
seanlip: Done.

seanlip:
Hi @rabidbit @theresa-lee -- thanks for the reviews! I expect this version to be mergeable, so PTAL when you get a chance.

I've fixed everything (including the modal-size issue and the dark mode colours), except for one thing, which is @rabidbit's suggestion that the modal cover the Light/Dark theme selector. Implementing that isn't straightforward because the modal adopts the current size of the feedback window, and the resizable region does not include the theme selector. We could hack around this, but that would make things somewhat messy because of the assumptions we'd need to make around the height and padding/margin of the selector, and the resulting code will be harder to maintain. So, I've opted to leave it as-is for now, but please let me know whether that's OK from a UI/UX perspective.

rabidbit:
I think for this PR, it's OK to leave the modal size as is.

seanlip:
OK, sounds good, thanks! @theresa-lee, PTAL and review/merge when you get a chance -- I'll work on the conversation bubble styling as a follow-up.

theresa-lee:
Changes look good except now I'm seeing that the two windows are no longer the same size? Both on my desktop and laptop.

theresa-lee:
1wrhmdna5q2

seanlip:
Fixed, thanks!

PR #457, opened 2017-12-12
Update frontend to use task ID string instead of int and add missing event thingies

eyurko:


client/question/components/LearnerViewDirective.js
```diff
@@ -737,13 +737,14 @@ tie.directive('learnerView', [function() {
           // or SessionResumeEvent, respectively.
           var hiddenAttributeName = (
             $scope.determineHiddenAttributeNameForBrowser());
-          if (hiddenAttributeName !== null) {
+          if (hiddenAttributeName !== null && tasks !== null &&
+              tasks.length > currentTaskIndex) {
```
seanlip: Why all these extra conditions?

client/question/components/LearnerViewDirective.js
```diff
@@ -871,12 +872,16 @@ tie.directive('learnerView', [function() {
           $scope.instructions = tasks[currentTaskIndex].getInstructions();
           $scope.previousInstructions = [];
           $scope.nextButtonIsShown = false;
+          EventHandlerService.initEventHandlerServiceState(
```
seanlip: I think you can just call this EventHandlerService.init()

client/question/services/EventHandlerService.js
```diff
@@ -53,20 +67,28 @@ tie.factory('EventHandlerService', [
...
+       * @param {string} taskId Id of the task being paused.
        *
        */
-      createSessionPauseEvent: function(sessionId) {
-        if (ServerHandlerService.doesServerExist()) {
+      createSessionPauseEvent: function(taskId) {
+        // Checking for null here in the off chance that a user pauses
+        // before the question is loaded, creating a minor race condition.
+        if (ServerHandlerService.doesServerExist() &&
+            _currentSessionId !== null &&
```
seanlip: Factor these three inequality checks into an isInitialized() method.

Also, if you want to be safe, then for other methods in EventHandlerService you can also do a "if not init-ed, throw Error" check at the outset. But that's optional.

client/question/components/LearnerViewDirective.js
eyurko: In case the question isn't fully initialized when a user switches tabs.

Just trying to be robust.

client/question/services/EventHandlerService.js
eyurko: Okay. I'll do the isInitialized bit.

client/question/components/LearnerViewDirective.js
eyurko: Done.

client/question/components/LearnerViewDirective.js
seanlip: OK. Is it possible that this might lead to an extra SessionResumeEvent (with no corresponding SessionPauseEvent)?

client/question/services/EventHandlerServiceSpec.js
```diff
@@ -232,5 +189,26 @@ describe('EventHandlerService', function() {
         EventHandlerService.sendCurrentEventBatch();
         expect(EventHandlerService._getCurrentEventBatchLength()).toEqual(0);
       });
+
+    it('does not send the batch if any of sId / qId / qVer are null',
```
seanlip: Where in this test do we check that the batch isn't sent? I'd expect it to spy on something and verify it never gets called.

Also, at the end, perhaps init() it correctly and check that the batch actually does get sent in that case?

eyurko: If the batch were sent it would trigger an httprequest that the test would be responsible for catching (see other send batch tests). That doesn't occur, and we see that the batch keeps getting longer as additional events aren't being sent.

Technically the other tests check that (as it's init'ed correctly in the beforeEach). Happy to add another check if you would like.

client/question/components/LearnerViewDirective.js
eyurko: I suppose that it's possible. What should we do about it?

client/question/services/EventHandlerServiceSpec.js
seanlip: Nope, that's fine I think. Thanks for the explanation!

client/question/components/LearnerViewDirective.js
seanlip: Discussed offline. It's probably fine; we can check later to see if there are any event discrepancies. Marking this as resolved.

PR #461, opened 2017-12-14
New chat-like UI

rabidbit:
Initial commit. Still need to review/tidy it and create required tests.


rabidbit:
But you can play around with it to see how it works. I haven't had a chance to work on the modal/pop-up styling and animation, but I'll in include that as well.

rabidbit:
I'll ping you guys when this is ready for a full review (just wanted to give you a heads up).

seanlip:
Just some quick UI feedback as requested. (I think it's OK to do the modal animations in a separate PR altogether, btw.)

- I think the animation looks nice! It could perhaps be better in one way: when the new feedback is loading, the existing balloon slides down to exactly the point it would later end up taking, rather than only by the height of the loading dots. That would be more elegant, since right now there's a bit of a "jump" at the end when the feedback is long. But this is also harder so I think it's fine to defer till later.

- The input/output in the sample code seem misaligned:

    screenshot from 2017-12-13 17-52-46

- Typing "def reverseWords(s):" and running leads to an uncaught TypeError: Cannot read property 'classList' of undefined

rabidbit:
I was aware of the stutter due to the dots and plan to address. I'll look into the other two bugs. Thanks!

rabidbit:
It looks like the TypeError error is an existing bug not related to this PR. Would you mind opening an issue?

seanlip:
Ah oops, didn't realize it's unrelated. I've fixed it: #463.

theresa-lee:
It's looking pretty good to me so far! I don't know if you were planning on doing this at the end, but just noting that dark mode will need some tweaks to the color of the bubbles or else is very hard to read!

rabidbit:
The UI works now; I just need to clean up the code and add tests. Feel free to look at the UI to see how it works. Supports night mode too. Will ping you when this is ready for a full review.

rabidbit:
Hi. This is ready for review. Thanks

client/question/components/LearnerViewDirective.js
```diff
@@ -763,6 +820,7 @@ tie.directive('learnerView', [function() {
          * @param {string} questionId ID of question whose data will be loaded
          */
         $scope.loadQuestion = function(questionId) {
+          $scope.isTieThemeSet = true;
```
seanlip: I don't really understand what's going on with this variable -- could you please add some explanation in comments? In particular, isn't the theme always either 'light' or 'dark', and why is isInDarkMode not a sufficient indicator?

Also I suggest $scope.tieThemeIsSet to avoid devs thinking that this is a function (since it starts with a verb).

client/question/components/LearnerViewDirective.js
```diff
@@ -25,79 +25,92 @@ tie.directive('learnerView', [function() {
...
                     ng-disabled="nextButtonIsShown">
-                  I think I'm done
+                  I think I&#39m done
                 </button>
                 <button class="tie-run-button tie-button tie-button-blue protractor-test-run-code-btn"
                     ng-class="{'active': !nextButtonIsShown}"
                     ng-click="submitCode(editorContents.code)"
                     ng-disabled="nextButtonIsShown">
-                  I need help
+                  Check my code
```
seanlip: "I need help" and "Check my code" have different meanings to me, and the latter feels more like "I think I'm done" IMO. Just a data point.

client/question/components/LearnerViewDirective.js
```diff
@@ -780,6 +838,8 @@ tie.directive('learnerView', [function() {
             question = QuestionDataService.getQuestion(questionId);
             initQuestionData(questionId);
           }
+          $scope.speechBalloonsList =
```
seanlip: I don't believe this is used. Delete?

client/question/components/LearnerViewDirective.js
```diff
@@ -1007,6 +1069,8 @@ tie.directive('learnerView', [function() {
          * question to its original state.
          */
         $scope.resetCode = function() {
+          $scope.speechBalloonsList =
```
seanlip: Ditto -- I don't believe this is used. Delete?

client/question/components/SpeechBalloonDirectives.js
```diff
@@ -0,0 +1,113 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+/**
+ * @fileoverview Directive for displaying speech balloons.
+ */
+
+tie.directive('tieSpeechBalloonContainer', ['$timeout', function($timeout) {
+  return {
+    restrict: 'A',
```
seanlip: Could we do `restrict: 'E'` instead? Better for directives to represent DOM elements, otherwise you introduce the possibility of one DOM tag having multiple of these attrs and that's confusing.

Ditto below.

client/question/components/SpeechBalloonDirectives.js
```diff
@@ -0,0 +1,113 @@
...
+    link: function(scope, iElement) {
+      if (iElement[0].nodeName !== "DIV") {
```
seanlip: - Why is this check needed?
- What does the "i" stand for in iElement?

client/question/components/SpeechBalloonDirectives.js
```diff
@@ -0,0 +1,113 @@
...
+            '-' + speechBalloonContainer.offsetHeight.toString() + 'px';
+        $timeout(function() {
+          speechBalloonContainer.removeAttribute("style");
+        }, 0);
+      }, 0);
+    }
+  };
+}]);
+
+tie.directive('tieSpeechBalloonLeft', ['$timeout', function($timeout) {
```
seanlip: Prefer having one directive per file. OK to put all the files in a subfolder if desired.

client/question/components/SpeechBalloonDirectives.js
```diff
@@ -0,0 +1,113 @@
...
+
+tie.directive('tieSpeechBalloonContainer', ['$timeout', function($timeout) {
+  return {
+    restrict: 'A',
+    link: function(scope, iElement) {
+      if (iElement[0].nodeName !== "DIV") {
+        return;
+      }
+      var speechBalloonContainer = iElement[0];
+      speechBalloonContainer.className = "tie-speech-balloon-container";
```
seanlip: Would "addClass" be better (in case, in the future, there are other classes set on the directive)?

client/question/components/LearnerViewDirective.js
rabidbit: This is intentional. Currently, there is functionally no difference between the two buttons. Having greater contrast in the button labels makes this more obvious in a way that feels like a feature is missing or broken. While "I think I'm done" implies a greater level of confidence in having a complete answer than "Check my code", they are close enough so that the functional equivalence is less jarring. Otherwise, I suggest we make "I need help" functionally distinct.

client/question/components/LearnerViewDirective.js
rabidbit: This is used in speechBalloonsContainer directive. The existing code calls getSpeechBalloonsList() in ng-repeat, but this results in repainting of the balloons, making animation very difficult. However, simply assigning speechBalloonsList in the speechBalloonsContainer controller breaks the code reset button, as this approach does not handle removal of items during reset. So, speechBalloonsList is set in the learnerView directive whenever the user clicks on reset code.

client/question/components/LearnerViewDirective.js
rabidbit: See above.

client/question/components/SpeechBalloonDirectives.js
rabidbit: OK, will do.

client/question/components/SpeechBalloonDirectives.js
rabidbit: If I change restrict property above to E, this is no longer relevant. i stands for instance, but this can be anything (let me know if you prefer a different name).

client/question/components/SpeechBalloonDirectives.js
rabidbit: I'll try addClass.

client/question/components/SpeechBalloonDirectives.js
rabidbit: I thought about that, but these are all very tightly related and used together (doesn't really make sense to use any one in isolation). If you feel strongly about this, I will split them up into different files.

client/question/components/LearnerViewDirective.js
rabidbit: Changed name to pulseAnimationEnabled, which is probably more accurate/descriptive.

rabidbit: Added comment. Done.

client/question/components/SpeechBalloonDirectives.js
rabidbit: Done.

client/question/components/SpeechBalloonDirectives.js
rabidbit: Breaks when replacing .className, but I was able to replace .classList.add()

rabidbit: Actually, both were not working. I would need to convert these to angular elements in order to use addClass. Is this required? Feels like it's making things more complicated for a theoretical situation I am not sure will happen.

client/question/components/SpeechBalloonsContainerDirective.js
```diff
@@ -50,147 +43,143 @@ tie.directive('speechBalloonsContainer', [function() {
...
+              </tie-speech-balloon-right>
+              <tie-speech-balloon-tail-right></tie-speech-balloon-tail-right>
             </div>
-          </div>
+          </tie-speech-balloon-container>
         </div>
       </div>
 
       <style>
+        .tie-feedback-paragraph {
```
theresa-lee: For the paragraphs, I feel like there might be a bit too much extra padding on the inside between the bubble and the text? You'll be able to see more of the feedback I think if you decreased the spacing. Might just be my own personal preference though.

client/question/components/LearnerViewDirective.js
seanlip: OK, let's go with "Check my code" for now -- I think the rationale makes sense, and thanks for bringing it up!

I think the main thing I'd want to check for during user studies is whether users are still reticent to click the "Check my code" button because they haven't finished solving the problem, which was what the "I need help" button was meant for in the first place. If we see this behaviour then perhaps we can try switching and see what happens.

client/question/components/LearnerViewDirective.js
seanlip: Ditto.

client/question/components/SpeechBalloonDirectives.js
```diff
@@ -0,0 +1,104 @@
...
+// limitations under the License.
+
+/**
+ * @fileoverview Directives for displaying speech balloons.
+ */
+
+tie.directive('tieSpeechBalloonContainer', ['$timeout', function($timeout) {
+  return {
+    restrict: 'E',
+    link: function(scope, iElement) {
```
seanlip: Put `scope: {}` on each of these to enforce isolation.

seanlip: Also I suggest iElement --> element.

client/question/components/SpeechBalloonDirectives.js
```diff
@@ -0,0 +1,104 @@
...
+      var speechBalloonContainer = iElement[0];
+      speechBalloonContainer.className = "tie-speech-balloon-container";
```
seanlip: > Actually, both were not working. I would need to convert these to angular elements in order to use addClass. Is this required? Feels like it's making things more complicated for a theoretical situation I am not sure will happen.

Yes, we should still do this, I can see it breaking if we were to add e.g. protractor test classes.

Converting using angular.element() seems fine to me.

client/question/components/SpeechBalloonDirectives.js
seanlip: Yeah, I'm quite torn here for the reason you mention. Let's leave it as is for now. These feel more like animation tags (and not so much "DOM structure") and I'll need to think of a more general way to handle these / distinguish these from other more "structural" directives.

client/question/components/SpeechBalloonsContainerDirective.js
```diff
@@ -22,24 +22,17 @@ tie.directive('speechBalloonsContainer', [function() {
...
       <div>
-        <div class="tie-dot-container" ng-if="ConversationLogDataService.isNewBalloonPending()">
-          <div class="tie-dot tie-dot-1"></div>
-          <div class="tie-dot tie-dot-2"></div>
-          <div class="tie-dot tie-dot-3"></div>
-          <br>
-        </div>
-        <div ng-repeat="balloon in ConversationLogDataService.getSpeechBalloonsList() track by $index">
-          <div class="tie-speech-balloon-container">
+        <div ng-repeat="balloon in $parent.speechBalloonsList">
```
seanlip: Noted above, but just to confirm here: avoid $parent.

client/question/components/LearnerViewDirective.js
rabidbit: OK, changed to option 2. Thanks seanlip for the tips on how exactly how to implement it.

client/question/components/LearnerViewDirective.js
rabidbit: Removed

client/question/components/SpeechBalloonsContainerDirective.js
rabidbit: Reduced padding on top and bottom as suggested.

client/question/components/SpeechBalloonDirectives.js
rabidbit: Done

rabidbit: Done

client/question/components/SpeechBalloonsContainerDirective.js
rabidbit: Done

client/question/components/LearnerViewDirective.js
seanlip: So we should still delete this and the invocation below, right (and possibly even `$scope.ConversationLogDataService = ConversationLogDataService;` above)?

seanlip: Oh sorry. Didn't realize you hadn't pushed commits yet.

client/question/components/SpeechBalloonDirectives.js
rabidbit: Done

client/question/components/LearnerViewDirective.js
```diff
@@ -25,79 +25,91 @@ tie.directive('learnerView', [function() {
...
                   <div class="tie-instructions">
                     <div ng-repeat="instruction in instructions">
-                      <p ng-if="instruction.type == 'text'">{{instruction.content}}</p>
+                      <p ng-if="instruction.type == 'text'">
+                        {{instruction.content}}
+                      </p>
                       <pre class="tie-question-code" ng-if="instruction.type == 'code'">{{instruction.content}}</pre>
                     </div>
                   </div>
+                  <div class="tie-dot-container" ng-class="{'tie-display-dots': ConversationLogDataService.isNewBalloonPending()}"}>
```
seanlip: Why can't this go in the speech balloons container directive? Whether a new balloon is pending seems to be a property of that container, it's a bit odd to split it like this. The exception would be if something outside that container triggers the dots.

That way you also don't need to expose $scope.ConversationLogDataService in its entirety here.

client/question/components/SpeechBalloonDirectives.js
```diff
@@ -0,0 +1,113 @@
...
+
+tie.directive('tieSpeechBalloonContainer', ['$timeout', function($timeout) {
+  return {
+    restrict: 'E',
+    scope: {},
+    link: function(scope, element) {
+      var speechBalloonContainerElement = element[0];
+      var speechBalloonContainer =
+          angular.element(speechBalloonContainerElement);
+      speechBalloonContainer.addClass("tie-speech-balloon-container");
```
seanlip: There's some single quotes and some double quotes, use singles throughout.

client/question/components/LearnerViewDirective.js
rabidbit: It was easier to structure the UI this way to get it to behave the way it does, I didn't feel the dots are the same thing as the balloons (the broader feedback window is waiting for the balloons to respond, so I view the dots as part of the window containing the balloons), and ConversationLogDataService seems to mostly be exposed anyway.

client/question/components/SpeechBalloonDirectives.js
rabidbit: Done

client/question/components/LearnerViewDirective.js
seanlip: I'm not sure why structuring the UI this way is easier -- couldn't you just put the loading dots at the start of the child directive, or does this interfere with the animation somehow? Also, I agree that the dots aren't the same things as the balloons, but the presence of the loading dots depends only on the speech bubbles themselves -- so I'd like to strongly suggest that they stay in the child directive, since all the stuff they depend on is in there as well.

ConversationLogDataService being exposed is a temporary thing; I'm planning to find a pattern that allows retrieval without direct access to properties on the service. I think I have an idea for how to do this but will defer it to a future PR -- but in the meantime let's not propagate its exposure.

rabidbit: Done

client/question/components/LearnerViewDirective.js
```diff
@@ -468,14 +487,10 @@ tie.directive('learnerView', [function() {
...
-        /**
-         * Number of milliseconds for TIE to wait for system to process code
-         * submission.
-         *
-         * @type {number}
-         * @constant
-         */
-        var DURATION_MSEC_WAIT_FOR_SCROLL = 20;
+
+        $scope.ConversationLogDataService = ConversationLogDataService;
```
seanlip: I think you can delete this line.

PR #462, opened 2017-12-14
Add code bubbles to the feedback log.

seanlip:


seanlip:
@eyurko @theresa-lee Please could one of you do a code review? Thanks!

client/question/components/CodeSnippetDirective.js
```diff
@@ -23,19 +23,24 @@ tie.directive('codeSnippet', [function() {
       getContent: '&content'
     },
     template: `
-      <span ng-repeat="line in snippetLines">
+      <span ng-repeat="line in abbreviatedSnippetLines track by $index">
         <span class="tie-code-snippet-line">{{line}}</span>
         <br>
       </span>
+      <span ng-if="abbreviatedSnippetLines.length <= 3">
```
eyurko: Is there any reason this is a 3? If so, could we make it a constant?

Numbers by themselves worry me.

client/question/components/CodeSnippetDirective.js
```diff
@@ -52,6 +73,16 @@ tie.directive('codeSnippet', [function() {
           // not get collapsed into a single one.
           var htmlFormattedContent = newValue.replace(/ /g, '\u00A0');
           $scope.snippetLines = htmlFormattedContent.split('\n');
+
+          if ($scope.snippetLines.length <= 3) {
```
eyurko: Ditto here.

client/question/components/CodeSnippetDirective.js
seanlip: I think this just follows @rabidbit's mocks. But yeah, good point, I'll constant-ize it.

rabidbit: Are you referring to the user's dialogue (user code in conversation)? If so, we should only show three lines or so--just enough for the user to see a small snippet of code before clicking to open the modal/pop-up to view the whole code.

seanlip: Done. (And yep, I pulled 3 out into a constant along the lines of MAX_NUM_LINES_TO_SHOW_IN_ABBREVIATED_SNIPPET.)

PR #464, opened 2017-12-14
When loading question, if server version, it will pull from datastore instead of local files.

theresa-lee:
The server specific backend file changes are in a CL.

### DIFF
--- a/client/data/services/QuestionDataService.js
+++ b/client/data/services/QuestionDataService.js
@@ -34,6 +34,8 @@ tieData.factory('QuestionDataService', [
        * @param {string} questionSetId
        */
       initCurrentQuestionSet: function(questionSetId) {
+        // Currently, hardcoding set of questions when pulling from the server.
+        // TODO(talee): Find a way to make this not hard coded.
         if (ServerHandlerService.doesServerExist()) {
           currentQuestionSet = QuestionSetObjectFactory.create(
             {
--- a/client/data/services/QuestionDataService.js
+++ b/client/data/services/QuestionDataService.js
@@ -18,11 +18,12 @@
  */
 
 tieData.factory('QuestionDataService', [
-  'QuestionObjectFactory', 'QuestionSetObjectFactory', 'ServerHandlerService',
-  '$http',
+  '$http', 'QuestionObjectFactory', 'QuestionSetObjectFactory',
+  'ServerHandlerService',
   function(
-    QuestionObjectFactory, QuestionSetObjectFactory, ServerHandlerService,
-    $http) {
+    $http, QuestionObjectFactory, QuestionSetObjectFactory,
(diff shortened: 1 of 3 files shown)


client/data/services/QuestionDataService.js
```diff
@@ -18,8 +18,11 @@
  */
 
 tieData.factory('QuestionDataService', [
-  'QuestionObjectFactory', 'QuestionSetObjectFactory',
-  function(QuestionObjectFactory, QuestionSetObjectFactory) {
+  'QuestionObjectFactory', 'QuestionSetObjectFactory', 'ServerHandlerService',
+  '$http',
```
theresa-lee: Move $http to top.

client/question/components/LearnerViewDirective.js
```diff
@@ -762,6 +757,32 @@ tie.directive('learnerView', [function() {
...
+         * Initializes the appropriate values in $scope for the question
+         * instructions, stored code, starter code and feedback.
+         *
+         * @param {string} questionId ID of question whose data will be loaded
+         */
+        $scope.loadQuestion = function(questionId) {
+          SessionIdService.resetSessionId();
+          if (SERVER_URL) {
+            try {
+              QuestionDataService.getQuestionAsync(questionId).then(
```
theresa-lee: Move server check into getQuestionAsync so that it's the only call here. (Look into q.defer)

client/data/services/QuestionDataServiceSpec.js
```diff
@@ -126,3 +126,57 @@ describe('QuestionDataService', function() {
...
+
+    it('should error if questionId does not exist', function() {
+      expect(function() {
+        QuestionDataService.getQuestionAsync(questionId);
+      }).toThrowError(
+        'The current question set does not contain a question with ID: ' +
+         questionId);
+    });
+
+    it('should correctly get the question data', function(done) {
```
theresa-lee: Use $httpbackend to mock out the http call and test that instead.

theresa-lee: Done.

client/question/components/LearnerViewDirective.js
seanlip: FYI, I read up a bit on this, and it looks like $q.resolve / $q.when might be a better replacement than $q.defer. See

  https://toddmotto.com/promises-angular-q#qwhen--qresolve
  http://www.codelord.net/2015/09/24/%24q-dot-defer-youre-doing-it-wrong/

seanlip:
That said, please only submit this when the coverage for QuestionDataService.js is 100%. Thanks!

theresa-lee:
QuestionDataService test at 100%, so going ahead with this submission!

PR #465, opened 2017-12-14
Changed the python primer button to a link (Fixes #456)

### LINKED ISSUE #456: Change "New to Python" button to a link
Some users were unsure if they should press the "New to Python" button because they were not sure if it would change something in the application (versus simply opening a page).

Changing this to a link should make it easier for the user to understand what they should expect (a link to some page vs a button that could change something).

theresa-lee:
I've kept the styling to be roughly the same as a button (with padding) but it's a link instead, and it follows the style of the links that are shown in the feedback.

### DIFF
--- a/client/question/components/LearnerViewDirective.js
+++ b/client/question/components/LearnerViewDirective.js
@@ -79,9 +79,7 @@ tie.directive('learnerView', [function() {
                 <select ng-if="SERVER_URL" class="tie-select-menu" name="lang-select-menu">
                   <option value="Python" selected>Python</option>
                 </select>
-                <button ng-if="!SERVER_URL" class="tie-python-primer tie-button">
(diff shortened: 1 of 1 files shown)


rabidbit:
UI LGTM

client/question/components/LearnerViewDirective.js
```diff
@@ -204,7 +202,7 @@ tie.directive('learnerView', [function() {
           float: left;
           margin-top: 10px;
         }
-        .night-mode .tie-code-reset, .night-mode .tie-python-primer {
+        .night-mode .tie-code-reset, .night-mode {
```
seanlip: If you're deleting .tie-python-primer should the .night-mode before it be deleted as well?

theresa-lee: Yeah, that's right. Fixed!

Issue #482, opened 2018-01-03
Displaying 'Next' when we don't have more than one question ready


theresa-lee:
When you complete a question, the Next thing pops up over the coding window. But we don't any more questions to navigate to (at this point) and also pressing next brings up a popup that congratulates you on finishing the problem, which seems redundant as the feedback window already tells you that.

rabidbit:
Next button makes sense for a sequence of questions, which we do not currently have and are unclear whether we want to implement that in the near future.

For now, since we only have one question, are using a chat conversation experience, and have conditioned the user to look at the chat window for feedback, why don't we remove the Next button and overlay and instead add an emoji to the congratulatory feedback? This could keep the UI/UX simple while giving the user a little nicer reward beyond congratulatory text.

rabidbit:
Fixed in #484

PR #484, opened 2018-01-05
Fix for #482 - Removed Next button and overlay

theresa-lee:
Not sure what emoji to add but can do that in a separate PR! I just removed the button and overlay for now.

### DIFF
--- a/client/question/components/LearnerViewDirective.js
+++ b/client/question/components/LearnerViewDirective.js
@@ -186,14 +186,6 @@ tie.directive('learnerView', [function() {
         .night-mode .tie-button-green:active {
           background-color: rgb(61, 79, 8);
         }
-        .tie-button-red {
-          background-color: #ef5350;
-          color: #ffffff;
-        }
-        .tie-button-red:hover {
-          background-color: #f44336;
-          border: 1px solid #e53935;
-        }
         .tie-code-auto-save {
           font-family: Roboto, 'Helvetica Neue', 'Lucida Grande', sans-serif;
           float: left;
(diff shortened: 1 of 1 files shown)


rabidbit:
Looks like the "Check my code" and "I think I'm done" buttons are back to being blue?

client/question/components/LearnerViewDirective.js
```diff
@@ -408,18 +355,16 @@ tie.directive('learnerView', [function() {
...
           position: relative;
         }
         .tie-run-button:hover {
           box-shadow: inset 0 1px 2px rgba(0,0,0.3);
         }
-        .tie-run-button:active {
-          background-color: rgb(42, 112, 232);
-          border: 1px solid rgb(32, 100, 200);
-          box-shadow: inset 0 1px 2px rgba(0,0,0.3);
-        }
```
rabidbit: I think it's OK to have an :active rule, to distinguish it from hover over. Usually, a darker shade of the same color is used though (I think we kept the dark blue color from prior versions, though I'm not certain that was intentional--darker green for example would make sense for "I think I''m done".

theresa-lee: Whoops, sorry, I think I got a little aggressive with removing "dead" code but I put it back now! I actually made some changes to the button hovering (including what happens in dark mode), but kept the general idea of it getting darker for hover and then even a bit darker for active. The dark mode has only the border highlighted on hover (similar to the reset and theme buttons in dark mode) but also turns a bit darker on press. I am happy to change color/affect if anyone has objections.

theresa-lee: I split this part up into their respective color CSS (since we were using blue for both).

rabidbit:
One last comment, otherwise LGTM. The buttons are now brighter in light mode, whereas we intentionally made them more muted to keep the user's attention on the code. In particular, the blue on the "Check my code" stands out more. Can we keep the more muted colors for now?

Thanks for the fix!

PR #487, opened 2018-01-05
Fix #474 - "Check My Code" Button does nothing and that stresses me out

### LINKED ISSUE #474: "Check My Code" Button does nothing and that stresses me out
So I talked to John and it sounds like it's a placeholder for a currently unimplemented feature.

I'm in favor of removing it until at least that feature is implemented, because right now it's two different pathways to the same endpoint and that's very confusing for me, as a rando.

I briefly proposed changing it to "How am I doing?" but that builds in an antipattern for interviewing as we can't provide feedback to candidates on their performance. Check my code, by the same logic, might also not be good.

rabidbit:
Fixes "Check My Code" Button does nothing and that stresses me out #474

### DIFF
--- a/client/question/components/LearnerViewDirective.js
+++ b/client/question/components/LearnerViewDirective.js
@@ -84,25 +84,19 @@ tie.directive('learnerView', [function() {
                     name="lang-select-menu">
                   <option value="Python" selected>Python</option>
                 </select>
-                <a ng-if="!SERVER_URL" class="tie-primer-link tie-python-primer" target="_blank" ng-href="{{getPythonPrimerUrl()}}">New to python?</a>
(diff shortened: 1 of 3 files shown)


rabidbit:
Hmm... not sure why DELAY_STYLE_REMOVAL edits shows up. I did not change that in this PR (changed in previous PR fix).

theresa-lee:
Hmm, that is weird... the master is updated and this is a diff from the master to your fix... did you branch from master before the changes and do you need to update this branch with the new master updates?

rabidbit:
Ah, it looks like it's good now after I resolved merge conflicts. PTAL.

Issue #398, opened 2017-08-25
Add a Reset Feedback button


theresa-lee:

rabidbit:
This was identified as a useful feature during user studies.

theresa-lee:
Fixed in #489

PR #489, opened 2018-01-08
Added Reset Feedback button and separating resetting code/feedback functionality

### LINKED ISSUE #479: There is still feedback after pressing Reset Code
I'm not sure if this is considered the same root problem as #279, but when pressing the 'Reset Code' button, it clears the full queue of feedback, but then the feedback loads the latest feedback. I think this might be confusing since your code is reset and the feedback shown is probably wrong.

theresa-lee:
Using Sean's suggestion, resetting either the feedback or the code will clear the LocalStorage feedback. That means that if a user resets the code, the current feedback won't be cleared but if they refresh the page, then it won't be able to fetch any feedback, and will be blank. Resetting the feedback will remove the feedback both in the feedback window and in LocalStorage.

This should fix #479, #279, and #398.

### DIFF
--- a/client/question/components/LearnerViewDirective.js
+++ b/client/question/components/LearnerViewDirective.js
@@ -55,7 +55,7 @@ tie.directive('learnerView', [function() {
                 </monospace-display-modal>
               </div>
               <button class="tie-code-reset tie-button" ng-click="resetFeedback()">
-                  Reset Feedback
+                Reset Feedback
               </button>
               <select class="tie-select-menu" name="theme-select"
                   ng-change="changeTheme(theme)" ng-model="theme"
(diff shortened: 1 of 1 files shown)


client/question/components/LearnerViewDirective.js
```diff
@@ -54,6 +54,9 @@ tie.directive('learnerView', [function() {
                     content="content">
                 </monospace-display-modal>
               </div>
+              <button class="tie-code-reset tie-button" ng-click="resetFeedback()">
+                  Reset Feedback
```
seanlip: Nit: indent by 2

client/question/components/LearnerViewDirective.js
```diff
@@ -939,11 +941,23 @@ tie.directive('learnerView', [function() {
...
 
+        /**
+         * Clears the feedback in the window, and deletes the feedback
+         * from local storage for the current question.
+         */
+        $scope.resetFeedback = function() {
+          ConversationLogDataService.clear();
+          LocalStorageService.clearLocalStorageFeedback(
+            $scope.currentQuestionId, language);
+        }
```
seanlip: Nit: missing semicolon

Issue #467, opened 2017-12-27
User can submit multiple code requests before feedback is received


rabidbit:
If a user clicks "I think I'm done" (submit code) button repeatedly before a feedback balloon is displayed to the user, the code submits stack...

<img width="1269" alt="multiple-submits" src="https://user-images.githubusercontent.com/25493966/34388187-2344f302-eae7-11e7-866c-9f8d92d4dffe.png">

We should disable code submits until the corresponding feedback is displayed.

seanlip:
Does this mean stopping the user from submitting at all during that time
(eg graying out the submit buttons)?

Or alternatively perhaps we could just not show feedback for requests that
get interrupted (ie in your screenshot, show only one feedback balloon
corresponding to the most recent submission).

>

rabidbit:
I think disabling the button (gray out) would work.. Alternatively, we could just ignore subsequent clicks until the feedback is displayed (same net effect as disabling the button except that we don't change the button's appearance or the user's ability to click it). I am leaning towards the latter to minimize the number of UI elements and UI behaviors we need to maintain.

seanlip:
Fixed in #490, thanks @theresa-lee!

PR #490, opened 2018-01-09
Disable button if there is feedback pending

theresa-lee:
Following discussion in #467, the button is now disabled (but not visually) to the user when there is feedback pending. I can make CSS changes to the button but it looked like @rabidbit preferred to have it basically be disabled but not visibly so.


rabidbit:
LGTM

PR #492, opened 2018-01-09
Increase contrast on some elements, add some accessibility attrs.

seanlip:


rabidbit:
LGTM, deferring to theresa-lee opinion

theresa-lee:
Can we use #424242 instead of the black/#111? I find that having completely black on white looks a little intense, and this color is part of the black/grey scale of suggested Material colors. I checked the contrast and it's over the ratio that is required for accessibility!

theresa-lee:
Actually, you might want the title to stand out more so #212121 for the title and #424242 for the line numbers?

seanlip:
Updated colours per @theresa-lee's suggestion, and also added aria-live attribute to announce the feedback messages as they appear. Thanks for the review!

PR #493, opened 2018-01-10
Added support for # comments in the global scope + made it more clear that helper functions are A-OK

eyurko:


client/question/services/code_evaluators/PythonPrereqCheckService.js
```diff
@@ -107,7 +107,8 @@ tie.factory('PythonPrereqCheckService', [
       for (var i = 0; i < codeLines.length; i++) {
         var line = rightTrim(codeLines[i]);
         var VALID_LINE_PATTERN = /(^\s+)|(^def)|(^import)/;
-        if (line.search(VALID_LINE_PATTERN) === -1 && line !== '') {
+        if (line.search(VALID_LINE_PATTERN) === -1 && line !== '' &&
+            !line.startsWith('#')) {
```
seanlip: Think it's worth changing this additional check to "first non-space char on line is a '#'"?

client/question/services/FeedbackGeneratorService.js
```diff
@@ -781,7 +781,8 @@ tie.factory('FeedbackGeneratorService', [
           feedback = FeedbackObjectFactory.create(
             FEEDBACK_CATEGORIES.FAILS_GLOBAL_CODE_CHECK);
           feedback.appendTextParagraph([
-            'Please keep your code within the existing predefined functions',
+            'Please keep your code within the existing predefined functions ',
+            'or define your own helper functions if you need to',
```
seanlip: Add space at end of this line. Ditto below.

rabidbit:
LGTM

client/question/services/code_evaluators/PythonPrereqCheckService.js
seanlip: Sorry, ignore this -- thinko on my part. It's caught in VALID_LINE_PATTERN, and those comments would be in a different scope anyway.

PR #494, opened 2018-01-10
Fix #476: change 'code history' modal title.

### LINKED ISSUE #476: "Code History" modal title may not be clear
This may need more data points, but I'm not sure that "Code History" is a good title for the modal that appears on clicking the link in a previous code submission. To me, it implies a history of _multiple_ code submissions. Perhaps changing the header text to "Previous Submission" might be better?

seanlip:


client/question/components/CodeSnippetDirective.js
```diff
@@ -65,7 +65,7 @@ tie.directive('codeSnippet', [function() {
          */
         $scope.openCodeModal = function() {
           MonospaceDisplayModalService.showModal(
-            'Code History', $scope.snippetLines);
+            'Previous Submission', $scope.snippetLines);
```
rabidbit: I was discussing this with eyurko awhile back and he mentioned the likelihood that users (students in particular) will hesitate when they see "submit". How about "Previous Code" or something like that?

rabidbit: Also consistent with us moving away from the "Submit" button.

seanlip: "Previous Code" sounds fine to me. I'll change that. Thanks!

seanlip:
Changes done -- PTAL.

rabidbit:
LGTM

Issue #469, opened 2017-12-27
Block "Next" Arrow Appearing Until Feedback Window Updates


eyurko:
Repro steps:

Open "Reverse Words" problem
Complete problem

The "Next" arrow will appear before the feedback window updates -- this should likely be blocked on the feedback window updating.

seanlip:
I can't repro this, but it might be symptomatic of another bug. When I finish the question, I get the following:

![screenshot from 2018-01-10 15-13-49](https://user-images.githubusercontent.com/10575562/34800507-ec82e08a-f618-11e7-9d24-e7e24574fd35.png)

The "I think I'm done" button is still clickable, and I can still edit the code and submit. Presumably this isn't the intended behaviour...? /cc @rabidbit

eyurko:
Yeah the arrow is gone

--
Eric Yurko

seanlip:
It seems to have been removed in https://github.com/google/tie/pull/484

Happy to make the appropriate fixes, but I think we need a better spec for what the "complete question" experience is supposed to be.

rabidbit:
This is now WAI per #482 and #484.

We currently only have one question and nowhere else for the user to go. So, the indication that the user completed the question is in the feedback, where it makes sense to look for validation. The user can continue playing with the app and question if they wish (perhaps we can add "you can continue to play with this question, if you wish" sentence).

The open TODO for a future PR is to create a cake (a visual reward for completing the question). This could be in the form of a big emoji or sticker in the feedback speech balloon an overlay over the coding window. I was thinking in the speech balloon (like in real chat apps), but I have no strong feelings (we could implement at least one of these and user test it).

seanlip:
Thanks @rabidbit. Filed #496 to clarify the text. Will file a new issue for the third paragraph; closing this one.

PR #496, opened 2018-01-11
Add final sentence to congratulatory feedback. Fix regression caused by #492.

seanlip:
Added final sentence to congratulatory feedback as suggested in [this comment](https://github.com/google/tie/issues/469#issuecomment-356783014).

Also fixed a regression caused by #492 that caused _all_ balloons to glow yellow when a new piece of code is submitted.


rabidbit:
LGTM

Issue #505, opened 2018-01-17
Correct state is not stored and does not survive page reload


rabidbit:
Reproduce:
1. Solve question
2. Reload page
Expected: congratulatory feedback is given
Observed: last known failure feedback is given

This may confuse the user into thinking the code in the coding window is incorrect.

rabidbit:
This was largely addressed in #509. In testing it, it seems a user could edit the code before reload, then the feedback and code will still not be in sync.

Suggest in a separate PR that we display the last user code submit before the last feedback. This would provide the user context for the feedback, even if the feedback no longer applies to the current code.

theresa-lee:
Sounds good, I created a new issue #510

PR #509, opened 2018-01-19
Feedback is now stored after question completion

theresa-lee:
Looks like we just weren't saving the feedback after completing the problem!


rabidbit:
Thanks theresa-lee. This does addresses #505. Though when I think about it, I can see the user modifying the code and not clicking "I think I'm done" before reloading the page, which would make the last feedback invalid with the current code.

How difficult would it be to store and display the last user code submit before the last feedback? This would provide the user context for the feedback, even if the feedback no longer applies to the current code.

rabidbit:
I think displaying the last code submission on reload can be a separate PR. This one LGTM.

PR #518, opened 2018-02-08
Remove dummy dropdown option from theme selector.

seanlip:
This addresses an accessibility bug where the dropdown menu seems to have three options (but it only has two).

I followed [this approach](https://webaim.org/techniques/forms/controls#select). However, note that this entails UI changes -- see bottom of screenshot below, in particular the added "Theme:" label. Please let me know what you think -- I'm happy to make any changes.

screenshot from 2018-02-08 13-42-32

### DIFF
--- a/client/question/components/LearnerViewDirective.js
+++ b/client/question/components/LearnerViewDirective.js
@@ -865,11 +865,11 @@ tie.directive('learnerView', [function() {
          */
         $scope.changeTheme = function(newTheme) {
           $scope.pulseAnimationEnabled = false;
-          if (newTheme === 'Dark') {
+          if (newTheme === THEME_NAME_DARK) {
             $scope.isInDarkMode = true;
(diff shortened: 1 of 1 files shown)


rabidbit:
What if we replace "theme" to "Select Theme". I wonder if the language would help the user understand that the other two are the actual options. If this will not address the issue, happy to discuss offline. In general, I am not a fan of the Label: [ button ] layout as it takes up more space and makes it a bit cluttered.

seanlip:
@rabidbit Changes done, PTAL. Thanks!

rabidbit:
Thanks Sean. Toggling the theme does not seem to do anything.

seanlip:
Whoops, sorry @rabidbit -- my bad, I missed one change. PTAL again.

rabidbit:
LGTM. Thanks!

PR #519, opened 2018-02-08
Fix a P1 accessibility bug: screen reader announces background contents even though a modal is displayed.

seanlip:
Please note that this change has one consequence for sighted users: the contents of the question area are wiped _before_ the modal animates down. (This wiping is needed because otherwise the screen reader will read background components that are not visible to sighted users.) However, there is some code complexity involved in introducing an additional timeout before wiping the background, so I wanted to check if the current implementation looks OK from a UI perspective.


rabidbit:
Thanks for working on this, seanlip. From a UI/UX standpoint, I think it would be better to not make content visibly disappear; the reason we have a modal that float in is to assure the user that the feedback didn't just disappear. This looks like feedback does disappear in addition to a modal that floats in. Also, when dismissed, the modal suddenly disappears rather than float back up, which makes it look glitchy/buggy/janky.

seanlip:
Superseded by #532. It took a while to figure out how to handle this properly because the CSS animations and timeouts lead to race conditions, so I eventually went with a different approach.

Issue #414, opened 2017-10-03
User study - Include line number in language detection


rabidbit:
When issue #413 occurred, it was very hard to know what part of the code was being referenced. Adding a line number to the feedback would help the user better understand where the problem is and also helps to better troubleshoot a potential bug.

seanlip:
Hi @rabidbit -- here's what this currently looks like on the master branch:

![screenshot from 2018-01-10 13-45-02](https://user-images.githubusercontent.com/10575562/34797286-c857056c-f60c-11e7-9889-2d4bc7856e46.png)

Is the aim to get it to look like this? (See first dialog balloon in the screenshot below)

![screenshot from 2018-01-10 13-46-46](https://user-images.githubusercontent.com/10575562/34797301-d6c6f3b4-f60c-11e7-8dab-2480399dfc8d.png)

If not, please could you clarify any changes to the desired behaviour? Thanks!

seanlip:
Btw, I thought about highlighting the lines in question, but one thing that's tricky about this is that some of the caught errors are not restricted to a single line -- e.g. do/while, switch, etc. So we need to figure out a UI for the multiline case too.

I'd suggest the UI from the previous comment, perhaps with the added text changed to "(See lines 2-3 of the code)", but let me know what you think!

rabidbit:
Highlighting the line is the preferred way, as it's easier to find in a glance and is consistent with syntax errors. Can we just highlight the first line and let the user figure out the rest? This would already be a big help.

Also, this may be a chance for us to point to documentation on the library in question. One of the feedback we got from user testing is that there seems to be a lack of pointers to external documentation that could help. This need not be done in this PR, but in a future PR, if we point to the python comparison doc, for example (and this will be included in the python primer), I don't think we need to worry too much about highlighting all the lines that are relevant to the wrong language detection message. The user should be clever enough to fix this issue, as expected in real life. Highlighting the first line will help eliminate unnecessary scanning of the code, since we are not really testing the user on how python does comparisons.

seanlip:
Hi @michaelrobbe, just to check, can we close this? (If so, please feel free to do that; I wasn't sure.)

michaelrobbe:
Fixed in #557. Closing

PR #520, opened 2018-02-08
Add line numbers to PrereqCheck Obj, to display on feedback (#414)

michaelrobbe:
Expands the PrereqCheck to find the error line numbers, and pass it to the feedback object. To eventually highlight the line, and append to the feedback bubble.

This is PR 1 of 2, loop through the student code, find the error and pass it to the PrereqCheckFailureObjectFactory.

### DIFF
--- a/client/question/domain/PrereqCheckErrorObjectFactory.js
+++ b/client/question/domain/PrereqCheckErrorObjectFactory.js
@@ -30,7 +30,8 @@ tie.factory('PrereqCheckErrorObjectFactory', [
      *
      * @constructor
      */
-    var PrereqCheckError = function(errorName, errorLineNumber) {
+    var PrereqCheckError = function(errorName, errorLineNumber,
+      errorColumnNumber) {
       /**
        * String standing for key in WRONG_LANGUAGE_ERRORS that correlates to
        * error that triggered.
@@ -49,6 +50,14 @@ tie.factory('PrereqCheckErrorObjectFactory', [
        * @private
        */
       this._errorLineNumber = errorLineNumber;
+
+      /**
+       * Column number that correlates to error that triggered.
+       *
+       * @type {integer}
+       * @private
+       */
+      this._errorColumnNumber = errorColumnNumber;
 
       /**
        * Indicates what type of failure occurred. Defaults to the only errorType
@@ -96,14 +105,35 @@ tie.factory('PrereqCheckErrorObjectFactory', [
       this._errorLineNumber = newLineNumber;
     };
 
+    /**
+     * A getter for the _errorColumnNumber property.
+     *
+     * @returns {string}
+     */
+    PrereqCheckError.prototype.getErrorColumnNumber = function() {
+      return this._errorColumnNumber;
+    };
+
+    /**
+     * A setter for the _errorColumnNumber property.
+     *
+     * @param {integer} errorLine to set the _errorColumnNumber property to.
+     */
+    PrereqCheckError.prototype.setErrorColumnNumber = function(
+      newColumnNumber) {
+      this._errorColumnNumber = newColumnNumber;
+    };
+
     // Static class methods.
     /**
      * Returns a new PrereqCheckError object.
      *
      * @returns {PrereqCheckError}
      */
-    PrereqCheckError.create = function(errorName, errorLineNumber) {
-      return new PrereqCheckError(errorName, errorLineNumber);
+    PrereqCheckError.create = function(
+        errorName, errorLineNumber, errorColumnNumber) {
+      return new PrereqCheckError(
+        errorName, errorLineNumber, errorColumnNumber);
     };
 
     return PrereqCheckError;
--- a/client/question/domain/PrereqCheckErrorObjectFactory.js
+++ b/client/question/domain/PrereqCheckErrorObjectFactory.js
@@ -30,8 +30,8 @@ tie.factory('PrereqCheckErrorObjectFactory', [
      *
      * @constructor
      */
-    var PrereqCheckError = function(errorName, errorLineNumber,
-      errorColumnNumber) {
+    var PrereqCheckError = function(
+      errorName, errorLineNumber, errorColumnNumber) {
       /**
        * String standing for key in WRONG_LANGUAGE_ERRORS that correlates to
        * error that triggered.
--- a/client/question/domain/PrereqCheckErrorObjectFactory.js
+++ b/client/question/domain/PrereqCheckErrorObjectFactory.js
@@ -1,0 +1,104 @@
+// Copyright 2017 The TIE Authors. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS-IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+/**
+ * @fileoverview Factory for creating a PrereqCheckError of a student's TIE
+ * session.
+ */
+
+tie.factory('PrereqCheckErrorObjectFactory', [
+  function() {
+    /**
+     * PrereqCheckError objects are used to store information related to Prereq
+     errors.
+     */
+
+    /**
+     * Constructor for PrereqCheckError.
+     *
+     * @constructor
+     */
+    var PrereqCheckError = function(errorName, errorLineNumber) {
+      /**
+       * String standing for key in WRONG_LANGUAGE_ERRORS that correlates to
+       * error that triggered.
+       * Should be null if error type is not 'wrong language.'
+       *
+       * @type {string | null}
+       * @private
+       */
+      this._errorName = errorName;
+
+      /**
+       * Line number that correlates to error that triggered.
+       * Should be null if error type does not supply a line number.
+       *
+       * @type {integer | null}
+       * @private
+       */
+      this._errorLineNumber = errorLineNumber;
+    };
+
+    /**
+     * A getter for the _errorName property.
+     *
+     * @returns {string}
+     */
+    PrereqCheckError.prototype.getErrorName = function() {
+      return this._errorName;
+    };
+
+    /**
+     * A setter for the _errorName property.
+     *
+     * @param {string} newName The key to set the _errorName property to.
+     */
+    PrereqCheckError.prototype.setErrorName = function(newName) {
+      this._errorName = newName;
+    };
+
+    /**
+     * A getter for the _errorLineNumber property.
+     *
+     * @returns {string}
+     */
+    PrereqCheckError.prototype.getErrorLineNumber = function() {
+      return (this._errorLineNumber < 0) ? null : this._errorLineNumber;
+    };
+
+    /**
+     * A setter for the _errorLineNumber property.
+     *
+     * @param {integer} errorLine to set the _errorLineNumber property to.
+     */
+    PrereqCheckError.prototype.setErrorLineNumber = function(newLineNumber) {
+      this._errorLineNumber = newLineNumber;
+    };
+
+    // Static class methods.
+    /**
+     * Returns a new PrereqCheckError object.
+     *
+     * @returns {PrereqCheckError}
+     */
+    PrereqCheckError.create = function(errorName, errorLineNumber) {
+      if (errorLineNumber < 0 && errorName === 'notOp') {
+        return null;
+      }
+      return new PrereqCheckError(errorName, errorLineNumber);
+    };
+
+    return PrereqCheckError;
+  }
+]);
--- a/client/question/domain/PrereqCheckErrorObjectFactory.js
+++ b/client/question/domain/PrereqCheckErrorObjectFactory.js
@@ -44,9 +44,8 @@ tie.factory('PrereqCheckErrorObjectFactory', [
 
       /**
        * Line number that correlates to error that triggered.
-       * Should be null if error type does not supply a line number.
        *
-       * @type {integer | null}
+       * @type {integer}
        * @private
        */
       this._errorLineNumber = errorLineNumber;
--- a/client/question/domain/PrereqCheckErrorObjectFactory.js
+++ b/client/question/domain/PrereqCheckErrorObjectFactory.js
@@ -18,7 +18,8 @@
  */
 
 tie.factory('PrereqCheckErrorObjectFactory', [
-  function() {
+  'PREREQ_CHECK_TYPE_WRONG_LANG',
+  function(PREREQ_CHECK_TYPE_WRONG_LANG) {
     /**
      * PrereqCheckError objects are used to store information related to Prereq
      errors.
@@ -48,6 +49,15 @@ tie.factory('PrereqCheckErrorObjectFactory', [
        * @private
        */
       this._errorLineNumber = errorLineNumber;
+
+      /**
+       * Indicates what type of failure occurred. Defaults to the only errorType
+       * currently tracked.
+       *
+       * @type {string}
+       * @private
+       */
+      this._errorType = PREREQ_CHECK_TYPE_WRONG_LANG;
     };
 
     /**
@@ -74,7 +84,7 @@ tie.factory('PrereqCheckErrorObjectFactory', [
      * @returns {string}
      */
     PrereqCheckError.prototype.getErrorLineNumber = function() {
-      return (this._errorLineNumber < 0) ? null : this._errorLineNumber;
+      return this._errorLineNumber;
     };
 
     /**
@@ -93,9 +103,6 @@ tie.factory('PrereqCheckErrorObjectFactory', [
      * @returns {PrereqCheckError}
      */
     PrereqCheckError.create = function(errorName, errorLineNumber) {
-      if (errorLineNumber < 0 && errorName === 'notOp') {
-        return null;
-      }
       return new PrereqCheckError(errorName, errorLineNumber);
     };
 
--- a/client/question/domain/PrereqCheckErrorObjectFactorySpec.js
+++ b/client/question/domain/PrereqCheckErrorObjectFactorySpec.js
@@ -24,27 +24,36 @@ describe('PrereqCheckErrorObjectFactory', function() {
     PrereqCheckErrorObjectFactory = $injector.get(
       'PrereqCheckErrorObjectFactory');
   }));
-  var errorName = 'notOp';
-  var errorLineNumber = 2;
+  var error = {
+    errorName: 'notOp',
+    errorLineNumber: 2
+  };
 
   describe('create', function() {
     it('should generate an object with name and line number', function() {
-      var TestObject = {
-        _errorName: 'notOp',
-        _errorLineNumber: 2,
-        _errorType: 'wrongLang'
-      };
-      var PrereqCheckError = PrereqCheckErrorObjectFactory.create(
-        errorName, errorLineNumber);
+      var newErrorName = 'andOp';
+      var newErrorLineNumber = 35;
+      var PrereqCheckError;
+      PrereqCheckError = PrereqCheckErrorObjectFactory.create(
+        error.errorName, error.errorLineNumber);
 
-      expect(PrereqCheckError).toEqual(jasmine.objectContaining(TestObject));
+      expect(PrereqCheckError.getErrorName()).toEqual(error.errorName);
+      expect(PrereqCheckError.getErrorLineNumber()).toEqual(
+        error.errorLineNumber);
+
+      PrereqCheckError.setErrorName(newErrorName);
+      PrereqCheckError.setErrorLineNumber(newErrorLineNumber);
+
+      expect(PrereqCheckError.getErrorName()).toEqual(newErrorName);
+      expect(PrereqCheckError.getErrorLineNumber()).toEqual(newErrorLineNumber);
     });
   });
 
   describe('getErrorName', function() {
     it('should return string from error object', function() {
-      var PrereqCheckError = PrereqCheckErrorObjectFactory.create(
-        errorName, errorLineNumber);
+      var PrereqCheckError;
+      PrereqCheckError = PrereqCheckErrorObjectFactory.create(
+        error.errorName, error.errorLineNumber);
       expect(PrereqCheckError.getErrorName()).toEqual('notOp');
     });
   });
--- a/client/question/domain/PrereqCheckErrorObjectFactorySpec.js
+++ b/client/question/domain/PrereqCheckErrorObjectFactorySpec.js
@@ -31,7 +31,8 @@ describe('PrereqCheckErrorObjectFactory', function() {
     it('should generate an object with name and line number', function() {
       var TestObject = {
         _errorName: 'notOp',
-        _errorLineNumber: 2
+        _errorLineNumber: 2,
+        _errorType: 'wrongLang'
       };
       var PrereqCheckError = PrereqCheckErrorObjectFactory.create(
         errorName, errorLineNumber);
--- a/client/question/domain/PrereqCheckErrorObjectFactorySpec.js
+++ b/client/question/domain/PrereqCheckErrorObjectFactorySpec.js
@@ -1,0 +1,69 @@
+// Copyright 2017 The TIE Authors. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS-IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+/**
+ * @fileoverview Unit tests for PrereqCheckErrorObjectFactory domain objects.
+ */
+
+describe('PrereqCheckErrorObjectFactory', function() {
+  var PrereqCheckErrorObjectFactory;
+
+  beforeEach(module('tie'));
+  beforeEach(inject(function($injector) {
+    PrereqCheckErrorObjectFactory = $injector.get(
+      'PrereqCheckErrorObjectFactory');
+  }));
+  var errorName = 'notOp';
+  var errorLineNumber = 2;
+
+  describe('create', function() {
+    it('should generate an object with name and line number', function() {
+      var TestObject = {
+        _errorName: 'notOp',
+        _errorLineNumber: 2
+      };
+      var PrereqCheckError = PrereqCheckErrorObjectFactory.create(
+        errorName, errorLineNumber);
+
+      expect(PrereqCheckError).toEqual(jasmine.objectContaining(TestObject));
+    });
+  });
+
+  describe('getErrorName', function() {
+    it('should return string from error object', function() {
+      var PrereqCheckError = PrereqCheckErrorObjectFactory.create(
+        errorName, errorLineNumber);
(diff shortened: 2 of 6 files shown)


michaelrobbe:
Signed the CLA

client/question/services/code_evaluators/PythonPrereqCheckService.js
```diff
@@ -65,18 +65,28 @@ tie.factory('PythonPrereqCheckService', [
...
 
         if (error.allowMultiline) {
           if (code.search(regexp) !== -1) {
-            return error.errorName;
+            var errorFirstLine = code.search(regexp);
+            for (var j = 0; errorFirstLine > 0; j++) {
+              var sub = rawCodeArray[j].length;
+              errorFirstLine -= sub;
+            }
+            error.errorLine = errorFirstLine;
```
seanlip: Nope -- please treat `error` as readonly, and don't set new attrs on it or pass it around. It's part of an overall constant that anything can reference, so modifying it isn't a good idea. Instead, I would recommend creating a new type of domain object to wrap the error information and send it back.

client/question/services/code_evaluators/PythonPrereqCheckService.js
```diff
@@ -65,18 +65,28 @@ tie.factory('PythonPrereqCheckService', [
...
+            return error;
           }
         } else {
-          for (var j = 0; j < codeLines.length; j++) {
-            if (codeLines[j].search(regexp) !== -1) {
-              return error.errorName;
+          for (var k = 0; k < codeLines.length; k++) {
+            if (codeLines[k].search(regexp) !== -1) {
+              error.errorLine = rawCodeArray.findIndex(function(el) {
+                return el === codeLines[k];
```
seanlip: I think, in order to do this properly, we should introduce a mapping from rawCodeArray to codeLines? Multiple lines may have the same code.

client/question/domain/PrereqCheckFailureObjectFactory.js
```diff
@@ -86,6 +86,15 @@ tie.factory('PrereqCheckFailureObjectFactory', [
...
       this._wrongLangKey = wrongLangKey;
+
+      /**
+       * Line number that correlates to error that triggered.
+       * Should be null if error type does not supply a line number
+       *
+       * @type {integer | null}
+       * @private
+       */
+      this._errorLine = errorLine;
```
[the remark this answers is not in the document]
michaelrobbe: Done.

client/question/services/code_evaluators/PythonPrereqCheckService.js
```diff
@@ -65,18 +65,28 @@ tie.factory('PythonPrereqCheckService', [
...
 
+      var rawCodeArray = code.split('\n');
       for (var i = 0; i < WRONG_LANGUAGE_ERRORS.python.length; i++) {
         var error = WRONG_LANGUAGE_ERRORS.python[i];
         var regexp = new RegExp(error.regExString);
 
         if (error.allowMultiline) {
           if (code.search(regexp) !== -1) {
-            return error.errorName;
+            var errorFirstLine = code.search(regexp);
```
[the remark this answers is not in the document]
michaelrobbe: Reworked the error line check, Done.

client/question/services/code_evaluators/PythonPrereqCheckService.js
michaelrobbe: Domain Object written up to store errorName and errorLineNumber.
Done.

client/question/services/code_evaluators/PythonPrereqCheckService.js
michaelrobbe: Error returns at the first instance found. At this point, we are only concerned with one error at a time.
We can revisit if we want to show all errors at once.
Done.

seanlip:
Hi @michaelrobbe, is this PR ready for review? It seems to be failing the Karma tests on Travis.

client/question/domain/PrereqCheckErrorObjectFactory.js
```diff
@@ -0,0 +1,104 @@
...
+     * PrereqCheckError objects are used to store information related to Prereq
+     errors.
+     */
+
+    /**
+     * Constructor for PrereqCheckError.
+     *
+     * @constructor
+     */
+    var PrereqCheckError = function(errorName, errorLineNumber) {
```
seanlip: If this is just going to be scoped to wrong language errors then the domain object should be WrongLanguageCheckError or similar.

Alternatively if this is going to be used for other types of errors as well then perhaps we should have a this._errorType that's set to ERROR_TYPE_WRONG_LANGUAGE or null (for now). That would make it easier to add other error types in the future.

client/question/domain/PrereqCheckErrorObjectFactory.js
```diff
@@ -0,0 +1,104 @@
...
+      this._errorName = newName;
+    };
+
+    /**
+     * A getter for the _errorLineNumber property.
+     *
+     * @returns {string}
+     */
+    PrereqCheckError.prototype.getErrorLineNumber = function() {
+      return (this._errorLineNumber < 0) ? null : this._errorLineNumber;
```
seanlip: What if this._errorLineNumber is null?

Why would this._errorLineNumber be < 0?

client/question/domain/PrereqCheckErrorObjectFactory.js
```diff
@@ -0,0 +1,104 @@
...
+    };
+
+    // Static class methods.
+    /**
+     * Returns a new PrereqCheckError object.
+     *
+     * @returns {PrereqCheckError}
+     */
+    PrereqCheckError.create = function(errorName, errorLineNumber) {
+      if (errorLineNumber < 0 && errorName === 'notOp') {
```
seanlip: I don't think I understand this, it might need an explanatory comment. What's special about 'notOp'?

client/question/domain/PrereqCheckErrorObjectFactorySpec.js
```diff
@@ -0,0 +1,69 @@
...
+  describe('getErrorName', function() {
+    it('should return string from error object', function() {
+      var PrereqCheckError = PrereqCheckErrorObjectFactory.create(
+        errorName, errorLineNumber);
+      expect(PrereqCheckError.getErrorName()).toEqual('notOp');
+    });
+  });
+
+  describe('nullMultilineReturn', function() {
+    var negativeErrorLineNumber = -5;
```
seanlip: When will this case occur in practice?

client/question/domain/PrereqCheckFailureObjectFactory.js
```diff
@@ -243,6 +252,25 @@ tie.factory('PrereqCheckFailureObjectFactory', [
       this._wrongLangKey = newKey;
     };
 
+    /**
+     * A getter for the _errorLineNumber property.
+     *
+     * @returns {string}
+     */
+    PrereqCheckFailure.prototype.geterrorLineNumber = function() {
```
seanlip: error --> Error

client/question/domain/PrereqCheckFailureObjectFactory.js
```diff
@@ -243,6 +252,25 @@ tie.factory('PrereqCheckFailureObjectFactory', [
...
+      return this._errorLineNumber;
+    };
+
+    /**
+     * A setter for the _errorLineNumber property.
+     *
+     * @param {integer} errorLine to set the _errorLineNumber property to.
+     */
+    PrereqCheckFailure.prototype.seterrorLineNumber = function(
```
seanlip: Ditto

client/question/services/code_evaluators/PythonPrereqCheckService.js
```diff
@@ -65,20 +67,31 @@ tie.factory('PythonPrereqCheckService', [
...
+          // Match found, Subtract each line.length until error line pops.
+          for (var j = 0; firstErrorCharIndex > 0; j++) {
+            var sub = rawCodeArray[j].length;
+            firstErrorCharIndex -= sub;
           }
-        } else {
-          for (var j = 0; j < codeLines.length; j++) {
-            if (codeLines[j].search(regexp) !== -1) {
-              return error.errorName;
+          errorLine = firstErrorCharIndex;
```
seanlip: Why are we setting errorLine to firstErrorCharIndex? Not sure I understand the logic here...

client/question/services/code_evaluators/PythonPrereqCheckService.js
```diff
@@ -65,20 +67,31 @@ tie.factory('PythonPrereqCheckService', [
...
+        var errorLine = null;
 
-        if (error.allowMultiline) {
-          if (code.search(regexp) !== -1) {
-            return error.errorName;
+        if (firstErrorCharIndex !== -1) {
+          // Match found, Subtract each line.length until error line pops.
+          for (var j = 0; firstErrorCharIndex > 0; j++) {
+            var sub = rawCodeArray[j].length;
+            firstErrorCharIndex -= sub;
```
seanlip: Off-by-one? (Note newline char at end of each line.)

client/question/services/code_evaluators/PythonPrereqCheckService.js
```diff
@@ -65,20 +67,31 @@ tie.factory('PythonPrereqCheckService', [
...
           }
-        } else {
-          for (var j = 0; j < codeLines.length; j++) {
-            if (codeLines[j].search(regexp) !== -1) {
-              return error.errorName;
+          errorLine = firstErrorCharIndex;
+          // Walk through codelines, if regexp matches, return matched line.
+          for (var k = 0; k < codeLines.length; k++) {
+            if (codeLines[k].search(regexp) !== -1) {
```
seanlip: Might be superfluous? (since you've already done a search above)

client/question/domain/PrereqCheckErrorObjectFactorySpec.js
michaelrobbe: It won't, its a reflection of a bug while testing multiline errors, where I got negative numbers. Removed.
Done.

client/question/domain/PrereqCheckFailureObjectFactory.js
michaelrobbe: Done.

client/question/domain/PrereqCheckFailureObjectFactory.js
michaelrobbe: Done.

client/question/services/code_evaluators/PythonPrereqCheckService.js
michaelrobbe: Updated to subtract +1, tested on long code snippets successfully.
Done.

client/question/services/code_evaluators/PythonPrereqCheckService.js
michaelrobbe: Removed this code, now the search is done by character position, instead of line-by-line.
Done

client/question/services/code_evaluators/PythonPrereqCheckService.js
michaelrobbe: Corrected. It should be the index variable, not firstErrorCharIndex.
Done.

michaelrobbe:
To allow for multiline detection, I pre-process the code obscuring all code strings before running wrong language checks.

client/question/domain/PrereqCheckErrorObjectFactorySpec.js
```diff
@@ -0,0 +1,51 @@
...
+  var errorLineNumber = 2;
+
+  describe('create', function() {
+    it('should generate an object with name and line number', function() {
+      var TestObject = {
+        _errorName: 'notOp',
+        _errorLineNumber: 2,
+        _errorType: 'wrongLang'
+      };
+      var PrereqCheckError = PrereqCheckErrorObjectFactory.create(
```
seanlip: var prereqCheckError (it's an instance, not a class)

client/question/domain/PrereqCheckErrorObjectFactorySpec.js
```diff
@@ -0,0 +1,51 @@
...
+        errorName, errorLineNumber);
+
+      expect(PrereqCheckError).toEqual(jasmine.objectContaining(TestObject));
```
seanlip: Hm, feels a bit indirect. Maybe expect(prereqCheckError.getErrorName()).toEqual(...), etc.? In general all functions in the object's "public API" should be tested, so I would recommend focusing your tests on that (and making sure you get full coverage of the different methods).

client/question/domain/PrereqCheckFailureObjectFactory.js
seanlip: Not done.

client/question/domain/PrereqCheckFailureObjectFactory.js
seanlip: No, not done.

client/question/services/code_evaluators/PythonPrereqCheckService.js
```diff
@@ -53,6 +55,33 @@ tie.factory('PythonPrereqCheckService', [
...
+     * to filler character "x".
+     * This is to preserve character length for error checks.
+     *
+     * @param {string} [code] The code to convert
+     * @return {string} code where all 'foo' & "bar" converted to xxxxx & xxxxx
+     */
+    var obscureStringCharacters = function(code) {
+      var regexp = new RegExp(/'([^']*)'|"([^"]*)"/, 'g');
+      var matches = [];
+      var obscureCode = '';
```
seanlip: nit: obscuredCode

client/question/services/code_evaluators/PythonPrereqCheckService.js
```diff
@@ -53,6 +55,33 @@ tie.factory('PythonPrereqCheckService', [
...
+      var pointer = 0;
+
+      while ((matches = regexp.exec(code)) !== null) {
```
seanlip: This is a bit clumsy IMO; consider looking into https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/replace

client/question/services/code_evaluators/PythonPrereqCheckService.js
```diff
@@ -63,22 +92,27 @@ tie.factory('PythonPrereqCheckService', [
      *    WRONG_LANGUAGE_ERRORS.
      */
     var detectAndGetWrongLanguageType = function(code) {
-      var codeLines = getNonStringLines(code);
+      var rawCodeArray = obscureStringCharacters(code).split('\n');
```
seanlip: Is it possible for a string to contain a (non-escaped) newline character? If so, would that mess up the line numbers?

client/question/services/code_evaluators/PythonPrereqCheckService.js
```diff
@@ -63,22 +92,27 @@ tie.factory('PythonPrereqCheckService', [
...
-          }
-        } else {
-          for (var j = 0; j < codeLines.length; j++) {
-            if (codeLines[j].search(regexp) !== -1) {
-              return error.errorName;
-            }
+        // Lookup character location of error
+        var firstErrorCharIndex = code.search(regexp);
+        if (firstErrorCharIndex !== -1) {
+          var errorLine = null;
```
seanlip: Consider calling this firstErrorLineNumber (and the above as firstErrorColumnNumber) for ease of understanding.

Also this should perhaps be initialized to 0 (and incremented every time the next for loop runs)? That way there's this sense of "it's on line 0 and column C", which means "it's on line 1 and column (C - length-of-line-0 - 1)", etc.

client/question/services/code_evaluators/PythonPrereqCheckService.js
```diff
@@ -53,6 +55,33 @@ tie.factory('PythonPrereqCheckService', [
...
 
+    /**
+     * Processes studentCode and converts all single quote + double quote str
+     * to filler character "x".
+     * This is to preserve character length for error checks.
+     *
+     * @param {string} [code] The code to convert
+     * @return {string} code where all 'foo' & "bar" converted to xxxxx & xxxxx
+     */
+    var obscureStringCharacters = function(code) {
```
seanlip: getObscuredCode might be better. The issue is that "obscure" could serve as a verb or an adjective, so it's not immediately clear from reading the name whether obscureStringCharacters is a method or a set/list of obscure string chars.

client/question/services/code_evaluators/PythonPrereqCheckServiceSpec.js
```diff
@@ -493,7 +493,33 @@ describe('PythonPrereqCheckService', function() {
       ].join('\n');
       var prereqFailureType =
         PythonPrereqCheckService.detectAndGetWrongLanguageType(code);
-      expect(prereqFailureType).toEqual(null);
+      expect(prereqFailureType).toEqual(
+        jasmine.objectContaining({
```
seanlip: See above comment.

client/question/services/code_evaluators/PythonPrereqCheckServiceSpec.js
```diff
@@ -493,7 +493,33 @@ describe('PythonPrereqCheckService', function() {
...
+          _errorName: 'notOp',
+          _errorLineNumber: 2,
+          _errorType: 'wrongLang'
+        }));
+    });
+  });
+
+  describe('obscureStringCharacters', function() {
```
seanlip: This method has complicated logic and needs way more tests. Strings within strings, multiple strings, strings of different types, single-and-double-quotes interlacing, etc.

You don't need to use actual "looks like code" strings, btw. Any strings will do.

client/question/domain/PrereqCheckErrorObjectFactorySpec.js
michaelrobbe: Rewritten.

client/question/services/code_evaluators/PythonPrereqCheckServiceSpec.js
michaelrobbe: Added new test conditions to flesh out functionality.

client/question/services/code_evaluators/PythonPrereqCheckService.js
michaelrobbe: Reasoning make sense, updated as getObscuredCode.

client/question/services/code_evaluators/PythonPrereqCheckServiceSpec.js
michaelrobbe: Simplified test, checking via functions, rather than a object.

michaelrobbe:
@seanlip Ready for review. Biggest changes, incorporating errorColumnNumber into the PrereqCheckError and PrereqCheckFailure Domain Objects.

client/question/domain/PrereqCheckErrorObjectFactory.js
```diff
@@ -0,0 +1,140 @@
...
+     errors.
+     */
+
+    /**
+     * Constructor for PrereqCheckError.
+     *
+     * @constructor
+     */
+    var PrereqCheckError = function(errorName, errorLineNumber,
+      errorColumnNumber) {
```
seanlip: Nit: when the contents of parens span multiple lines, break after the initial '(' instead.

client/question/services/code_evaluators/PythonPrereqCheckService.js
```diff
@@ -53,6 +55,28 @@ tie.factory('PythonPrereqCheckService', [
...
+    /**
+     * Processes studentCode and converts all escaped characters, then all
+     * single quote + double quote str to filler character "x".
+     * This is to preserve character length for error checks.
+     *
+     * @param {string} [code] The code to convert
+     * @return {string} code where all 'foo' & "bar" converted to xxxxx & xxxxx
+     */
+    var getObscuredCode = function(code) {
+      var escapeRegexp = new RegExp(/\\[\w,',",\\]/, 'g');
```
seanlip: Why do you need the commas in this regex?

client/question/services/code_evaluators/PythonPrereqCheckService.js
```diff
@@ -53,6 +55,28 @@ tie.factory('PythonPrereqCheckService', [
...
+      var regexp = new RegExp(/'([^']*)'|"([^"]*)"/, 'g');
```
seanlip: stringRegexp

client/question/services/code_evaluators/PythonPrereqCheckService.js
```diff
@@ -63,22 +87,31 @@ tie.factory('PythonPrereqCheckService', [
...
-              return error.errorName;
-            }
+        // Lookup character location of error
+        var firstErrorColumnNumber = obscuredCode.search(regexp);
+        if (firstErrorColumnNumber !== -1) {
+          var firstErrorLineNumber = 0;
+
+          // After regex matches an error, loop through codelines
+          for (var l = 0; firstErrorColumnNumber >= 0; l++) {
+            // Subtract l.length from firstErrorCharIndex.
```
seanlip: l is a number, so there's no l.length. I think you mean "Subtract the length of the corresponding line from firstErrorCharIndex" -- except that there's no firstErrorCharIndex either so you'll want to rename firstErrorColumnNumber to firstErrorCharIndex here.

client/question/services/code_evaluators/PythonPrereqCheckService.js
```diff
@@ -63,22 +87,31 @@ tie.factory('PythonPrereqCheckService', [
...
+            // At/below 0: index === linenumber. Subtract +1 for newline char.
+            firstErrorColumnNumber -= (rawCodeArray[l].length + 1);
+            firstErrorLineNumber++;
           }
+
+          firstErrorColumnNumber += rawCodeArray[
```
seanlip: At this point it's reasonable to say firstErrorColumnNumber = firstErrorCharIndex + ...

client/question/services/code_evaluators/PythonPrereqCheckServiceSpec.js
```diff
@@ -493,7 +493,70 @@ describe('PythonPrereqCheckService', function() {
       ].join('\n');
       var prereqFailureType =
         PythonPrereqCheckService.detectAndGetWrongLanguageType(code);
-      expect(prereqFailureType).toEqual(null);
+
+      expect(prereqFailureType.getErrorName()).toEqual('notOp');
+      expect(prereqFailureType.getErrorLineNumber()).toEqual(
```
seanlip: Here and below: tests should not be introspecting private properties. Instead, hardcode the expected value based on the input that is given, and test that the values of getErrorLineNumber(), getErrorColumnNumber(), etc. match what you expect them to be.

client/question/services/code_evaluators/PythonPrereqCheckServiceSpec.js
```diff
@@ -493,7 +493,70 @@ describe('PythonPrereqCheckService', function() {
...
+
+    });
+  });
+
+  describe('obscureStringCharacters', function() {
+    it('obscures escaped characters', function() {
+
+      var backslash = "\\\\";
+      var backslashedBackslash = "\\\\\\\\";
+      var singlequotes = '\\\'\\\'\\\'';
```
seanlip: singleQuotes, doubleQuotes

(Although, technically, these are constants, so BACKSLASH, SINGLE_QUOTES, etc.)

client/question/services/code_evaluators/PythonPrereqCheckServiceSpec.js
```diff
@@ -493,7 +493,70 @@ describe('PythonPrereqCheckService', function() {
...
+    it('obscures escaped characters, preserving character length', function() {
+
+      var varLine = 'test_var = \\\'1\\\' + \\\'1\\\'';
+      var obscuredVarLine = PythonPrereqCheckService.getObscuredCode(varLine);
+
+      expect(obscuredVarLine).toEqual("test_var = xx1xx + xx1xx");
+      expect(obscuredVarLine.length).toEqual(varLine.length);
+    });
+
+    it('obscure strings to test while preserving character length', function() {
```
seanlip: obscure --> obscures

character length --> line length

client/question/services/code_evaluators/PythonPrereqCheckServiceSpec.js
```diff
@@ -493,7 +493,70 @@ describe('PythonPrereqCheckService', function() {
...
+        .toEqual('xxxxxx');
+      expect(PythonPrereqCheckService.getObscuredCode(mixed))
+        .toEqual('  xx hello worldxxs xx  ');
+      expect(PythonPrereqCheckService.getObscuredCode(newLine))
+        .toEqual('xx');
+      expect(PythonPrereqCheckService.getObscuredCode(tabChar))
+        .toEqual('xx');
+    });
+
+    it('obscures escaped characters, preserving character length', function() {
```
seanlip: line length?

client/question/services/code_evaluators/PythonPrereqCheckServiceSpec.js
```diff
@@ -493,7 +493,70 @@ describe('PythonPrereqCheckService', function() {
...
+        '',
+        '    if (!arg + "abc"):',
+        '        return arg',
+        ''
+      ].join('\n');
+      var obscureCode =
+        PythonPrereqCheckService.getObscuredCode(code);
+
+      expect(code.length).toEqual(obscureCode.length);
+      expect(obscureCode).toContain('if xxxxxx');
```
seanlip: Would be a nicer test if you could just do "expect(obscureCode).toEqual(...long string...)"

client/question/services/code_evaluators/PythonPrereqCheckService.js
michaelrobbe: The loop ends when firstErrorColumnNumber >= 0, so without this line, The Column number ends up a negative string index.

client/question/services/code_evaluators/PythonPrereqCheckServiceSpec.js
michaelrobbe: Done.

client/question/services/code_evaluators/PythonPrereqCheckServiceSpec.js
michaelrobbe: Done.

client/question/services/code_evaluators/PythonPrereqCheckServiceSpec.js
michaelrobbe: Done.

client/question/services/code_evaluators/PythonPrereqCheckServiceSpec.js
michaelrobbe: That makes sense/feels right, I just keep getting caught by the linter, complaining about 'magic numbers'. So I pass them in as local vars.

client/question/services/code_evaluators/PythonPrereqCheckService.js
michaelrobbe: Makes sense. Done.

client/question/services/code_evaluators/PythonPrereqCheckService.js
michaelrobbe: Done.

client/question/services/code_evaluators/PythonPrereqCheckService.js
michaelrobbe: I guess I wasn't aware the commas were optional. Probably in part due to the tools and documentation suggesting them by default. Works without em? Done.

client/question/services/code_evaluators/PythonPrereqCheckService.js
michaelrobbe: I guess I wasn't aware the commas were optional. Probably in part due to the tools and documentation suggesting them by default. Works without em? Done.

client/question/services/code_evaluators/PythonPrereqCheckService.js
michaelrobbe: Done.

client/question/domain/PrereqCheckErrorObjectFactory.js
michaelrobbe: Noted. Done.

michaelrobbe:
@seanlip Let me know what else you see, I'd love to start up on the second half of the ticket.

PR #522, opened 2018-02-14
Change format of protractor tests to allow for less mandatory duplication.

eyurko:


protractor_tests/question.spec.js
```diff
@@ -17,13 +17,17 @@
...
 var QuestionsPage = require('./questions.pageObject.js');
-var utils = require('./utils.js');
+var utils = browser.params.utils;
 
 describe('submitting questions', function() {
   it('should successfully submit wrong code to a task', function() {
     var questionsPage = new QuestionsPage();
 
+    var setting = browser.waitForAngularEnabled();
+    browser.waitForAngularEnabled(!setting);
```
seanlip: This seems a strange construct to me. Can you explain what it does?

eyurko: http://www.protractortest.org/#/api?view=ProtractorBrowser.prototype.waitForAngularEnabled

protractor_tests/question.spec.js
```diff
@@ -16,19 +16,26 @@
...
-var utils = require('./utils.js');
+var QuestionsPage = browser.params.questionsPage;
+var utils = browser.params.utils;
 
 describe('submitting questions', function() {
+
   it('should successfully submit wrong code to a task', function() {
     var questionsPage = new QuestionsPage();
 
+    browser.waitForAngularEnabled(false);
```
seanlip: This is a lot of boilerplate to repeat for every test. I wonder whether it makes sense for the six lines to go into a utils.setUpQuestionPage() method?

eyurko: I'm going to put it in questionsPage.setUp(), instead.

protractor_tests/questions.pageObject.js
```diff
@@ -46,11 +46,27 @@ var QuestionsPage = function() {
    */
   var runCodeBtn = element(by.css('.protractor-test-run-code-btn'));
 
+  /**
+   * Performs boilerplate actions to get and set up the questions page.
+   *
+   * @param {object} browser
+   */
+  this.setUp = function(browser) {
```
seanlip: I believe `browser` is global, you shouldn't need to pass it in.

eyurko: so that's what I thought, but if I don't do that it tells me browser isn't defined. Would I have to do this.browser or?

seanlip: That's odd. I just ran it locally after removing the arg, and it passed. (This is client-side.) Are you referring to the server version?

Happy to work through it tomorrow too.

eyurko: huh, weird. I'll try removing it tomorrow and if it works I'll just push.

No, just client version.

seanlip:
LGTM

PR #528, opened 2018-02-22
Get rid of code related to question sets.

seanlip:


seanlip:
Also did some cleanup to unbreak the menu page and remove some try/catch logic that was hiding errors.

PR #533, opened 2018-03-06
Update Analytics to capture clicks

rabidbit:
Since links are technically external to the page, they are not captured. This allows us to know whether users actually clicked the download link, whether they visited our page, whether they opened the license file, or whether they clicked the feedback form link.


seanlip:
lgtm

PR #535, opened 2018-03-12
Update button labels and titles

rabidbit:
Changing "I think I'm done" to "Get Feedback" to address observations and comments in user studies, where users are reluctant to click "I think I'm done", thinking they should have a complete solution first.

Adding title attributes to most of the buttons and links for context to the user.

### DIFF
--- a/client/question/components/LearnerViewDirective.js
+++ b/client/question/components/LearnerViewDirective.js
@@ -56,12 +56,17 @@ tie.directive('learnerView', [function() {
                 </monospace-display-modal>
               </div>
               <div ng-hide="MonospaceDisplayModalService.isDisplayed()">
-                <button class="tie-code-reset tie-button" ng-click="resetFeedback()">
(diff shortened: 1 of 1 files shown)


rabidbit:
Hey Sean, Eric. What do you think of these labels?

They are an attempt to address user study results (users are still reluctant to click "I think I'm done" because they think they need completed, final code) and to make clear what each button does.

client/question/components/LearnerViewDirective.js
```diff
@@ -115,7 +121,7 @@ tie.directive('learnerView', [function() {
                 <li class="tie-leave-feedback-button"
                     title="Click to leave your feedback about TIE">
                   <a href="https://goo.gl/CcfCq4" target="_blank">
-                    Leave Feedback
+                    Leave feedback about TIE
```
seanlip: I think I'd use "on" rather than "about", but I'm happy to defer if you think "about" is better.

PR #538, opened 2018-03-15
Removed references to splicing.

### LINKED ISSUE #537: Syntax error refers to splice (which does not apply Python)
One of the syntax error messages refers to splice method (which does not apply Python). This confuses users and wastes their time as they try to research why this should be used and how.

eyurko:
Fixes #537

### DIFF
--- a/client/question/question.js
+++ b/client/question/question.js
@@ -623,9 +623,9 @@ tie.constant('RUNTIME_ERROR_FEEDBACK_MESSAGES', {
     generateMessage: function() {
       return [
         "Unfortunately Python doesn't support directly assigning ",
-        "characters in a string. If you need to do so, try splitting the ",
-        "string, removing the old characters, and adding new ones. ",
-        "If you need a refresher on string manipulation, check out the primer."
+        "characters in a string. If you need to do so, try slicing the ",
+        "string and adding new characters instead of assigning them. ",
+        "If you need a refresher on slicing, check out the primer."
       ].join('');
     }
   }, {
--- a/client/question/question.js
+++ b/client/question/question.js
@@ -621,10 +621,12 @@ tie.constant('RUNTIME_ERROR_FEEDBACK_MESSAGES', {
           " item assignment");
     },
     generateMessage: function() {
-      return ["Unfortunately Python doesn't support directly assigning ",
-        "characters in a string. If you need to do so, try splicing the ",
-        "string and reassigning the characters that way. If you need a ",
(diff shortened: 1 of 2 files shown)


client/question/question.js
```diff
@@ -621,10 +621,12 @@ tie.constant('RUNTIME_ERROR_FEEDBACK_MESSAGES', {
           " item assignment");
     },
     generateMessage: function() {
-      return ["Unfortunately Python doesn't support directly assigning ",
-        "characters in a string. If you need to do so, try splicing the ",
-        "string and reassigning the characters that way. If you need a ",
-        "refresher on splicing, check out the primer."].join('');
+      return [
+        "Unfortunately Python doesn't support directly assigning ",
+        "characters in a string. If you need to do so, try splitting the ",
```
seanlip: Hm. Are we talking here about something like this?

```
s = "cat"
s[0] = 'b'
assert s == "bat"
```

If so, wouldn't the most natural way to do it be something like `'b' + s[1:]`? I'm not sure I understand the "split()" approach.

eyurko:
Sure, or it might be

s = "cot"
s[1] = 'a'
assert s == "cat"

in which case you'd need to split the string.

--
Eric Yurko

seanlip:
Sorry, I think I'm still missing something. What does the code for splitting the string look like?

For example if I wanted to change "carat" to "caret", what would I do? I thought it would be `s[:3] + 'e' + s[4:]` but I can't understand how using s.split() would work.

eyurko:
Oh, that's what I meant by splitting the string. I forgot that "split" is
overloaded. Would "separate" work better, here?

seanlip:
I think the technical term for notation like `s[1:3]` is "slice". Perhaps something like "try building up the new string by slicing parts of the existing string and combining them with the new character"?

Should also verify that the Python primer we link to actually has an entry for "slice".

seanlip:
See e.g. https://developers.google.com/edu/python/strings: "The handy "slice" syntax (below) also works to extract any substring from a string."

eyurko:
Ah, excellent. I'll update.

--
Eric Yurko

seanlip:
(Travis seems unhappy though. Lint error?)

seanlip:
Also, Python primer doesn't seem to have a reference to "slicing"... that might need to be updated too.

eyurko:
yeah sorry I'm using the online editor which doesn't let me run my changes before I commit -- I don't have an IDE set up at home for a variety of W/L balance reasons :)

PR #548, opened 2018-03-24
Allow text in feedback to include links.

seanlip:
This now contains tests and is ready for review.


seanlip:
@eyurko could you PTAL at this? I updated it to reference the other PR that pulls out ThemeNameService.

Thanks!

client/question/question.js
```diff
@@ -624,7 +625,8 @@ tie.constant('RUNTIME_ERROR_FEEDBACK_MESSAGES', {
         "Unfortunately Python doesn't support directly assigning ",
         "characters in a string. If you need to do so, try slicing the ",
         "string and adding new characters instead of assigning them. ",
-        "If you need a refresher on slicing, check out the primer."
+        "If you need a refresher on slicing, check out the ",
```
eyurko: wait this should be whatever we fixed to not be about slicing.

client/question/services/FeedbackGeneratorServiceSpec.js
```diff
@@ -499,7 +499,8 @@ describe('FeedbackGeneratorService', function() {
             "Unfortunately Python doesn't support directly assigning ",
             "characters in a string. If you need to do so, try slicing the ",
             "string and adding new characters instead of assigning them. ",
-            "If you need a refresher on slicing, check out the primer."
+            "If you need a refresher on slicing, check out the ",
```
eyurko: same here

client/question/question.js
seanlip: I don't understand. I think this is the fixed version?

client/question/services/FeedbackGeneratorServiceSpec.js
seanlip: Ditto.

client/question/question.js
eyurko: ah, you're right. Sorry, I got slicing and splicing confused; my mistake.

PR #550, opened 2018-04-05
Extract theme-name-related stuff into a separate data service.

seanlip:
(And, as a bonus, adds Karma tests for that functionality.)

Once this is in I will update #548 to use this service to determine which Python primer page to route to.

### DIFF
--- a/client/question/components/LearnerViewDirective.js
+++ b/client/question/components/LearnerViewDirective.js
@@ -21,7 +21,7 @@ tie.directive('learnerView', [function() {
     restrict: 'E',
     scope: {},
     template: `
-      <div class="tie-wrapper" ng-class="{'night-mode': isInDarkMode}">
+      <div class="tie-wrapper" ng-class="{'night-mode': isInDarkMode()}">
         <div class="tie-question-ui-outer">
           <div class="tie-question-ui-inner">
             <div class="tie-question-ui">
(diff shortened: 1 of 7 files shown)


client/question/components/LearnerViewDirective.js
```diff
@@ -21,7 +21,7 @@ tie.directive('learnerView', [function() {
     restrict: 'E',
     scope: {},
     template: `
-      <div class="tie-wrapper" ng-class="{'night-mode': isInDarkMode}">
+      <div class="tie-wrapper" ng-class="{'night-mode': isInDarkMode()}">
```
eyurko: why not have this be IsDarkModeEnabled or DarkModeEnabled? IsInDarkMode seems like an odd construction, to me, since it's a setting.

client/question/services/CurrentThemeNameService.js
```diff
@@ -0,0 +1,47 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+/**
+ * @fileoverview A service that stores the user's light/dark mode preference.
+ */
+
+tie.constant('THEME_NAME_LIGHT', 'Light Theme');
+tie.constant('THEME_NAME_DARK', 'Dark Theme');
+
+tie.factory('CurrentThemeNameService', [
```
eyurko: Why not just ThemeService or ThemeNameService? CurrentThemeNameService is ... very specific. Seems like calling it ThemeService or TieThemeService would allow us to offload more theme options to this service in the future, should we expand it out.

client/question/components/LearnerViewDirective.js
seanlip: Done. This was a holdover from before, I agree that your suggestion is better. Thanks!

client/question/services/CurrentThemeNameService.js
seanlip: Done.

seanlip:
Hi @eyurko, all comments addressed, PTAL. Thanks!

PR #557, opened 2018-04-13
Bug #414 Pt. 2 - Adding errorLineNumber to feedbackParagraph, and highlighting the line

michaelrobbe:
Added Error Line Number to FeedbackParagraph, feedback generator adds: '(See line N of the code.)' when errorLineNumber is added feedback.appendTYPEParagraph().

### DIFF
--- a/client/question/components/LearnerViewDirective.js
+++ b/client/question/components/LearnerViewDirective.js
@@ -884,11 +884,11 @@ tie.directive('learnerView', [function() {
             }
           } else {
             var feedbackParagraphs = feedback.getParagraphs();
-            var errorLine = feedback.getErrorLineNumber();
+            var errorLineNumber = feedback.getErrorLineNumber();
             for (var i = 0; i < feedbackParagraphs.length; i++) {
               clearAllHighlights();
-              if (errorLine !== null) {
-                highlightLine(errorLine);
+              if (errorLineNumber !== null) {
+                highlightLine(errorLineNumber);
                 break;
               }
             }
--- a/client/question/components/LearnerViewDirective.js
+++ b/client/question/components/LearnerViewDirective.js
@@ -884,10 +884,11 @@ tie.directive('learnerView', [function() {
             }
           } else {
             var feedbackParagraphs = feedback.getParagraphs();
+            var errorLine = feedback.getErrorLineNumber();
             for (var i = 0; i < feedbackParagraphs.length; i++) {
               clearAllHighlights();
-              if (feedbackParagraphs[i].isErrorParagraph()) {
-                highlightLine(feedbackParagraphs[i].getErrorLineNumber());
+              if (errorLine !== null) {
+                highlightLine(errorLine);
                 break;
               }
             }
--- a/client/question/domain/FeedbackObjectFactory.js
+++ b/client/question/domain/FeedbackObjectFactory.js
@@ -61,6 +61,14 @@ tie.factory('FeedbackObjectFactory', [
        * @private
        */
       this._feedbackCategory = feedbackCategory;
+
+      /**
+       * Line number that correlates to error that triggered.
+       *
+       * @type {integer}
+       * @private
+       */
+      this._errorLineNumber = null;
     };
 
     // Instance methods.
@@ -175,6 +183,24 @@ tie.factory('FeedbackObjectFactory', [
      */
     Feedback.prototype.getFeedbackCategory = function() {
       return this._feedbackCategory;
+    };
+
+    /**
+     * A getter for the _errorLineNumber property.
+     *
+     * @returns {string}
+     */
+    Feedback.prototype.getErrorLineNumber = function() {
+      return this._errorLineNumber;
+    };
+
+    /**
+     * A setter for the _errorLineNumber property.
+     *
+     * @param {integer} errorLine to set the _errorLineNumber property to.
+     */
+    Feedback.prototype.setErrorLineNumber = function(newLineNumber) {
+      this._errorLineNumber = newLineNumber;
     };
 
     /**
--- a/client/question/domain/FeedbackObjectFactory.js
+++ b/client/question/domain/FeedbackObjectFactory.js
@@ -197,7 +197,7 @@ tie.factory('FeedbackObjectFactory', [
     /**
      * A setter for the _errorLineNumber property.
      *
-     * @param {integer} errorLine to set the _errorLineNumber property to.
+     * @param {integer} newLineNumber to set the _errorLineNumber property to.
      */
     Feedback.prototype.setErrorLineNumber = function(newLineNumber) {
       this._errorLineNumber = newLineNumber;
--- a/client/question/services/FeedbackGeneratorService.js
+++ b/client/question/services/FeedbackGeneratorService.js
@@ -803,6 +803,14 @@ tie.factory('FeedbackGeneratorService', [
                   feedback.appendErrorParagraph(paragraph.content);
                 }
               });
+
+              var errorLineNumber = prereqCheckFailure.getErrorLineNumber();
+              if (errorLineNumber) {
+                var errorStr = '(See line ' + errorLineNumber +
+                  ' of the code.)';
+                feedback.setErrorLineNumber(errorLineNumber);
+                feedback.appendTextParagraph(errorStr);
+              }
             }
           });
         } else if (prereqCheckFailure.hasInvalidAuxiliaryCodeCall()) {
--- a/client/question/services/FeedbackGeneratorService.js
+++ b/client/question/services/FeedbackGeneratorService.js
@@ -806,10 +806,9 @@ tie.factory('FeedbackGeneratorService', [
 
               var errorLineNumber = prereqCheckFailure.getErrorLineNumber();
               if (errorLineNumber) {
-                var errorStr = '(See line ' + errorLineNumber +
-                  ' of the code.)';
                 feedback.setErrorLineNumber(errorLineNumber);
-                feedback.appendTextParagraph(errorStr);
+                feedback.appendTextParagraph('(See line ' + errorLineNumber +
+                  ' of the code.)');
               }
             }
           });
--- a/client/question/services/FeedbackGeneratorServiceSpec.js
+++ b/client/question/services/FeedbackGeneratorServiceSpec.js
@@ -1401,7 +1401,7 @@ describe('FeedbackGeneratorService', function() {
       expect(feedback.getErrorLineNumber()).toBeNull();
     });
 
-    it('should return append the errorStr', function() {
+    it('should generate a errorLine message as the 2nd paragraph', function() {
       var incrementErrorCode = [
         'def myFunction(arg):',
         '    arg++',
@@ -1418,27 +1418,6 @@ describe('FeedbackGeneratorService', function() {
 
       expect(feedback.getErrorLineNumber()).toBe(2);
       expect(paragraphs[1].getContent()).toBe('(See line 2 of the code.)');
-    });
-
-    it('only return first errorLineNumber when multiple errors', function() {
-      var incrementErrorCode = [
-        'def myFunction(arg):',
-        '    arg = arg / 2',
-        '    arg--',
-        '    return arg',
-        'def myFunction2(arg):',
-        '    arg++',
-        '    return arg',
-        ''
-      ].join('\n');
-
-      var prereqFailure = PrereqCheckFailureObjectFactory.create(
-        PREREQ_CHECK_TYPE_WRONG_LANG, null, incrementErrorCode, 'decrementOp',
-        3, 4);
-      var feedback = FeedbackGeneratorService.getPrereqFailureFeedback(
-        prereqFailure);
-
-      expect(feedback.getErrorLineNumber()).toBe(3);
     });
   });
 });
--- a/client/question/services/FeedbackGeneratorServiceSpec.js
+++ b/client/question/services/FeedbackGeneratorServiceSpec.js
@@ -1384,4 +1384,61 @@ describe('FeedbackGeneratorService', function() {
       }
     );
   });
+
+  describe('prereqCheck errorLineNumber', function() {
+    it('should return null when there are no errors', function() {
+      var starterCode = [
+        'def myFunction(arg):',
+        '    return arg',
+        ''
+      ].join('\n');
+
+      var prereqFailure = PrereqCheckFailureObjectFactory.create(
+        PREREQ_CHECK_TYPE_WRONG_LANG, null, starterCode, 'notOp');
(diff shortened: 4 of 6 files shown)


client/question/components/LearnerViewDirective.js
```diff
@@ -886,7 +886,7 @@ tie.directive('learnerView', [function() {
             var feedbackParagraphs = feedback.getParagraphs();
             for (var i = 0; i < feedbackParagraphs.length; i++) {
               clearAllHighlights();
-              if (feedbackParagraphs[i].isErrorParagraph()) {
+              if (feedbackParagraphs[i].getErrorLineNumber()) {
```
seanlip: Is the error line number 0 or 1 indexed? If the former then what happens if it is 0?

In any case, might want to say "!== null" instead if that's really what you're checking.

client/question/domain/FeedbackObjectFactory.js
```diff
@@ -88,22 +88,24 @@ tie.factory('FeedbackObjectFactory', [
      *
      * @param {string} text String to be inserted in text paragraph.
      */
-    Feedback.prototype.appendTextParagraph = function(text) {
+    Feedback.prototype.appendTextParagraph = function(text, errorLineNumber) {
```
seanlip: I thought we decided not to do this? It seems messy to have every paragraph know about an "error line number". I also don't think that makes sense conceptually because the error line number relates to the entire feedback and not its individual paragraphs.

Can we pass the errorLineNumber to the learner view directive alongside the feedback as a whole (and not as part of the feedback)?

Ditto below.

client/question/domain/FeedbackParagraphObjectFactory.js
```diff
@@ -56,6 +56,14 @@ tie.factory('FeedbackParagraphObjectFactory', [
...
        */
       this._content = content;
+
+      /**
+       * Line number that correlates to error that triggered.
+       *
+       * @type {integer}
+       * @private
+       */
+      this._errorLineNumber = (errorLineNumber) ? errorLineNumber : null;
```
seanlip: `this._errorLineNumber = errorLineNumber | null;` is more JavaScript-y. If the first operand is falsy then the evaluated value is the second operand.

Could you also add a note in the docstring about whether this is 1- or 0-indexed?

client/question/services/FeedbackGeneratorService.js
```diff
@@ -803,6 +803,13 @@ tie.factory('FeedbackGeneratorService', [
                   feedback.appendErrorParagraph(paragraph.content);
                 }
               });
+
+              var errorLineNumber = prereqCheckFailure.getErrorLineNumber();
+              if (errorLineNumber) {
+                var errorStr = "(See line " + errorLineNumber +
```
seanlip: Use single quotes throughout.

michaelrobbe:
@seanlip Reversed the FeedbackParagraph object changes, and added a variable + Getter/Setter to the Feedback object.

client/question/components/LearnerViewDirective.js
```diff
@@ -884,10 +884,11 @@ tie.directive('learnerView', [function() {
             }
           } else {
             var feedbackParagraphs = feedback.getParagraphs();
+            var errorLine = feedback.getErrorLineNumber();
```
seanlip: For consistency call this errorLineNumber

client/question/domain/FeedbackObjectFactory.js
```diff
@@ -177,6 +185,24 @@ tie.factory('FeedbackObjectFactory', [
...
+     * @returns {string}
+     */
+    Feedback.prototype.getErrorLineNumber = function() {
+      return this._errorLineNumber;
+    };
+
+    /**
+     * A setter for the _errorLineNumber property.
+     *
+     * @param {integer} errorLine to set the _errorLineNumber property to.
```
seanlip: Param name in docstring does not match arg name

client/question/services/FeedbackGeneratorService.js
```diff
@@ -803,6 +803,14 @@ tie.factory('FeedbackGeneratorService', [
                   feedback.appendErrorParagraph(paragraph.content);
                 }
               });
+
+              var errorLineNumber = prereqCheckFailure.getErrorLineNumber();
```
seanlip: Interesting that this is in the forEach loop. If a piece of code exhibits multiple such errors, do we show all of them? If so, there would be different error line numbers for each. Would it make more sense to show just the first one?

client/question/services/FeedbackGeneratorService.js
```diff
@@ -803,6 +803,14 @@ tie.factory('FeedbackGeneratorService', [
...
+              if (errorLineNumber) {
+                var errorStr = '(See line ' + errorLineNumber +
```
seanlip: Optional: Consider just inlining this in line 812 rather than declaring a new var.

client/question/components/LearnerViewDirective.js
michaelrobbe: Done.

client/question/domain/FeedbackObjectFactory.js
michaelrobbe: Fixed.

client/question/services/FeedbackGeneratorService.js
michaelrobbe: This lives inside of the hasWrongLanguage Prereq check error loop. Line errors are only captured this way for hasWrongLanguage Errors. The loop breaks after finding the first error.

Syntax errors and line numbers are currently in the spoiler link popout. It wasn't clear if they should be added as feedback bubbles at this time.

michaelrobbe:
@seanlip Karma tests have been added, I went back and forth on which test doc they should live inside. Settled on FeedbackGeneratorServiceSpec, since PythonPrereqCheckServiceSpec already tests that errorLineNumber and errorCharacterNumber make it into the PrereqFailureObject. Still learning how to write decent tests.

client/question/services/FeedbackGeneratorService.js
seanlip: I think what I mean is, should this also break once we've successfully set an error line number and appended a text paragraph? Looking at this code in isolation suggests that the error line number that is set corresponds to the latest one found, and there may be one or more text paragraphs containing error strings.

client/question/services/FeedbackGeneratorServiceSpec.js
```diff
@@ -1384,4 +1384,61 @@ describe('FeedbackGeneratorService', function() {
...
+
+      var prereqFailure = PrereqCheckFailureObjectFactory.create(
+        PREREQ_CHECK_TYPE_WRONG_LANG, null, starterCode, 'notOp');
+      var feedback = FeedbackGeneratorService.getPrereqFailureFeedback(
+        prereqFailure);
+
+      expect(feedback.getErrorLineNumber()).toBeNull();
+    });
+
+    it('should return append the errorStr', function() {
```
seanlip: Could you please write this as a full sentence? I'm not sure what it's saying.

client/question/services/FeedbackGeneratorServiceSpec.js
```diff
@@ -1384,4 +1384,61 @@ describe('FeedbackGeneratorService', function() {
...
+        2, 4);
+      var feedback = FeedbackGeneratorService.getPrereqFailureFeedback(
+        prereqFailure);
+      var paragraphs = feedback.getParagraphs();
+
+      expect(feedback.getErrorLineNumber()).toBe(2);
+      expect(paragraphs[1].getContent()).toBe('(See line 2 of the code.)');
+    });
+
+    it('only return first errorLineNumber when multiple errors', function() {
```
seanlip: This test should be broader, since the code is not run -- so what this test is really testing is that if you supply line number 3 then the feedback will say line 3, which is unsurprising.

Instead, I think you can have this particular test as one of the end-to-end tests in SolutionHandlerServiceSpec.js. Just submit some code that fails more than one prerequisite check and check what feedback comes out.

seanlip:
Thanks @michaelrobbe. The tests look good in general! Just had a couple of comments.

michaelrobbe:
@seanlip Let me know if there's anything left to take care of.

client/question/services/SolutionHandlerServiceSpec.js
```diff
@@ -355,6 +355,29 @@ describe('SolutionHandlerService', function() {
           done();
         });
       });
+
+      it('should return 1st errorLineNumber, when multiple errors', function() {
```
seanlip: Add done as an arg here and call it on line 379 when all async processing finishes. See the [Jasmine docs](https://jasmine.github.io/2.0/introduction#section-Asynchronous_Support) and the other existing tests in this file.

client/question/services/code_evaluators/PythonPrereqCheckService.js
```diff
@@ -89,32 +89,45 @@ tie.factory('PythonPrereqCheckService', [
     var detectAndGetWrongLanguageType = function(code) {
       var obscuredCode = getObscuredCode(code);
       var rawCodeArray = code.split('\n');
+      var lowestErrorCharNumber;
+      var firstError;
```
seanlip: Initialize to null.

client/question/services/code_evaluators/PythonPrereqCheckService.js
```diff
@@ -89,32 +89,45 @@ tie.factory('PythonPrereqCheckService', [
     var detectAndGetWrongLanguageType = function(code) {
       var obscuredCode = getObscuredCode(code);
       var rawCodeArray = code.split('\n');
+      var lowestErrorCharNumber;
```
seanlip: Initialize appropriately to a very large number.

client/question/services/code_evaluators/PythonPrereqCheckService.js
```diff
@@ -89,32 +89,45 @@ tie.factory('PythonPrereqCheckService', [
...
         // Lookup character location of error
-        var firstErrorCharNumber = obscuredCode.search(regexp);
-        if (firstErrorCharNumber !== -1) {
+        var errorCharNumber = obscuredCode.search(regexp);
+
+        // Regex search returns -1 for no match. Use reverse King-of-the-hill
+        // check to find lowest number.
+        if (errorCharNumber > -1) {
+          if (firstError && errorCharNumber > lowestErrorCharNumber) {
+            break;
```
seanlip: Why break and not continue?

client/question/services/code_evaluators/PythonPrereqCheckService.js
```diff
@@ -89,32 +89,45 @@ tie.factory('PythonPrereqCheckService', [
...
 
           // Add line length to the >0 firstErrorColumnNumber
-          var firstErrorColumnNumber = firstErrorCharNumber + rawCodeArray[
+          var firstErrorColumnNumber = errorCharNumber + rawCodeArray[
             firstErrorLineNumber - 1].length;
 
-          return PrereqCheckErrorObjectFactory.create(
-            error.errorName, firstErrorLineNumber, firstErrorColumnNumber);
+          firstError = PrereqCheckErrorObjectFactory.create(
+          error.errorName, firstErrorLineNumber, firstErrorColumnNumber);
```
seanlip: Indent by 2.

client/question/services/code_evaluators/PythonPrereqCheckService.js
```diff
@@ -89,32 +89,45 @@ tie.factory('PythonPrereqCheckService', [
...
 
         // Lookup character location of error
-        var firstErrorCharNumber = obscuredCode.search(regexp);
-        if (firstErrorCharNumber !== -1) {
+        var errorCharNumber = obscuredCode.search(regexp);
+
+        // Regex search returns -1 for no match. Use reverse King-of-the-hill
+        // check to find lowest number.
+        if (errorCharNumber > -1) {
+          if (firstError && errorCharNumber > lowestErrorCharNumber) {
```
seanlip: should be `>=`, perhaps

client/question/services/FeedbackGeneratorServiceSpec.js
michaelrobbe: Done.

client/question/services/FeedbackGeneratorServiceSpec.js
michaelrobbe: Done.

client/question/services/SolutionHandlerServiceSpec.js
michaelrobbe: done().

client/question/services/code_evaluators/PythonPrereqCheckService.js
michaelrobbe: Done.

client/question/services/code_evaluators/PythonPrereqCheckService.js
michaelrobbe: In my draft I did initialized it, but then reworked it so I didn't have some arbitrary number set. I readded my high-number point. Done.

client/question/services/code_evaluators/PythonPrereqCheckService.js
michaelrobbe: Done.

client/question/services/code_evaluators/PythonPrereqCheckService.js
michaelrobbe: Done

client/question/services/code_evaluators/PythonPrereqCheckService.js
michaelrobbe: The only way that >= would return valid for 105 would be if two wrongLang errors occur in the exact same place. If that's the case, the line number would be correct.

client/question/services/FeedbackGeneratorService.js
michaelrobbe: Updated, will go out with the next commit.

client/question/services/code_evaluators/PythonPrereqCheckService.js
seanlip: Ah, I was thinking more in terms of saving time; presumably you don't need to process stuff if the char number is the same. But this is optional.

client/question/services/code_evaluators/PythonPrereqCheckService.js
```diff
@@ -122,7 +122,7 @@ tie.factory('PythonPrereqCheckService', [
             firstErrorLineNumber - 1].length;
 
           firstError = PrereqCheckErrorObjectFactory.create(
-          error.errorName, firstErrorLineNumber, firstErrorColumnNumber);
+           error.errorName, firstErrorLineNumber, firstErrorColumnNumber);
```
seanlip: Indent by 2 not 1.

PR #560, opened 2018-04-19
Add fix for timeout python errors.

eyurko:


client/question/domain/ErrorTracebackObjectFactory.js
```diff
@@ -143,6 +143,10 @@ tie.factory('ErrorTracebackObjectFactory', [
       // Split the error string by newlines for easier processing.
       var splitErrorMessage = pythonError.trim().split(/\r?\n/);
       var errorMessage = splitErrorMessage[splitErrorMessage.length - 1];
+      if (splitErrorMessage.length === 1) {
```
seanlip: This seems like a rather indirect way of figuring out whether there's a line number. In particular, is it possible that multi-line messages might have no line number? If so then that means this check is insufficient.

One suggestion is to check for the presence of LINE_DELIMITER_PYTHON in the error message -- or, more robustly, use a regex to check for things like ` line \d+ ` to prevent false positives.

eyurko: I can do that.

eyurko: That should be a bit better.

client/question/domain/ErrorTracebackObjectFactorySpec.js
```diff
@@ -128,6 +128,18 @@ NameError: global name 'potato' is not defined`;
       ).toEqual(errorTraceback.getErrorString());
     });
 
+    it('should return correct traceback for Python timeout errors', function() {
+      var errorTraceback = ErrorTracebackObjectFactory.create(
+      'TimeLimitError: Your code ran real slow, my dude.', null);
```
seanlip: Indent by 2.

PR #563, opened 2018-04-23
Allow server to catch user syntax errors before runtime.

eyurko:
whoooops~

Like I said, we should really get a flow chart.

### DIFF
--- a/client/question/services/code_evaluators/PythonCodeRunnerServiceSpec.js
+++ b/client/question/services/code_evaluators/PythonCodeRunnerServiceSpec.js
@@ -67,25 +67,26 @@ describe('PythonCodeRunnerService', function() {
     it('processes a compile-time syntax error', function() {
       var code = [
         'def myFunction(arg):',
-        '    result = arg.rstrip()',
+        '    result = arg.rstrip()x',
         '    return result',
         ''
       ].join('\n');
       responseDict.stderr = [
         'Traceback (most recent call last):',
-        '  File "main.py", line 54, in <module>',
-        '    StudentCode().fmcc, all_tasks_test_inputs[0][0]))',
-        '  File "main.py", line 19, in runTest',
-        '    output = func(input)',
-        '  File "main.py", line 28, in fmcc',
-        '    return 2 / 0',
-        'ZeroDivisionError: integer division or modulo by zero'
+        '  File "main.py", line 2',
+        '    result = arg.rstrip()x',
+        'SyntaxError: invalid syntax'
       ].join('\n');
       $httpBackend.expectPOST('/ajax/compile_code').respond(
         HTTP_STATUS_CODE_OK, responseDict);
       spyOn(ServerHandlerService, 'doesServerExist').and.returnValue(true);
       spyOn(PythonCodeRunnerService,
         '_processCodeCompilationServerResponse').and.returnValue(null);
+      PythonCodeRunnerService.compileCodeAsync(code).then(
+        function(result) {
+          expect(result.getErrorString()).toEqual(
+            'SyntaxError: invalid syntax on line 2');
+        });
       PythonCodeRunnerService.compileCodeAsync(code);
       $httpBackend.flush();
     });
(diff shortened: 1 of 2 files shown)


client/question/services/code_evaluators/PythonCodeRunnerServiceSpec.js
```diff
@@ -202,5 +202,33 @@ describe('PythonCodeRunnerService', function() {
       expect(codeEvalResult.getBuggyOutputTestResults()).toEqual(null);
       expect(codeEvalResult.getErrorInput()).toEqual(null);
     });
+
+    it('returns a CodeEvalResult with syntax error post-compile', function() {
```
seanlip: Might want to add another test at _compileCodeAsync level for the error case.

client/question/services/code_evaluators/PythonCodeRunnerServiceSpec.js
```diff
@@ -64,6 +64,32 @@ describe('PythonCodeRunnerService', function() {
       $httpBackend.flush();
     });
 
+    it('processes a compile-time syntax error', function() {
+      var code = [
+        'def myFunction(arg):',
+        '    result = arg.rstrip()',
+        '    return result',
```
seanlip: I don't see a syntax error here

client/question/services/code_evaluators/PythonCodeRunnerServiceSpec.js
```diff
@@ -64,6 +64,32 @@ describe('PythonCodeRunnerService', function() {
...
+        ''
+      ].join('\n');
+      responseDict.stderr = [
```
seanlip: Not clear how this stderr relates to the code above?

client/question/services/code_evaluators/PythonCodeRunnerServiceSpec.js
eyurko: It doesn't have or need one -- it mocks it out. I can add one, but it's not really in the scope of the function. Would you prefer a comment?

client/question/services/code_evaluators/PythonCodeRunnerServiceSpec.js
eyurko: They're unrelated -- this is just to mock out the functionality. The post returns responseDict on line 84 -- the code is just ... boilerplate, essentially.

client/question/services/code_evaluators/PythonCodeRunnerServiceSpec.js
seanlip: I think you have two choices: (a) either make the code and error consistent so that it is a reasonable real-world input, or (b) make the code clearly say "mock code" and the error string clearly say "mock stderr" or something. (b) is probably better in order to avoid the reader thinking that the content actually matters.

Having pseudo-content that doesn't represent a real case just seems ... kind of confusing.

client/question/services/code_evaluators/PythonCodeRunnerServiceSpec.js
```diff
@@ -64,6 +64,32 @@ describe('PythonCodeRunnerService', function() {
...
+        '  File "main.py", line 19, in runTest',
+        '    output = func(input)',
+        '  File "main.py", line 28, in fmcc',
+        '    return 2 / 0',
+        'ZeroDivisionError: integer division or modulo by zero'
+      ].join('\n');
+      $httpBackend.expectPOST('/ajax/compile_code').respond(
+        HTTP_STATUS_CODE_OK, responseDict);
+      spyOn(ServerHandlerService, 'doesServerExist').and.returnValue(true);
+      spyOn(PythonCodeRunnerService,
```
seanlip: Can we not mock this? I understand the idea of keeping a unit test "unit", but this is an internal method called by the compile code function and my original concern when asking for an additional test was to check that the interoperation between the multiple functions works correctly (at least within the scope of compileCodeAsync).

Or we could leave this here and add another test that checks that. Up to you.

seanlip:
Oh whoops, sorry. I just realized I wasn't supposed to merge this, right?

Let me know, I can revert.

PR #575, opened 2018-05-17
Fixed #572: Updated starter code to clarify language and method param

### LINKED ISSUE #572: (a) Make it more clear what language the coding window is in and (b) Not clear that 's' should be treated as a string.

michaelrobbe:
Added two comment lines to the starter code.

No impact on Karma tests.
Sanity tested, everything seems to work as expected.


assets/questions/reverseWords.js
```diff
@@ -20,7 +20,11 @@ globalData.questions['reverseWords'] = {  // eslint-disable-line dot-notation
   title: 'Reverse Words',
   starterCode: {
     python:
-`def reverseWords(s):
+`# Python version 2.7
+
+def reverseWords(s):
+    # This reverseWords method processes a string (s)
```
seanlip: Probably should add period at end.

assets/questions/reverseWords.js
```diff
@@ -20,7 +20,11 @@ globalData.questions['reverseWords'] = {  // eslint-disable-line dot-notation
   title: 'Reverse Words',
   starterCode: {
     python:
-`def reverseWords(s):
+`# Python version 2.7
```
seanlip: @rabidbit -- I missed your comment about this in the issue thread, but one concern I have with this is that it would need to be manually added for every question. This seems like it would be tedious.

Is there a way to put it in the UI instead? For example, could it exist as text above/below the coding window (that might be later replaced by a dropdown when we support multiple languages)?

rabidbit: Are you talking about stating what language is being used? If so, my last comment in #572 was to have the language name next to "New to Python?", not be included as a comment (see my screenshot and subsequent discussion on font size). What's included as a comment are question-specific info, like what type the input is. IIRC, further offline team discussion was to use Angular to hide the text language specifying the language and display the drop down based on the number of languages available.

michaelrobbe:
@seanlip Merge Conflict resolved.

client/question/components/LearnerViewDirective.js
```diff
@@ -89,14 +89,19 @@ tie.directive('learnerView', [function() {
...
                 </div>
-                <select ng-if="SERVER_URL"
-                    class="tie-select-menu"
-                    name="lang-select-menu">
-                  <option value="Python" selected>Python</option>
-                </select>
                 <button class="tie-code-reset tie-button protractor-test-reset-code-btn" name="code-reset" ng-click="resetCode()" title="Click to clear your code">
                   Reset Code
                 </button>
+                <p class="tie-language-label">Language: {{supportedLanguageCount == 1 ? languageLabel : ""}}</p>
```
seanlip: Could you do: `<p ng-if="supportedLanguageCount === 1">` instead (and remove the ternary expression)? I think that would be cleaner.

client/question/components/LearnerViewDirectiveSpec.js
```diff
@@ -421,4 +423,18 @@ describe('LearnerViewDirective', function() {
...
+
+  describe('supportedLanguageLabels', function() {
+    it('should provide a descriptive label for each language', function() {
+      expect($scope.supportedLanguageLabels[LANGUAGE]).toEqual('Python 2.7');
+
+      expect($scope.supportedLanguageCount).toEqual(
+        ALL_SUPPORTED_LANGUAGES.length);
+
+      for (var i = ALL_SUPPORTED_LANGUAGES.length - 1; i >= 0; i--) {
+        var $label = $scope.supportedLanguageLabels[ALL_SUPPORTED_LANGUAGES[i]];
```
seanlip: Why the "$" prefix? Should just be `label`.

client/question/components/LearnerViewDirectiveSpec.js
```diff
@@ -421,4 +423,18 @@ describe('LearnerViewDirective', function() {
...
+        expect($label).toBeDefined();
```
seanlip: I guess this just checks "not undefined"? Would be stronger to check: "is a string", and "is not the empty string".

assets/questions/reverseWords.js
```diff
@@ -21,6 +21,8 @@ globalData.questions['reverseWords'] = {  // eslint-disable-line dot-notation
   starterCode: {
     python:
 `def reverseWords(s):
+    # This reverseWords method processes a string (s)
```
seanlip: I think I mentioned this before, but we should add a period at the end.

seanlip:
Thanks @michaelrobbe! Merging.

client/question/components/LearnerViewDirective.js
michaelrobbe: Updated to remove the ternary expression. Intentionally left the "Language:", as a label for the dropdown as needed.

client/question/components/LearnerViewDirectiveSpec.js
michaelrobbe: Removed $ label. PHP programming tic.

client/question/components/LearnerViewDirectiveSpec.js
michaelrobbe: Jasmine doesn't have a type checking method, so I check 'typeof label' against 'string', and "toBeTruthy()" method (which validates false for an empty string).

assets/questions/reverseWords.js
michaelrobbe: Period added.

PR #576, opened 2018-05-22
Add new css for dark mode.

eyurko:
Don't let your themes be themes


eyurko:
No reason why not. I'll delete it in this PR and then merge, tomorrow. I'll get Antoine to approve the sync.

PR #578, opened 2018-05-23
Move pages into client directory for consistency.

eyurko:
This is defining a new pattern for TIE.


seanlip:
LGTM.

One of your commit messages says "It's weird that the e2e tests don't run automatically for the client, either." As an FYI, it is intentional that they aren't part of the submit-to-github process, since e2e tests typically take a while to run and we have the Travis guard to do that asynchronously as a safety check. Happy to discuss more offline if you like.

PR #579, opened 2018-05-23
Fix #558: Loading dots appear right after submission.

### LINKED ISSUE #558: Loading dots should show as soon as submit button is clicked
In certain cases (e.g. when compiling/running the code takes time), there are delays in showing the loading dots when the submit button is clicked.

We should always show the loading dots as soon as the submit button is clicked, and they should display for at least DURATION_MSEC_WAIT_FOR_FEEDBACK (cf ConversationLogDataService) before getting replaced by the actual feedback. @eyurko noted that the problem seems to be that addFeedbackBalloon is called with the feedback itself, which ... requires the feedback to already be there. Since that's called **after** the code's been executed, there's no way to get the dots to show up before code execution without a decent shift to that code.

This issue is aimed at doing that shift and ensuring that the loading dots behave correctly in all cases.

kchen05:
* Add new function to check if right code balloon is present.
* Base loading dots off of right code balloon's presence.

### DIFF
--- a/client/question/components/SpeechBalloonsContainerDirective.js
+++ b/client/question/components/SpeechBalloonsContainerDirective.js
@@ -22,7 +22,7 @@ tie.directive('speechBalloonsContainer', [function() {
     scope: {},
     template: `
       <div>
-        <div class="tie-dot-container" ng-class="{'tie-display-dots': ConversationLogDataService.isNewBalloonPending()}"}>
+        <div class="tie-dot-container" ng-class="{'tie-display-dots': (ConversationLogDataService.data.speechBalloonList.length > 0 ? ConversationLogDataService.data.speechBalloonList[0].isDisplayedOnRight() : false)}"}>
           <div class="tie-dot tie-dot-1"></div>
           <div class="tie-dot tie-dot-2"></div>
           <div class="tie-dot tie-dot-3"></div>
@@ -77,7 +77,7 @@ tie.directive('speechBalloonsContainer', [function() {
         }
         .tie-dot {
           -webkit-animation-name: tie-dot;
-          -webkit-animation-duration: 1.5s;
+          -webkit-animation-duration: 1.2s;
           -webkit-animation-iteration-count: infinite;
           background-color: black;
           border-radius: 2px;
@@ -92,6 +92,7 @@ tie.directive('speechBalloonsContainer', [function() {
           height: 10px;
           opacity: 0;
           padding-left: 5px;
+          transition-delay: 0.2s;
         }
         .night-mode .tie-dot {
           background-color: #E0E0E0;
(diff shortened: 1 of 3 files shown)


client/question/domain/SpeechBalloonObjectFactory.js
```diff
@@ -67,6 +67,15 @@ tie.factory('SpeechBalloonObjectFactory', [
       return this._type === SPEECH_BALLOON_TYPE_FEEDBACK;
     };
 
+    /**
+     * Returns whether the code balloon is displayed on the right.
+     *
+     * @returns {boolean}
+     */
+    SpeechBalloon.prototype.isDisplayedOnRight = function() {
```
eyurko: I'd recommend calling this something like isCodeSubmission rather than isDisplayedOnRight; the latter seems more arbitrary whereas the former seems like "we want this to happen on code submission"

client/question/domain/SpeechBalloonObjectFactory.js
```diff
@@ -67,6 +67,15 @@ tie.factory('SpeechBalloonObjectFactory', [
       return this._type === SPEECH_BALLOON_TYPE_FEEDBACK;
     };
 
+    /**
+     * Returns whether the code balloon is displayed on the right.
```
eyurko: Update comment to reflect name change (see below)

client/question/components/SpeechBalloonsContainerDirective.js
```diff
@@ -22,7 +22,7 @@ tie.directive('speechBalloonsContainer', [function() {
     scope: {},
     template: `
       <div>
-        <div class="tie-dot-container" ng-class="{'tie-display-dots': ConversationLogDataService.isNewBalloonPending()}"}>
+        <div class="tie-dot-container" ng-class="{'tie-display-dots': (ConversationLogDataService.data.speechBalloonList.length > 0 ? ConversationLogDataService.data.speechBalloonList[0].isDisplayedOnRight() : false)}"}>
```
eyurko: Might be better to avoid the ternary and just use ConversationLogDataService.data.speechBalloonList.length > 0 && ConversationLogDataService.data.speechBalloonList[0].isDisplayedOnRight()

client/question/domain/SpeechBalloonObjectFactory.js
kchen05: Fixed

eyurko: Nit: Returns whether or not the current speech balloon is a code submission balloon.

Not gonna block submit, though, so change that and you're good to go.

seanlip: This does not appear to have been changed (see @eyurko's last comment).

kchen05: Hm, seems that the latest commit did not get merged in. I can make a new pull request for this.

PR #581, opened 2018-05-24
Fix 558: Fix function description.

kchen05:
* Fix function description so that testing and function descriptions match.
* Loading dots appear upon code submission.


client/question/domain/SpeechBalloonObjectFactory.js
```diff
@@ -67,6 +67,15 @@ tie.factory('SpeechBalloonObjectFactory', [
       return this._type === SPEECH_BALLOON_TYPE_FEEDBACK;
     };
 
+    /**
+     * Returns whether the current balloon is code submission.
```
eyurko: Returns whether or not the current balloon is a code submission

PR #583, opened 2018-05-25
Refactor ConversationDataLogService to SessionHistoryService, which is tasked with maintaining the TIE learner transcript.

seanlip:
Note: there will be a follow-up PR to this one that refactors LocalStorageService and connects it with SessionHistoryService.

This is largely a rename + slight reconceptualization of ConversationLogDataService with a bit of tidying-up (e.g. adding some serialization logic), in preparation for migrating local storage responsibility to it.


client/question/services/SessionHistoryService.js
```diff
@@ -11,45 +11,57 @@
...
-            SpeechBalloonObjectFactory.createCodeBalloon(submittedCode));
-      },
-      /**
-       * Clears the feedback log.
+       * Resets the session transcript.
        */
-      clear: function() {
-        data.speechBalloonList = [];
+      reset: function() {
+        // Setting the length of the existing array to 0 allows us to preserve
```
eyurko: huh; didn't know you could do that. That's fun.

client/question/services/SessionHistoryServiceSpec.js
```diff
@@ -0,0 +1,126 @@
...
+  beforeEach(inject(function($injector, _$timeout_) {
+    SessionHistoryService = $injector.get('SessionHistoryService');
+    FeedbackParagraphObjectFactory = $injector.get(
+      'FeedbackParagraphObjectFactory');
+    DURATION_MSEC_WAIT_FOR_FEEDBACK = $injector.get(
+      'DURATION_MSEC_WAIT_FOR_FEEDBACK');
+    $timeout = _$timeout_;
+  }));
+
+  describe('core behaviour', function() {
```
eyurko: there's a huge opportunity for me to get american sassy and correct this to "behavior" but I'll allouw it

PR #584, opened 2018-05-25
Update filepath in step 3 of README.

seanlip:


rabidbit:
How will we distinguish between different questions for both the stand-alone and hosted versions? One option mentioned is to use params for the hosted version. I don't have strong feelings about using params versus using different filenames, as long as it's concise and simple.

seanlip:
We've always been using a qid parameter for this, regardless of the filepath/url leading up to it. E.g. /client/question.html?qid=reverseWords

rabidbit:
Will this also be the case for the stand-alone version? (I realize that is out of scope for this particular PR)

seanlip:
Yes.

PR #588, opened 2018-06-13
Disable "Get Feedback" button when code submission is processing.

kchen05:
* Students can no longer spam "Get Feedback" button.

### DIFF
--- a/client/question/components/LearnerViewDirective.js
+++ b/client/question/components/LearnerViewDirective.js
@@ -227,7 +227,7 @@ tie.directive('learnerView', [function() {
         .tie-button-green[disabled] {
           opacity: 0.4;
         }
-        .tie-button-green[disabled]:hover{
+        .tie-button-green[disabled]:hover {
           border: none;
         }
         .night-mode .tie-button-green {
--- a/client/question/components/LearnerViewDirective.js
+++ b/client/question/components/LearnerViewDirective.js
@@ -102,7 +102,7 @@ tie.directive('learnerView', [function() {
                     ng-show="autosaveTextIsDisplayed">
                   Saving code...
                 </div>
-                <button class="tie-run-button tie-button tie-button-green protractor-test-run-code-btn" ng-click="submitCode(editorContents.code)" ng-disabled="SessionHistoryService.isNewBalloonPending()" title="Click anytime you want feedback on your code">
+                <button class="tie-run-button tie-button tie-button-green protractor-test-run-code-btn" ng-click="submitCode(editorContents.code)" ng-disabled="processing" title="Click anytime you want feedback on your code">
                   Get Feedback
                 </button>
               </div>
@@ -224,6 +224,12 @@ tie.directive('learnerView', [function() {
         .tie-button-green:active {
           background-color: #265221;
         }
+        .tie-button-green[disabled] {
+          opacity: 0.4;
+        }
+        .tie-button-green[disabled]:hover{
+          border: none;
+        }
         .night-mode .tie-button-green {
           background-color: #3C5C14;
           color: #ffffff;
@@ -233,6 +239,12 @@ tie.directive('learnerView', [function() {
         }
         .night-mode .tie-button-green:active {
           background-color: #265221;
+        }
+        .night-mode .tie-button-green[disabled] {
+          opacity: 0.4;
+        }
+        .night-mode .tie-button-green[disabled]:hover {
+          border: none;
         }
         .tie-code-auto-save {
           font-family: Roboto, 'Helvetica Neue', 'Lucida Grande', sans-serif;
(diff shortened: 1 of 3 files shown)


client/question/components/LearnerViewDirective.js
```diff
@@ -590,6 +602,13 @@ tie.directive('learnerView', [function() {
          */
         $scope.codeChangeLoopPromise = null;
 
+        /**
+        * Keeps track of whether a current code submission is being run.
+        *
+        * @type {boolean}
+        */
+        $scope.processing = false;
```
seanlip: This name is too vague. Perhaps something like $scope.submissionIsBeingProcessed?

kchen05: Hm, I'll try to take a deeper look tomorrow morning as well (seems to already work as intended when I also repulled from master) . The original reasoning behind this change was because I'm able to spam the "Get Feedback" button on my desktop local and server versions (as well as the official TIE link). I originally thought that it had to do with the isBalloonPending function not updating fast enough.

kchen05:
Hm, I'll try to take a deeper look tomorrow morning as well (seems to already work as intended when I also repulled from master) . The original reasoning behind this change was because I'm able to spam the "Get Feedback" button on my desktop local and server versions (as well as the official TIE link). I originally thought that it had to do with the isBalloonPending function not updating fast enough.

kchen05:
When I pull from master, it seems that the desired effect already exists (only one code submission at a time). However, it seems that the server version introduces some delay, so a user is able to rapidly click the "Get Feedback" button. @seanlip besides making the variable name more specific, is there something else that I should do to better organize the code?

seanlip:
Can you talk more about _why_ the server version introduces some delay? An analysis of why this occurs would be helpful in order to figure out what remedy is needed.

seanlip:
Great analysis, thanks @kchen05!

I'd say that, since every code submission is supposed to always lead to some feedback, it's fine to increment data.numBalloonsPending as soon as we display the code balloon. Would that work?

kchen05:
Of course!

That should work. Would it make sense then that as soon as we add a codeBalloon, this specifies that we are waiting for a feedbackBalloon even though the function might not yet have explicitly been called?

seanlip:
I think that sounds fine to me! Just add some documentation in the code to explain what is going on, specifically the assumption that a code balloon implies that a subsequent feedback balloon is coming.

kchen05:
Makes sense! Thanks Sean. Will fix the PR to represent the new implementation. Also, any thoughts on the CSS change once the submission is being processed? It currently just makes the button less opaque so the student knows the feedback is still "thinking".

seanlip:
The CSS change sounds entirely reasonable to me; I would just suggest getting feedback from @rabidbit on it. I remember playing with it last night and I thought it seemed quite natural.

rabidbit:
I think the CSS looks fine and is what I would expect. Thanks!

client/question/components/LearnerViewDirective.js
```diff
@@ -224,6 +224,12 @@ tie.directive('learnerView', [function() {
         .tie-button-green:active {
           background-color: #265221;
         }
+        .tie-button-green[disabled] {
+          opacity: 0.4;
+        }
+        .tie-button-green[disabled]:hover{
```
seanlip: Nit: add one space before {

client/question/services/SessionHistoryService.js
```diff
@@ -75,6 +75,10 @@ tie.factory('SessionHistoryService', [
        * Adds a new code balloon to the beginning of the list.
        */
       addCodeBalloon: function(submittedCode) {
+        // The number of balloons pending is increased by assumption that
```
seanlip: We increment the number of balloons here, because adding a code balloon implies that a feedback balloon will soon follow.

(Also, conceptually, I recommend doing this only _after_ you've added the code balloon. It makes no practical difference code-wise but is easier for devs to understand, since otherwise both balloons are "pending" at the time of the increment.)

client/question/services/SessionHistoryServiceSpec.js
```diff
@@ -100,12 +109,12 @@ describe('SessionHistoryService', function() {
...
       expect(SessionHistoryService.isNewBalloonPending()).toBe(true);
       expect(
-        SessionHistoryService.getBindableSessionTranscript().length).toBe(0);
+        SessionHistoryService.getBindableSessionTranscript().length).toBe(1);
 
       $timeout.flush(DURATION_MSEC_WAIT_FOR_FEEDBACK);
       expect(
         SessionHistoryService.getBindableSessionTranscript().length
-      ).toBe(1);
+      ).toBe(2);
```
seanlip: Maybe here, also expect isNewBalloonPending() to be false.

kchen05:
Noted. Thanks for all the feedback and dialogue!

PR #589, opened 2018-06-13
Logic for displaying student stdOut.

kchen05:
* Should be merged in after PR #588 . 
* Handles both client and server side logic for print although not currently displayed anywhere.


client/question/domain/CodeEvalResultObjectFactory.js
```diff
@@ -174,6 +189,39 @@ tie.factory('CodeEvalResultObjectFactory', [
...
+     * @param {Array<Task>} tasks The list of tasks for the current question.
+     * @returns {number} The index of the first failed test case, or index of
+     * last test case if all test cases passed.
+     */
+    CodeEvalResult.prototype.getIndexOfFirstFailedTest = function(tasks) {
+      if (this._observedOutputs.length === 0) {
+        return 0;
+      }
+
+      var userOutput = this._observedOutputs;
```
seanlip: Why rename, especially since you use the original variable in line 214?

Also consider `observedOutputs` instead, it's a bit confusing to call the same thing "output" and "outputs".

client/question/domain/CodeEvalResultObjectFactorySpec.js
```diff
@@ -165,6 +165,132 @@ describe('CodeEvalResultObjectFactory', function() {
...
+        ERROR_STRING, ERROR_INPUT);
+      expect(codeEvalResult6.getIndexOfFirstFailedTest(tasks))
+        .toEqual(0);
+
+      var codeEvalResult7 = CodeEvalResultObjectFactory.create(
+        CODE, OUTPUT, [[[true, true]]],
+        BUGGY_OUTPUT_TEST_RESULTS, PERFORMANCE_TEST_RESULTS,
+        ERROR_STRING, ERROR_INPUT);
+      expect(codeEvalResult7.getIndexOfFirstFailedTest(tasks))
+        .toEqual(1);
```
seanlip: This seems wrong? Shouldn't it be 2?

Also, one question -- is it possible for this._observedOutputs to have a different length compared to the length of `tasks` (or for its subcomponents to have lengths not matching the corresponding subparts of `tasks`)? If so then perhaps this should be more explicitly handled in the main function.

client/question/question.js
```diff
@@ -251,6 +251,7 @@ tie.constant('SYSTEM_CODE', {
     '        global most_recent_input',
     '        most_recent_input = copy.deepcopy(input)',
     '        output = func(input)',
+    '        print separator',
```
seanlip: Hm -- I think it would be better for the _caller_ of this function to print the separator instead. Given its name, this function should probably just return the output without side-effects.

client/question/question.js
```diff
@@ -723,6 +724,14 @@ tie.constant('SECONDS_TO_MILLISECONDS', 1000);
  */
 tie.constant('CODE_CHANGE_DEBOUNCE_SECONDS', 5);
 
+/** Default user stdOut separator length.
```
seanlip: Perhaps give a bit more context about what the "stdout separator" is and how/why it's used.

client/question/services/StdOutSeparatorService.js
```diff
@@ -0,0 +1,112 @@
...
+  */
+
+tie.factory('StdOutSeparatorService', [
+  function() {
+    /**
+     * The separator used in between test cases' stdOut.
+     *
+     * @type {string}
+     **/
+    var separator = "";
```
seanlip: Start with this being null, rather than an empty string.

Change setSeparator to init() (if it's only ever set once) or resetSeparator() (if it's set multiple times during a question session). That makes it clearer that it's supposed to be more of a "do once and then don't do it again" action.

client/question/services/StdOutSeparatorService.js
```diff
@@ -0,0 +1,112 @@
...
+      */
+      getSeparator: function() {
+        return separator;
+      },
+
+      /**
+      * Returns the corresponding stdOut for the given test case number for
+      * client version.
+      *
+      * @param {Array} stdOut Array of entire stdOut for all test cases.
```
seanlip: From reading this description, I don't understand what this variable represents. I think of stdout as a string; what does it mean for it to be an array? And if it is an array with one element per test case, why isn't the method just `return stdOut[testNum]`?

This might need to be decribed more clearly.

seanlip: Also, this function looks like it's getting into the weeds a bit ... and it's not clear that this service should know about the client/server difference. Ideally we would abstract that away so that services that process the stdout don't have to care about where the stdout comes from.

Would it be helpful to change codeEvalResult.output to an array which contains the stdout per test (or null, if we never get to that test case)? That way, it would already be preprocessed and callers can use it easily after that.

client/question/services/code_evaluators/PythonCodeRunnerService.js
```diff
@@ -89,9 +91,19 @@ tie.factory('PythonCodeRunnerService', [
         }
 
         // The run was successful.
-        return CodeEvalResultObjectFactory.create(
+        var codeEvalResult = CodeEvalResultObjectFactory.create(
           code, outputLines.join('\n'), observedOutputs,
           buggyOutputTestResults, performanceTestResults, null, null);
+        var question = CurrentQuestionService.getCurrentQuestion();
+        if (!question) {
```
seanlip: Why would there be no question?

client/question/services/code_evaluators/PythonCodeRunnerService.js
```diff
@@ -89,9 +91,19 @@ tie.factory('PythonCodeRunnerService', [
...
+          return codeEvalResult;
+        }
+        var tasks = question.getTasks();
+        var testToDisplay = codeEvalResult.getIndexOfFirstFailedTest(tasks);
```
seanlip: A few issues here:

- This doesn't seem to be relevant to the purpose of this service, which is to run the code and return the result of doing so.
- It's not a good idea to redefine outputLines in the middle of a function. It makes it hard to tell what the variable represents since it refers to different things at different places.
- I'm not sure why outputLines is being modified in the first place. Where is it being used after this?

From a conceptual design perspective, probably better to just store the output-per-test-case in the codeEvalResult and refer to it (or the relevant part of it) later, when needed.

client/question/services/code_evaluators/PythonCodeRunnerService.js
```diff
@@ -166,12 +178,23 @@ tie.factory('PythonCodeRunnerService', [
...
       } else if (responseData.results) {
-        return CodeEvalResultObjectFactory.create(
+        var codeEvalResult = CodeEvalResultObjectFactory.create(
             code, responseData.stdout,
             responseData.results[VARNAME_OBSERVED_OUTPUTS],
             responseData.results[VARNAME_BUGGY_OUTPUT_TEST_RESULTS],
             responseData.results[VARNAME_PERFORMANCE_TEST_RESULTS],
             null, null);
+        var question = CurrentQuestionService.getCurrentQuestion();
+        if (!question) {
```
seanlip: In what case would there be no question?

client/question/services/code_evaluators/PythonCodeRunnerService.js
```diff
@@ -166,12 +178,23 @@ tie.factory('PythonCodeRunnerService', [
...
+          return codeEvalResult;
+        }
+        var tasks = question.getTasks();
+        var testToDisplay = codeEvalResult.getIndexOfFirstFailedTask(tasks);
+        // eslint-disable-next-line no-unused-vars
+        var stdOutToDisplay = StdOutSeparatorService.getTestCaseOutput(
```
seanlip: See notes above; also, where are these variables being used?

client/question/services/code_preprocessors/PythonCodePreprocessorService.js
```diff
@@ -313,6 +313,28 @@ tie.factory('PythonCodePreprocessorService', [
       return VALID_WHITESPACE_REGEX.test(c);
     };
 
+    /**
+    * Creates and sets the string separator for user print stdOut.
+    *
+    * @returns {string}
+    * @private
+    */
+    var _generateStringSeparator = function() {
```
seanlip: _generateStdoutSeparator perhaps -- "string separator" is quite generic.

Also, there is a mismatch between function name and implementation. The function sounds like it's supposed to generate a "string separator", but instead returns something called separatorCode...

client/question/services/code_preprocessors/PythonCodePreprocessorService.js
```diff
@@ -313,6 +313,28 @@ tie.factory('PythonCodePreprocessorService', [
...
+      var chars = 'AaBbCcDdEeFfGgHhIiJjKkLlMmNnOoPpQqRrSsTtUuVvWwXxYyZz';
+      var separator = '';
+      for (var i = 0; i < SEPARATOR_LENGTH; i++) {
+        var index = Math.floor(Math.random() * chars.length);
+        separator += chars.charAt(index);
+      }
+      StdOutSeparatorService.setSeparator(separator);
```
seanlip: Per notes above, it might not be worth having a StdOutSeparatorService.

Instead, you could pass the separator string along with the preprocessed code to the code runner service. The latter can then use it to split the output returned from the code run into per-test-case output.

kchen05:
Thanks for the detailed pass @seanlip ! I'll be sure to go through and address them today.

client/question/domain/CodeEvalResultObjectFactorySpec.js
kchen05: I'll modify this to null later to signify that all test cases were passed, but originally it was 1 since the 2nd test case (1 via 0-indexing) is the last test case to be executed. 

If a user only finished 1/3 tasks for example, `this._observedOutputs` will be shorter than the length of `tasks` since `tasks` will have all of those for that question, which was tested here since it gets handled in the `getIndexOfFirstFailedTest` function (loops over `this._observedOutputs` rather than `tasks`). Would it be better to test this elsewhere?

seanlip: But they passed the 2nd test case (the one that was 0-indexed as 1), right? So it can't be the "first failed test case". If anything, perhaps it should be the first case for which we got no results?

I think the right thing to do depends on the stdout behaviour we want when a task is completely passed (and in other cases with partial results). Did you get a decision on that? It might be worth enumerating them.

kchen05: After my conversation with John, it seems that the best option is to display whichever test case was last run (regardless of if the code completely passed or not). In this case, would it be best to modify the function name to more accurately depict that it returns the last test case executed?

client/question/domain/CodeEvalResultObjectFactory.js
```diff
@@ -106,8 +107,22 @@ tie.factory('CodeEvalResultObjectFactory', [
      */
     CodeEvalResult.prototype.hasSamePreprocessedCodeAs = function(
         otherCodeEvalResult) {
-      return (
-        this._preprocessedCode === otherCodeEvalResult.getPreprocessedCode());
+      var separatorStart = this._preprocessedCode.indexOf('separator =');
```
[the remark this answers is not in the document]
kchen05: Done.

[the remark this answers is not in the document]
kchen05: Done.

client/question/domain/CodeEvalResultObjectFactory.js
```diff
@@ -106,8 +107,22 @@ tie.factory('CodeEvalResultObjectFactory', [
...
+      var otherCode = otherCodeEvalResult.getPreprocessedCode();
+      var otherSeparatorStart = otherCode.indexOf('separator =');
+
+      // No separator exists in the preprocess code.
```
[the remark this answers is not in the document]
kchen05: Fixed.

client/question/domain/CodeEvalResultObjectFactory.js
```diff
@@ -106,8 +107,22 @@ tie.factory('CodeEvalResultObjectFactory', [
...
+      if (separatorStart === -1 && otherSeparatorStart === -1) {
+        return this._preprocessedCode === otherCode;
+      }
+      // Separator values change per run.
+      var offset = SEPARATOR_LENGTH + 'separator = ""'.length;
```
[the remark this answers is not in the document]
kchen05: Done.

client/question/domain/CodeEvalResultObjectFactory.js
```diff
@@ -174,6 +189,39 @@ tie.factory('CodeEvalResultObjectFactory', [
       return null;
     };
 
+  /**
```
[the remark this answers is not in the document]
kchen05: Fixed.

client/question/domain/CodeEvalResultObjectFactory.js
```diff
@@ -174,6 +189,39 @@ tie.factory('CodeEvalResultObjectFactory', [
...
+     * Compares the results to the expected values of the test cases, and
+     * returns the test number of the first failed test case (0-indexed). The
+     * test number is independent of tasks and testSuites but rather the overall
+     * test case number.
+     *
+     * @param {Array<Task>} tasks The list of tasks for the current question.
+     * @returns {number} The index of the first failed test case, or index of
```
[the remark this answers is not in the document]
kchen05: Changed so null if all test cases passed and returns object (contains task number, test suite number, test case number, and overall test case number) of which test failed if something failed.

client/question/question.js
kchen05: Done. Moved to PythonCodePreprocessor

client/question/question.js
kchen05: Done.

client/question/services/StdOutSeparatorService.js
```diff
@@ -0,0 +1,112 @@
...
+ // distributed under the License is distributed on an "AS-IS" BASIS,
+ // WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ // See the License for the specific language governing permissions and
+ // limitations under the License.
+
+ /**
+  * @fileoverview Service for displaying printed output.
+  */
+
+tie.factory('StdOutSeparatorService', [
```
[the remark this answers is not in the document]
kchen05: StdOutSeparatorService removed.

client/question/services/StdOutSeparatorService.js
kchen05: StdOutSeparatorService removed.

client/question/services/StdOutSeparatorService.js
kchen05: Moved this logic to be handled in PythonCodeRunner and associated stdOut with CodeEvalResultObjects rather than having a StdOutSeparatorService.

client/question/services/code_evaluators/PythonCodeRunnerService.js
kchen05: Done. Thanks for the good suggestion as well!

client/question/services/code_evaluators/PythonCodeRunnerService.js
kchen05: Fixed specs to always have questions.

client/question/services/code_evaluators/PythonCodeRunnerService.js
kchen05: Fixed.

kchen05:
@eyurko @seanlip  I have fixed the logic structure as per suggestions -- your review would be greatly appreciated. Thanks for all the help along the way, and let me know what further changes could be implemented to improve the code!

seanlip:
@kchen05 please make sure to reply to all open comment threads. I think some of the open comment threads don't have a response yet. Thanks!

client/question/domain/CodeEvalResultObjectFactory.js
```diff
@@ -174,6 +189,39 @@ tie.factory('CodeEvalResultObjectFactory', [
       return null;
     };
 
+  /**
+     * Compares the results to the expected values of the test cases, and
+     * returns the test number of the first failed test case (0-indexed). The
+     * test number is independent of tasks and testSuites but rather the overall
```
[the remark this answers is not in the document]
kchen05: Fixed description.

client/question/domain/CodeEvalResultObjectFactory.js
kchen05: Reverted to original name.

client/question/services/code_evaluators/PythonCodeRunnerService.js
kchen05: Fixed specs to accurately always have a question.

client/question/services/code_preprocessors/PythonCodePreprocessorService.js
kchen05: Fixed.

client/question/services/code_preprocessors/PythonCodePreprocessorService.js
kchen05: Thanks for the suggestion! Removed StdOutSeparatorService and instead created PreprocessedCode to handle both attributes.

kchen05:
Sorry must have missed some comment threads. Should have replied to all now!

client/question/domain/CodeEvalResultObjectFactory.js
```diff
@@ -174,6 +202,79 @@ tie.factory('CodeEvalResultObjectFactory', [
...
+                testSuiteNum: j,
+                testNum: k,
+                overallTestNum: overallTestNum
+              };
+            }
+            overallTestNum += 1;
+          }
+        }
+      }
+     // Returns null if all test cases passed.
```
seanlip: Indent by 1.

client/question/domain/CodeEvalResultObjectFactory.js
```diff
@@ -174,6 +202,79 @@ tie.factory('CodeEvalResultObjectFactory', [
...
+     * numbers and overall test number (0-indexed) of the first failed test
+     * case. The overall test number is defined by concatenating all the
+     * test cases together, in order.
+     *
+     * @param {Array<Task>} tasks The list of tasks for the current question.
+     * @returns {*|null} An object containing task, testSuite, test
+     * index numbers, and overall test number of the first failing test
+     * case. Returns null if no tests for the given tasks failed.
+     */
+    CodeEvalResult.prototype.getIndexOfFirstFailedTest = function(tasks) {
```
seanlip: This seems misnamed. The return value isn't an index.

That said, maybe it should be? The only actual call I can find to this just requires the overallTestNum. So maybe we should just return that?

client/question/domain/PreprocessedCodeObjectFactory.js
```diff
@@ -0,0 +1,88 @@
...
+     * PreprocessedCode stores the entire preprocessed code as well as the
+     * string separator generated during print output. This string separator
+     * is utilized in splitting the total output into the corresponding
+     * outputs for each test case.
+     */
+
+    /**
+     * Constructor for PreprocessedCode
+     *
+     * @param {string} preprocessedCode Preprocessed student code
```
seanlip: Maybe we should call this preprocessedCodeString, to distinguish it from the PreprocessedCode object that contains it.

client/question/services/code_evaluators/PythonCodeRunnerService.js
```diff
@@ -52,7 +58,19 @@ tie.factory('PythonCodeRunnerService', [
       outputLines.push(line);
     };
 
-    var _runCodeInClient = function(code) {
+    var _runCodeInClient = function(inputCode) {
+      var code;
+      var separator;
+
+      // The input code is a PreprocessedCode object and ready to be run.
+      if (typeof (inputCode) === 'object') {
```
seanlip: This seems indirect. I think this method should still do what it used to do, which is to just take a piece of code, run it, and return the output.

This method could then be called by a couple of more "public-facing" methods along the lines of compileCode() and runPreprocessedCode() (or similar). Those methods would each do the processing required to get the code that actually needs to be run, send that code to this helper method, then get the output, parse it, and return it.

client/question/services/code_evaluators/PythonCodeRunnerService.js
```diff
@@ -89,9 +107,13 @@ tie.factory('PythonCodeRunnerService', [
         }
 
         // The run was successful.
-        return CodeEvalResultObjectFactory.create(
-          code, outputLines.join('\n'), observedOutputs,
+        var standardizedOutput = _standardizeOutputInClient(
```
seanlip: Output standardization should be independent of whether the client/server version is being used.

client/question/services/code_evaluators/PythonCodeRunnerService.js
```diff
@@ -89,9 +107,13 @@ tie.factory('PythonCodeRunnerService', [
...
+          outputLines, separator);
+        var codeEvalResult = CodeEvalResultObjectFactory.create(
+          code, standardizedOutput, observedOutputs,
           buggyOutputTestResults, performanceTestResults, null, null);
+        _displayPrintedOutput(codeEvalResult);
```
seanlip: This call seems to be in the wrong place. This is a code runner, it shouldn't be displaying anything. It should just run the code and return the evaluation result.

client/question/services/code_evaluators/PythonCodeRunnerService.js
```diff
@@ -106,6 +128,37 @@ tie.factory('PythonCodeRunnerService', [
       });
     };
 
+    // Converts the output returned from running the code.
+    // Before: array in which each entry was one line of stdOut
```
seanlip: Use standard jsdoc. Document types and descriptions of args, return value, etc.

client/question/services/code_evaluators/PythonCodeRunnerService.js
```diff
@@ -106,6 +128,37 @@ tie.factory('PythonCodeRunnerService', [
...
+      var standardizedOutput = [];
+      var currentStringStartIndex = 0;
+      var currentStringEndIndex = 0;
+      for (currentStringEndIndex; currentStringEndIndex < stdOutArray.length;
+        currentStringEndIndex++) {
+        if (stdOutArray[currentStringEndIndex] === separator) {
+          // Convert the sub-array into a condensed string of the entire
+          // stdOut for that given test case.
+          var currentStdOutStringRepresentation = '';
+          for (var i = currentStringStartIndex;
```
seanlip: This seems very manual. What about the following approach:

- Concatenate stdOut into a string (which it might already be originally?)
- Split the string by the separator value to get an array for each test case

client/question/services/code_evaluators/PythonCodeRunnerService.js
```diff
@@ -158,25 +214,58 @@ tie.factory('PythonCodeRunnerService', [
...
-            code, responseData.stdout,
+        var standardizedOutput = _standardizeServerOutput(
+          responseData.stdout, separator);
+        var codeEvalResult = CodeEvalResultObjectFactory.create(
+            code, standardizedOutput,
             responseData.results[VARNAME_OBSERVED_OUTPUTS],
             responseData.results[VARNAME_BUGGY_OUTPUT_TEST_RESULTS],
             responseData.results[VARNAME_PERFORMANCE_TEST_RESULTS],
             null, null);
+        _displayPrintedOutput(codeEvalResult);
```
seanlip: This should not be here.

client/question/services/code_evaluators/PythonCodeRunnerService.js
```diff
@@ -158,25 +214,58 @@ tie.factory('PythonCodeRunnerService', [
...
       } else {
         throw Error('A server error occurred. Please try again.');
       }
     };
 
+    // Converts the output string generated by running the code.
+    // Before: string of all stdOut concatenated together.
+    // After: array in which the i-th entry is the string corresponding
+    // to the stdOut of the i-th test case.
+    var _standardizeServerOutput = function(stdOutString, separator) {
```
seanlip: As above -- don't have separate functions for client and server. Instead, update the stdout after runCode so that it is in a consistent format in both cases, and so that there's no need for future execution to know whether the client or the server was used for execution.

client/question/services/code_evaluators/PythonCodeRunnerService.js
```diff
@@ -158,25 +214,58 @@ tie.factory('PythonCodeRunnerService', [
...
+      // StdOut will sometimes have extra characters at the end. Cleaning up
+      // the stdout here while also ensuring that the stdOut ends with
+      // a newline character.
+      var lastSeparatorIndex = stdOutString.lastIndexOf(separator);
+      var cleanedStdOutString = stdOutString.slice(
+          0, lastSeparatorIndex + SEPARATOR_LENGTH) + '\n';
+      return cleanedStdOutString.split(separator + '\n');
+    };
+
+    var _displayPrintedOutput = function(codeEvalResult) {
```
seanlip: This should not be here.

client/question/services/code_preprocessors/PythonCodePreprocessorService.js
```diff
@@ -82,6 +84,12 @@ tie.factory('PythonCodePreprocessorService', [
      */
     var UPPER_BOUND_RATIO_IF_LINEAR = (LARGE_INPUT_SIZE / SMALL_INPUT_SIZE) * 3;
 
+    /**
+     * Used to separate the print outputs of different test cases, which are
+     * all combined into one output string after the code is run.
+     */
+    var separator = null;
```
seanlip: Better not to make this a global within this service.

Instead, in line 653, do something like `var separator = generateNewSeparator();` and then do generateOutputSeparatorCode(separator) when you need the code, and finally return the value of the local variable. There's no need for this to be global/persistent, it's a per-run thing.

client/question/domain/CodeEvalResultObjectFactory.js
```diff
@@ -174,6 +202,79 @@ tie.factory('CodeEvalResultObjectFactory', [
...
+      var testToDisplay = this.getIndexOfFirstFailedTest(tasks);
+
+      // If user passed all tests, display output of the last test case of
+      // the last task that they completed.
+      if (testToDisplay === null) {
+        var overallNumTests = 0;
+        for (var i = 0; i < this._observedOutputs.length; i++) {
+          for (var j = 0; j < this._observedOutputs[i].length; j++) {
+            for (var k = 0; k < this._observedOutputs[i][j].length; k++) {
+              overallNumTests += 1;
```
eyurko: is there any reason that we can't do

overallNumTests += this._observedOutputs[i][j].length

instead of the third inner for loop?

client/question/domain/CodeEvalResultObjectFactorySpec.js
```diff
@@ -165,6 +166,298 @@ describe('CodeEvalResultObjectFactory', function() {
...
+        .toEqual({
+          taskNum: 0,
+          testSuiteNum: 0,
+          testNum: 0,
+          overallTestNum: 0});
+    });
+  });
+
+  describe('getOutputToDisplay', function() {
+    it('should correctly get the output to display', function() {
```
eyurko: should correctly get the user's output to display

client/question/services/code_evaluators/PythonCodeRunnerServiceSpec.js
```diff
@@ -19,13 +19,85 @@ describe('PythonCodeRunnerService', function() {
...
   var ServerHandlerService;
+  var PreprocessedCodeObjectFactory;
   var responseDict = {};
   var VARNAME_OBSERVED_OUTPUTS = 'correctness_test_results';
   var VARNAME_BUGGY_OUTPUT_TEST_RESULTS = 'buggy_output_test_results';
   var VARNAME_PERFORMANCE_TEST_RESULTS = 'performance_test_results';
   var VARNAME_MOST_RECENT_INPUT = 'most_recent_input';
   var HTTP_STATUS_CODE_OK = 200;
   var HTTP_STATUS_CODE_SERVER_ERROR = 500;
+  var QuestionObjectFactory;
```
eyurko: Group all your factories, services, and constants together.

client/question/services/code_evaluators/PythonCodeRunnerServiceSpec.js
```diff
@@ -254,4 +343,62 @@ describe('PythonCodeRunnerService', function() {
       expect(codeEvalResult.getErrorInput()).toEqual(null);
     });
   });
+
+  describe('_standardizeServerOutput', function() {
+    it('should work for task completion stdOut', function() {
```
eyurko: maybe be a bit more specific here -- "should correctly display stdOut on task completion" even, is better.

client/question/services/code_evaluators/PythonCodeRunnerServiceSpec.js
```diff
@@ -254,4 +343,62 @@ describe('PythonCodeRunnerService', function() {
...
+      var sep = '12345678901234567890';
+      var stdOut = '1\n2\nHi\n' + sep + '\n1\n2\nHey\n' + sep +
+        '\n1\n2\nHello\n' + sep;
+      expect(PythonCodeRunnerService._standardizeServerOutput(
+        stdOut, sep)).toEqual(
+        ['1\n2\nHi\n', '1\n2\nHey\n', '1\n2\nHello\n', '']);
+    });
+
+    it('should work for question completion stdOut', function() {
```
eyurko: ditto here and below

client/question/services/code_evaluators/PythonCodeRunnerService.js
```diff
@@ -177,6 +211,22 @@ tie.factory('PythonCodeRunnerService', [
...
     };
 
+    // Converts the consolidated output string into an array where the
+    // i-th entry in the array corresponds to the i-th overall test
+    // case.
+    var _createOutputArray = function(stdOutString, separator) {
+      if (stdOutString.length === 0) {
+        return [];
+      }
+      // StdOut will sometimes have extra characters at the end. Cleaning up
```
seanlip: Can you explain why? (Where do these chars come from?)

client/question/question.js
```diff
@@ -723,6 +723,25 @@ tie.constant('SECONDS_TO_MILLISECONDS', 1000);
...
+// eslint-disable-next-line no-magic-numbers
+tie.constant('SEPARATOR_LENGTH', 20);
+
+/**
+* Default line prefix to denote separator code.
+*
+* @type {string}
+* @constant
+*/
+tie.constant('SEPARATOR_LINE_PREFIX', 'separator = "');
```
seanlip: Where is this used?

client/question/services/code_preprocessors/PythonCodePreprocessorService.js
```diff
@@ -375,12 +411,17 @@ tie.factory('PythonCodePreprocessorService', [
 
         testCode += '\n';
         testCode += [
+          '',
+          'def testOutputWithStdOutSeparator(test_input):',
```
seanlip: Hm, the name seems a bit off. printSeparatorAndGetTestOutput() perhaps?

client/question/services/code_evaluators/PythonCodeRunnerService.js
kchen05: Hm, I'm actually not too sure. It's occurring in the run somewhere since the output returned from the run already has the extra characters. I'll look into it more but it seems to be occurring on both the server version and client version with the same "garbage" at the end.

seanlip: OK. We should have an understanding of why this happens; thanks for looking into it.

seanlip:
Well, almost. Looks like you have some test failures...

I restarted the tests; you might also need to merge master into your branch.

PR #590, opened 2018-06-14
Fix cake path

rabidbit:
Moving question.html up a level in path invalidated the path to cake.


seanlip:
Thanks! Adding @eyurko to review to check compatibility with server version.

seanlip:
Btw, @eyurko -- we should also make sure our e2e test paths cover this sort of thing. I can see it being a common issue since the cake isn't displayed on the page at the outset.

PR #591, opened 2018-06-15
Fix #556: Fix when "View full code" link is displayed

### LINKED ISSUE #556: The "View full code" link should be omitted in contexts that don't require a pop-up modal.
See screenshot:

screenshot from 2018-04-12 18-20-25

There are two spurious "View full code" links in the conversation transcript.

rabidbit:
Note: This fixes a bug in the user submitted code speech balloons. It also fixed the issue observed in #556 ; however, if feedback speech balloon prints code snippets longer than 3 lines, it will be truncated as well. But I think that's fine because this places a limit on how much code text is placed in feedback speech balloons, and the point of the separate code windows is to offload lengthy code into a separate UI element.

### DIFF
--- a/client/question/components/CodeSnippetDirective.js
+++ b/client/question/components/CodeSnippetDirective.js
@@ -27,7 +27,7 @@ tie.directive('codeSnippet', [function() {
         <span class="tie-code-snippet-line">{{line}}</span>
         <br>
       </span>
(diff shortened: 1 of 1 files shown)


rabidbit:
Is it possible to do this in another PR? I'm not 100% sure I follow, and I would not want to introduce a regression on such a simple PR.

rabidbit:
If you wouldn't mind doing that PR, that would be great.

PR #592, opened 2018-06-16
Refactor code snippet directive to have clearer variable names.

seanlip:
This is a follow-up to #591, specifically [this comment](https://github.com/google/tie/pull/591#pullrequestreview-129026731).

### DIFF
--- a/client/question/components/CodeSnippetDirective.js
+++ b/client/question/components/CodeSnippetDirective.js
@@ -23,11 +23,11 @@ tie.directive('codeSnippet', [function() {
       getContent: '&content'
     },
     template: `
-      <span ng-repeat="line in abbreviatedSnippetLines track by $index">
+      <span ng-repeat="line in snippetLines track by $index">
         <span class="tie-code-snippet-line">{{line}}</span>
         <br>
       </span>
-      <span ng-if="abbreviatedSnippetLines.length > MAX_NUM_LINES_IN_ABBREVIATED_SNIPPET">
+      <span ng-if="codeLines.length > MAX_NUM_LINES_IN_SNIPPET">
         <a href ng-click="openCodeModal()" ng-if="!isModalOpen()">View full code</a>
         <span ng-if="isModalOpen()">View full code</span>
       </span>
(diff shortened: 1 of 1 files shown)


client/question/components/CodeSnippetDirective.js
```diff
@@ -23,11 +23,11 @@ tie.directive('codeSnippet', [function() {
       getContent: '&content'
     },
     template: `
-      <span ng-repeat="line in abbreviatedSnippetLines track by $index">
+      <span ng-repeat="line in snippetLines track by $index">
         <span class="tie-code-snippet-line">{{line}}</span>
         <br>
       </span>
-      <span ng-if="abbreviatedSnippetLines.length > MAX_NUM_LINES_IN_ABBREVIATED_SNIPPET">
+      <span ng-if="codeLines.length > MAX_NUM_LINES_IN_SNIPPET">
```
rabidbit: Just to confirm, abbreviatedSnippetLines is replaced by codeLines here but it's replaced by snippetLines elsewhere?

seanlip: Yes, that was intended; I think the old code was less clear. Does this version make sense to you (logically)?

client/question/components/CodeSnippetDirective.js
```diff
@@ -50,31 +50,30 @@ tie.directive('codeSnippet', [function() {
          */
         $scope.isModalOpen = MonospaceDisplayModalService.isDisplayed;
 
-        // The maximum number of lines to show in the abbreviated code snippet.
-        $scope.MAX_NUM_LINES_IN_ABBREVIATED_SNIPPET = 3;
+        // The maximum number of lines to show in a code snippet.
```
rabidbit: Is context lost when we drop "abbreviated"? This is in reference to user code that is > 3 lines in the feedback window and is therefore truncated, right? Would this cause confusion with code snippets that are not truncated (e.g., short code or code displayed in a detailed code pop-up/modal)?

client/question/components/CodeSnippetDirective.js
rabidbit: I see. In that case, it seems the older code was actually different? I think the newer version makes sense.

client/question/components/CodeSnippetDirective.js
rabidbit: I see. I agree that the two terms sound redundant. As long as the convention of calling truncated code "snippet" is consistent throughout, SGTM.

client/question/components/CodeSnippetDirective.js
seanlip: Yes, that's right. Thanks!

seanlip:
Thanks for the review!

PR #593, opened 2018-06-28
Added pattern for verifying correct behavior in the event of a compile-time error.

eyurko:
This also fixes any bugs that might occur with processing an error happening during the compile step via a server.

### DIFF
--- a/client/question/services/FeedbackGeneratorService.js
+++ b/client/question/services/FeedbackGeneratorService.js
@@ -464,7 +464,9 @@ tie.factory('FeedbackGeneratorService', [
       feedback.appendTextParagraph([
         'A server error has occurred. We are looking into it ',
         'and will fix it as quickly as possible. We apologize ',
-        'for the inconvenience.'
+        'for the inconvenience. Please refresh and try again.\n',
+        'If the problem still has not been resolved, please email ',
+        'tie-team@google.com to escalate.'
       ].join(''));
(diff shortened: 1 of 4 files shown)


client/question/services/FeedbackGeneratorService.js
```diff
@@ -464,7 +464,9 @@ tie.factory('FeedbackGeneratorService', [
       feedback.appendTextParagraph([
         'A server error has occurred. We are looking into it ',
         'and will fix it as quickly as possible. We apologize ',
-        'for the inconvenience.'
+        'for the inconvenience. Please refresh and try again.\n',
```
seanlip: Did you test this manually? I'm not sure \n plays well with HTML.

client/question/services/FeedbackGeneratorService.js
```diff
@@ -594,6 +596,46 @@ tie.factory('FeedbackGeneratorService', [
...
 
+    /**
+     * Returns the Feedback object associated with a compilation submission.
+     *
+     * @param {string} errorString A string detailing any errors.
+     * @returns {Feedback}
+     * @private
+     */
+    var _getCompileFeedback = function(errorString) {
+      // If we get a timeout error, here, it's not due to the student's code.
```
seanlip: I think the timeout error should not be sent to the frontend at all if it's not actually a timeout error.

PR #594, opened 2018-06-29
Implement ConversationManagerService and migrate some feedback types.

seanlip:


seanlip:
Note: this PR was originally quite a bit bigger, but it got too complicated, so I'm splitting it up.

In particular, there's some existing logic in FeedbackGeneratorService.js pertaining to "unfamiliar language" feedback that hasn't been incorporated in the feedback flowchart, and needs to be accounted for. The relevant counters for this need to be moved out into ConversationManagerService, which is supposed to track state (and therefore allow FeedbackGeneratorService to be a stateless generator). I'll tackle this in a separate PR.

client/question/services/ConversationManagerService.js
```diff
@@ -78,9 +108,43 @@ tie.factory('SolutionHandlerService', [
             return CodeRunnerDispatcherService.runCodeAsync(
               language, codeSubmission.getPreprocessedCode()
             ).then(function(preprocessedCodeEvalResult) {
-              feedback = FeedbackGeneratorService.getFeedback(
-                tasks, preprocessedCodeEvalResult,
-                codeSubmission.getRawCodeLineIndexes());
+              var feedbackDetails = computeFeedbackDetails(
+                preprocessedCodeEvalResult);
+
+              // TODO(sll): Remove this case.
```
eyurko: Please add a reason why we should remove this case, just in case you get hit by a bus or something and we need to crack the code. :P

client/question/services/ConversationManagerService.js
```diff
@@ -13,18 +13,48 @@
...
+  'FeedbackDetailsObjectFactory',
   function(
       $q, CodePreprocessorDispatcherService, CodeRunnerDispatcherService,
       FeedbackGeneratorService, PrereqCheckDispatcherService,
-      TranscriptService, CodeSubmissionObjectFactory) {
+      TranscriptService, CodeSubmissionObjectFactory, FEEDBACK_CATEGORIES,
+      FeedbackDetailsObjectFactory) {
+
+    // TODO(sll): Modify this to also depend on the learner's submission
+    // history.
```
eyurko: Depend on a history how? Sorry, I'm gonna stickler on the TODOs here just in case they end up dangling.

client/question/services/ConversationManagerService.js
```diff
@@ -78,9 +108,43 @@ tie.factory('SolutionHandlerService', [
...
+              var feedbackDetails = computeFeedbackDetails(
+                preprocessedCodeEvalResult);
+
+              // TODO(sll): Remove this case.
+              if (feedbackDetails === null) {
+                feedback = FeedbackGeneratorService.getFeedback(
+                  tasks, preprocessedCodeEvalResult,
+                  codeSubmission.getRawCodeLineIndexes());
+              } else {
+                switch (feedbackDetails.getFeedbackCategory()) {
```
eyurko: oh hey, a switch statement. Haven't seen one of those in years.

Nothing wrong with this; just fun to see.

client/question/services/ConversationManagerService.js
```diff
@@ -13,18 +13,48 @@
...
+     * @returns {FeedbackDetails}
+     */
+    var computeFeedbackDetails = function(codeEvalResult) {
+      var errorString = codeEvalResult.getErrorString();
+      if (errorString) {
+        if (errorString.startsWith('TimeLimitError')) {
+          return FeedbackDetailsObjectFactory.createTimeLimitErrorFeedback();
+        } else if (
+            errorString.startsWith('ExternalError: RangeError') ||
+            errorString.includes('maximum recursion depth exceeded')) {
```
eyurko: Given how we use these in test, as well, should these string literals be constants?

client/question/services/ConversationManagerService.js
```diff
@@ -13,18 +13,48 @@
...
+    // TODO(sll): Modify this to also depend on the learner's submission
+    // history.
+
+    /**
+     * Determines the feedback details that fully characterize the feedback
+     * that the learner should receive.
+     *
+     * @returns {FeedbackDetails}
+     */
+    var computeFeedbackDetails = function(codeEvalResult) {
```
eyurko: So I'm going to end up implementing a similar version of this for the compile step, since we'll need to make sure we have similar checks for the compile step (except it would return a SyntaxError rather than a RuntimeError).

Not to put my work on you, but is there a way you could generalize this such that I can use the same method, or at least call it something that makes it obvious that it's FeedbackDetails from running a user's code, rather than compiling it?

client/question/services/ConversationManagerService.js
seanlip: Thanks! Clarified, PTAL.

client/question/services/ConversationManagerService.js
seanlip: I thought about this, and I don't think so. If we were to use them in other places in the (non-test) _code_, then making them constants makes sense for maintainability. But these are really used in just one place, and breaking them out into constants just moves the values far away from the code that uses them.

I'm fine with having to change the tests manually if the signature we're looking for ends up changing for some reason. It ensures that such a change is done intentionally.

client/question/services/ConversationManagerService.js
seanlip: Done :)

client/question/services/ConversationManagerService.js
seanlip: I know, right? I actually had a similar reaction; I generally don't go with switch/case stuff but it seemed to fit really well here.

client/question/services/ConversationManagerService.js
seanlip: Done -- added an "enum" to pass in information about the context. Thanks for catching this!

client/question/services/ConversationManagerService.js
eyurko: I'm fine with that; just wanted to make sure it was intentional. :)

client/question/domain/FeedbackDetailsObjectFactorySpec.js
```diff
@@ -0,0 +1,66 @@
...
+    it('should create a time-limit feedback details object', function() {
+      var feedbackDetails = (
+        FeedbackDetailsObjectFactory.createTimeLimitErrorFeedback());
+      expect(feedbackDetails.getFeedbackCategory()).toBe(
+        FEEDBACK_CATEGORIES.TIME_LIMIT_ERROR);
+    });
+  });
+
+  describe('stack-exceeded error feedback details', function() {
+    it('should create a stack-exceeded feedback details object', function() {
```
eyurko: nit: stack exceeded

client/question/domain/FeedbackDetailsObjectFactorySpec.js
```diff
@@ -0,0 +1,66 @@
...
+  describe('server error feedback details', function() {
+    it('should create a server error feedback details object', function() {
+      var feedbackDetails = (
+        FeedbackDetailsObjectFactory.createServerErrorFeedback());
+      expect(feedbackDetails.getFeedbackCategory()).toBe(
+        FEEDBACK_CATEGORIES.SERVER_ERROR);
+    });
+  });
+
+  describe('runtime error feedback', function() {
```
eyurko: all of the others are feedback details

in general, though, I think it makes more sense to have them be * error FeedbackDetails, to refer directly to the object. It's a nit, but I think it reads cleaner.

client/question/question.js
```diff
@@ -822,3 +822,19 @@ tie.constant('CODE_RESET_CONFIRMATION_MESSAGE', [
...
   'existing progress on the question.'
 ].join(''));
+
+/**
+ * Constant that denotes the compile-time execution context.
+ *
+ * @type {string}
+ * @constant
+ */
+tie.constant('EXECUTION_CONTEXT_COMPILATION', 'COMPILATION');
```
eyurko: Something about "Compilation" versus "Run with Tests" bothers me a smidge. Do you think it should be Compile and Run With Tests, so that they're both verbs?

client/question/domain/FeedbackDetailsObjectFactorySpec.js
seanlip: Done

client/question/domain/FeedbackDetailsObjectFactorySpec.js
seanlip: Done

client/question/question.js
seanlip: I don't think execution contexts should be verbs. I can't find a clearer way to describe them though (than what I currently have). Do you have any specific suggestions?

If not, I suggest keeping them as-is, as long as it's clear to the reader.

eyurko: yeah that's fine I suppose.

PR #599, opened 2018-07-04
Add print terminal to LearnerView.

kchen05:


client/question/components/LearnerViewDirective.js
```diff
@@ -70,23 +70,29 @@ tie.directive('learnerView', [function() {
...
+                      <ui-codemirror ng-model="editorContents.code"
+                          ui-codemirror-opts="accessibleCodeMirrorOptions"
+                          ng-change="onCodeChange()"
+                          ng-if="accessibleMode"
+                          class="protractor-test-code-input">
+                      </ui-codemirror>
+                    </div>
+                  </div>
+                  <div class="tie-print-terminal">
+                  <h1 class="tie-print-title"> Printed Output </h1>
```
seanlip: indent this and the next line by 2

client/question/components/LearnerViewDirective.js
```diff
@@ -70,23 +70,29 @@ tie.directive('learnerView', [function() {
...
+                      </ui-codemirror>
+                      <ui-codemirror ng-model="editorContents.code"
+                          ui-codemirror-opts="accessibleCodeMirrorOptions"
+                          ng-change="onCodeChange()"
+                          ng-if="accessibleMode"
+                          class="protractor-test-code-input">
+                      </ui-codemirror>
+                    </div>
+                  </div>
+                  <div class="tie-print-terminal">
```
seanlip: I thought we were doing this with a flag so that we can tweak the "display mode" as needed? I.e. by changing a constant in the frontend config we can control whether the print terminal is shown or not?

seanlip: Might also be cleaner to move this to its own directive with its own CSS classes.

client/question/question.js
```diff
@@ -514,17 +514,19 @@ tie.constant('WRONG_LANGUAGE_ERRORS', {
...
-    regexString: '\\bprint\\b',
-    message: [
-      'We noticed that you\'re using a print statement within your code. ',
-      'Since you will not be able to use such statements in a technical ',
-      'interview, TIE does not support this feature. We encourage you to ',
-      'instead step through your code by hand.'
-    ].join('')
-  }]
+  python: []
+  // The system generated tip when stdout is disabled.
```
seanlip: Either remove the code entirely or keep it in. Don't submit commented-out code.

If we're controlling the display by a flag then perhaps we will need an additional option for each tip type, along the lines of "requirePrintToBeDisabled" or similar.

client/question/components/LearnerViewDirective.js
kchen05: This ended up being quite buggy when I tried implementing it this afternoon -- I'll see if I can clean it up later.

kchen05:
image
image
@rabidbit Screenshots of the page.

client/question/question.js
seanlip: Actually I would suggest (if it's OK) to have a single field still. The problem with a dictionary is that, as you're seeing, it's already getting a bit complicated. So I would introduce this complexity only when it's clear we need it.

Also, can we not have the key be optional? It's better to have each dict have a fixed set of required keys. It makes it possible for the code to act on a standardly-formatted dict without having to check presence of the key.

rabidbit:
Layout LGTM. One nit, I don't think we need to align the output with the code (i.e., I think we can narrow the left margin so to allow more room for the output content). Thanks!

kchen05:
@eyurko @seanlip Ready for review. Should I keep this branch or make a new one since this branch still has all the commits from the print-logic branch?

seanlip:
Hi @kchen05 -- looking at this, but not sure what to review. Should I look at this or #603?

If we could close the one which isn't valid anymore, that'd be great.

Thanks!

kchen05:
My bad for the confusion! Closing this one and will address the comments on the new one tomorrow @seanlip .

PR #601, opened 2018-07-07
Add isLeapYear question.

seanlip:


assets/questions/isLeapYear.js
```diff
@@ -0,0 +1,196 @@
...
+    @classmethod
+    def handleDivisibleByFourCaseOnly(cls, year):
+        return year % 4 == 0
+`
+  },
+  tasks: [{
+    id: 'isLeapYear',
+    instructions: [{
+      content: [
+        'Three criteria must be taken into account to identify leap years:'
```
eyurko: The wording seems very ... formal, here. What about:

"To determine whether or not a given year is a leap year, we must consider three criteria:"

assets/questions/isLeapYear.js
```diff
@@ -0,0 +1,196 @@
...
+        allowedOutputs: [false]
+      }, {
+        input: 1905,
+        allowedOutputs: [false]
+      }, {
+        input: 702,
+        allowedOutputs: [false]
+      }]
+    }, {
+      id: 'DIVISIBLE_BY_FOUR_BUT_NOT_HUNDRED',
```
eyurko: BUT_NOT_ONE_HUNDRED

assets/questions/isLeapYear.js
```diff
@@ -0,0 +1,196 @@
...
+        allowedOutputs: [true]
+      }, {
+        input: 2004,
+        allowedOutputs: [true]
+      }, {
+        input: 2016,
+        allowedOutputs: [true]
+      }]
+    }, {
+      id: 'DIVISIBLE_BY_HUNDRED_BUT_NOT_FOUR_HUNDRED',
```
eyurko: ONE_HUNDRED

ditto elsewhere; I think it's more clear.

assets/questions/isLeapYear.js
```diff
@@ -0,0 +1,196 @@
...
+          'forget to include a "return" statement?'
+        ].join('')
+      ]
+    }, {
+      buggyFunctionName: 'AuxiliaryCode.returnTrue',
+      ignoredTestSuiteIds: [],
+      messages: [
+        [
+          "It looks like you're consistently returning True for all test ",
+          "cases. This implies that every year is a leap year, but that's ",
```
eyurko: This implies that every year is a leap year, which is not true.

assets/questions/isLeapYear.js
```diff
@@ -0,0 +1,196 @@
...
+          "not true..."
+        ].join('')
+      ]
+    }, {
+      buggyFunctionName: 'AuxiliaryCode.returnFalse',
+      ignoredTestSuiteIds: [],
+      messages: [
+        [
+          "It looks like you're consistently returning False for all test ",
+          "cases. This implies that there are no leap years, but that's not ",
```
eyurko: This implies there are no leap years, but that is not true.

assets/questions/isLeapYear.js
seanlip: Ah, I just followed the original question. I do prefer the updated wording, and have changed it -- thanks!

assets/questions/isLeapYear.js
seanlip: Done.

assets/questions/isLeapYear.js
seanlip: Done.

assets/questions/isLeapYear.js
seanlip: Done.

(Also, I thought "but" was the issue here, but we seem to have preserved it below, so just wanted to check what the issue was. Was it the ellipses? They were originally there to signal to the student "now it's your turn to think and continue the chain of reasoning", but I'm fine with removing them.)

assets/questions/isLeapYear.js
seanlip: Done. (Just for my edification, what was the primary issue with the original here?)

eyurko: It was the ellipses, yeah. I prefer a more authoritative tone when we're trying to like, tell them what's wrong. Trailing off seems like we're unsure. Or that's the impression I get.

PR #608, opened 2018-07-20
 Updated CodeEvalResult to use time and memory limit exceeded flags.

eyurko:


client/question/domain/CodeEvalResultObjectFactory.js
```diff
@@ -143,8 +159,7 @@ tie.factory('CodeEvalResultObjectFactory', [
      * @returns {boolean}
      */
     CodeEvalResult.prototype.hasTimeLimitError = function() {
-      var errorString = this.getErrorString();
-      return (errorString && errorString.startsWith('TimeLimitError'));
+      return this.getErrorString() && this._timeLimitExceeded;
```
seanlip: Do you actually need the this.getErrorString() bit?

client/question/domain/CodeEvalResultObjectFactory.js
```diff
@@ -154,10 +169,7 @@ tie.factory('CodeEvalResultObjectFactory', [
      * @returns {boolean}
      */
     CodeEvalResult.prototype.hasRecursionLimitError = function() {
-      var errorString = this.getErrorString();
-      return errorString && (
-        errorString.startsWith('ExternalError: RangeError') ||
-        errorString.includes('maximum recursion depth exceeded'));
+      return this.getErrorString() && this._memoryLimitExceeded;
```
seanlip: Ditto.

Also maybe rename to hasMemoryLimitError (this is a bit different from recursion limit, which is a subcase). Possibly we should change the error message shown to the user too.

client/question/services/code_evaluators/PythonCodeRunnerService.js
```diff
@@ -127,8 +127,14 @@ tie.factory('PythonCodeRunnerService', [
```
seanlip: Don't you need to add a couple more args in line 120 above?

Please double-check calls to CodeEvalResultObjectFactory.create and make sure you've updated all of them.

client/question/domain/CodeEvalResultObjectFactory.js
eyurko: I didn't update the error message -- what do you have in mind?

"Your code used more memory than we allow for this exercise; please edit your code to be more space-efficient."?

client/question/services/code_evaluators/PythonCodeRunnerService.js
eyurko: whoops yeah sorry, forgot that Javascript is chill with me just ignoring arguments. Should be fixed, now.

client/question/domain/CodeEvalResultObjectFactory.js
eyurko: Nope!

client/question/domain/CodeEvalResultObjectFactory.js
seanlip: Yes, something like that seems reasonable to me. (Though I might replace the semicolon with a period.)

Thanks!

PR #609, opened 2018-07-24
Add sendMessage function for Runestone.

kchen05:
* Adds the postMessage function so that when called, it will send the data to the Runestone parent frame.


client/question/services/RunestoneHandlerService.js
```diff
@@ -0,0 +1,31 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS-IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+/**
+ * @fileoverview Service that handles all TIE interactions with Runestone.
+ */
+tie.factory('RunestoneHandlerService', ['$window', function($window) {
```
seanlip: Make this more general -- e.g. ParentPageService or something.

client/question/services/RunestoneHandlerService.js
```diff
@@ -0,0 +1,31 @@
...
+  return {
+    /**
+     * Sends the message to the Runestone parent frame via postMessage.
+     *
+     * @param {String} messageType Type of message that is being sent.
+     * @param {String} messageData Data to be sent to parent frame.
+     */
+    sendMessage: function(messageType, messageData) {
+      $window.parent.postMessage(
+        JSON.stringify(messageData), 'https://runestone.academy');
```
seanlip: Pull target out into config constant. Set to null for standalone version. Set it to runestone in server version. Do "if" check here to determine if constant is set before making any postMessage calls.

client/question/services/RunestoneHandlerService.js
kchen05: Done.

client/question/services/RunestoneHandlerService.js
kchen05: Done.

kchen05:
@seanlip PTAL

client/config/deploymentSpecificConfig.js
```diff
@@ -21,3 +21,5 @@ tieConfig.constant('SERVER_URL', null);
 tieConfig.constant('PRIMER_DIRECTORY_URL', 'docs/');
 
 tieConfig.constant('ALLOW_PRINTING', false);
+
+tieConfig.constant('PARENT_PAGE_ORIGIN', null);
```
seanlip: PARENT_PAGE_URL? Don't know what an "origin" is.

client/question/services/ParentPageService.js
```diff
@@ -0,0 +1,44 @@
...
+tie.factory('ParentPageService', [
+  '$window', 'PARENT_PAGE_ORIGIN', function($window, PARENT_PAGE_ORIGIN) {
+    return {
+      /**
+       * Sends the message to the parent page, if it exists, via postMessage.
+       *
+       * @param {String} messageType Type of message that is being sent.
+       * @param {String} messageData Data to be sent to parent page.
+       */
+      sendMessage: function(messageType, messageData) {
```
seanlip: I think you want to make this private within the service (i.e between lines 20 and 21). Only expose public API methods like sendRawCode(rawCode), etc.

client/question/services/ParentPageService.js
```diff
@@ -0,0 +1,44 @@
...
+          $window.parent.postMessage(
+            JSON.stringify(messageData), PARENT_PAGE_ORIGIN);
+        }
+      },
+      /**
+       * Returns the origin of the parent page, if it exists. Otherwise,
+       * returns null.
+       * @return {String|null}
+       */
+      getParentPageOrigin: function() {
```
seanlip: url perhaps (see earlier comment)

client/question/services/ParentPageServiceSpec.js
```diff
@@ -0,0 +1,47 @@
...
+  var ParentPageService;
+
+  beforeEach(module('tie'));
+  var setParentPage = function(parentPageOrigin) {
+    module('tieConfig', function($provide) {
+      $provide.constant('PARENT_PAGE_ORIGIN', parentPageOrigin);
+    });
+  };
+
+  describe('getParentPageOrigin', function() {
```
seanlip: no tests for sendMessage? Maybe see if you can mock window.parent.

client/config/deploymentSpecificConfig.js
kchen05: Redefined to be PARENT_PAGE_URL_ORIGIN as the constant is only the URL origin rather than the entire URL (more info at https://developer.mozilla.org/en-US/docs/Web/Security/Same-origin_policy).

client/question/services/ParentPageService.js
kchen05: Added this as a private method and made a public method instead.

client/question/services/ParentPageService.js
kchen05: Modified.

client/question/services/ParentPageServiceSpec.js
kchen05: Added tests to verify postMessage.

client/config/deploymentSpecificConfig.js
```diff
@@ -21,3 +21,5 @@ tieConfig.constant('SERVER_URL', null);
 tieConfig.constant('PRIMER_DIRECTORY_URL', 'docs/');
 
 tieConfig.constant('ALLOW_PRINTING', false);
+
+tieConfig.constant('PARENT_PAGE_URL_ORIGIN', null);
```
seanlip: Ah OK, then PARENT_PAGE_ORIGIN is fine too (and so is the current name).

It might be helpful to add a comment to the MDN page. (`// For the definition of an origin, please see ...`)

client/question/services/ParentPageService.js
```diff
@@ -0,0 +1,54 @@
...
+    };
+
+    return {
+      /**
+       * Sends the raw user code to the parent page, if it exists.
+       *
+       * @param {String} User code to send to parent page.
+       */
+      sendRawCode: function(rawCode) {
+        sendMessage('raw code', rawCode);
```
seanlip: Define constant MESSAGE_TYPE_RAW_CODE = 'raw_code' and use that instead.

client/question/services/ParentPageService.js
```diff
@@ -0,0 +1,54 @@
...
+      },
+      /**
+       * Returns the URL origin of the parent page, if it exists. Otherwise,
+       * returns null.
+       * @return {String|null}
+       */
+      getParentPageURLOrigin: function() {
```
seanlip: nit: I think it's fine to do getParentPageUrlOrigin(). Typically I think people think of "url" as one word.

client/question/services/ParentPageServiceSpec.js
```diff
@@ -0,0 +1,91 @@
...
+
+/**
+ * @fileoverview Unit tests for ParentPageService.
+ */
+
+describe('ParentPageService', function() {
+  var ParentPageService;
+
+  beforeEach(module('tie'));
+  var setParentPage = function(parentPageURLOrigin) {
```
seanlip: setParentPageUrlOrigin

client/config/deploymentSpecificConfig.js
kchen05: Added. Left as PARENT_PAGE_URL_ORIGIN to make it more easily understood.

client/question/services/ParentPageService.js
kchen05: Fixed.

client/question/services/ParentPageServiceSpec.js
kchen05: Changed.

client/question/services/ParentPageService.js
kchen05: Added as a constant.

client/config/deploymentSpecificConfig.js
```diff
@@ -21,3 +21,9 @@ tieConfig.constant('SERVER_URL', null);
 tieConfig.constant('PRIMER_DIRECTORY_URL', 'docs/');
 
 tieConfig.constant('ALLOW_PRINTING', false);
+
+/**
+ * For the definition of an origin, please see
+ * https://developer.mozilla.org/en-US/docs/Web/Security/Same-origin_policy.
+ */
+tieConfig.constant('PARENT_PAGE_URL_ORIGIN', '*');
```
seanlip: null

kchen05: Thanks for the catch! Changed the default.

client/question/services/ParentPageService.js
```diff
@@ -0,0 +1,61 @@
...
+       */
+      sendRawCode: function(rawCode) {
+        sendMessage(MESSAGE_TYPE_RAW_CODE, rawCode);
+      },
+      /**
+       * Returns the URL origin of the parent page, if it exists. Otherwise,
+       * returns null.
+       * @return {String|null}
+       */
+      getParentPageUrlOrigin: function() {
```
seanlip: Just to check -- what external method do you envision calling this? I don't think I see the need to expose it, but I might be missing something.

kchen05: Later on I'll add the button to submit user code to the parent page, which calls this method when the button is clicked.

seanlip: Ah -- I think that may not be quite right, then.

Firstly, submitToParentPage should call sendRawCode, which it does. So that doesn't interact with this.

Secondly, there is no need for the learner view directive to know what the parent page origin is. All it needs to know is whether it's currently embedded or not (and if the embedding is "valid"). This requires three things: there actually is a parent page at the moment, the url origin we specify is not null, and the parent page matches the url origin we specify. (Note that even if a url origin is specified the page may not be embedded in the current session).

So, I would replace this with something like isIframed(). It's also fine to delete it altogether from this PR, and only add it in the next.

kchen05: That makes sense. I'll remove it for now and be sure to add it to the next PR as an isIframed method instead.

client/question/services/ParentPageService.js
```diff
@@ -0,0 +1,53 @@
...
+    var MESSAGE_TYPE_RAW_CODE = 'raw_code';
+
+    /**
+     * Sends the message to the parent page, if it exists, via postMessage.
+     *
+     * @param {String} messageType Type of message that is being sent.
+     * @param {String} messageData Data to be sent to parent page.
+     */
+    var sendMessage = function(messageType, messageData) {
+      if (PARENT_PAGE_URL_ORIGIN) {
```
seanlip: Actually wait hang on. Shouldn't this also check whether $window.parent exists?

kchen05: $window.parent will always exist since $window.parent is itself if no explicit parent is present. (See https://developer.mozilla.org/en-US/docs/Web/API/Window/parent). I'll be sure to add the checks in the next PR though regarding making sure that the $window.parent actually matches the PARENT_PAGE_URL_ORIGIN. Or is it now better to include it in this PR?

seanlip: Ah, OK. I think you might as well include it here, yep (and add tests as needed).

kchen05:
@seanlip PTAL

PR #612, opened 2018-07-26
Add full support for MemoryLimitErrors.

eyurko:


client/question/services/ConversationManagerServiceSpec.js
```diff
@@ -563,9 +563,8 @@ describe('ConversationManagerService', function() {
           var feedback = learnerViewSubmissionResult.getFeedback();
           var stdout = learnerViewSubmissionResult.getStdout();
           expect(feedback.getParagraphs()[0].getContent()).toEqual([
-            "Your code used more memory than we allow for this exercise, ",
-            "likely because your code might be hitting an infinite recursive ",
-            "loop. Check to see that your recursive calls terminate."
+            "Your code appears to be hitting an infinite recursive loop. ",
```
seanlip: Shouldn't this be a memory error (based on the description of the test in line 553)?

Probably also want to add a different, separate case for stack-exceeded error if it doesn't already exist.

eyurko: yeah we can't test the memory error; this was mislabeled.

eyurko: actually, not even mislabeled; I just forgot to save before I committed. Whoops

PR #613, opened 2018-07-27
Move buggy-output and suite-level test feedback to ConversationManagerService. Delete TranscriptService.

seanlip:
Hi @eyurko, sorry for the large PR. It was quite hard to make it smaller, since this deals with the most tangled parts of FeedbackGeneratorService. Unravelling it took a while.

I'm planning to add more tests for the newer part of the code tomorrow, but I suggest taking a look at this once you get a chance, since I don't expect any further changes to core code.

Also, please note that there's only one more migration PR after this one, which is for the residual incorrect-output feedback.

Thanks!


.eslintrc.json
```diff
@@ -59,7 +59,7 @@
         "keyword-spacing": "error",
         "line-comment-position": "error",
         "linebreak-style": "error",
-        "max-depth": "error",
```
eyurko: any particular reason for this change?

seanlip: Nesting in ConversationManagersService was one too deep.

eyurko: oh okay. Just wanted to document it somewhere.

eyurko:
Did you want to merge this PR, or are you adding more later?

seanlip:
Just need to add a few more tests for the new stuff, but that's it! (So if you have comments that are not along the lines of "add tests for X", please do make them :) )

seanlip:
Hi @eyurko, I think this is now ready for merge. PTAL?

PR #616, opened 2018-08-01
Remove question and add intro message if iframed.

kchen05:
Reference for when TIE is iframed. When the page is first loaded, the balloon flashes yellow like a normal feedback balloon, then fades to the default white: 
image

### DIFF
--- a/client/question/components/LearnerViewDirective.js
+++ b/client/question/components/LearnerViewDirective.js
@@ -28,7 +28,7 @@ tie.directive('learnerView', [function() {
               <div class="tie-question-window">
                 <div class="tie-question-container" ng-class="{'pulse-animation-enabled': pulseAnimationEnabled}" ng-attr-aria-hidden="{{MonospaceDisplayModalService.isDisplayed()}}">
                   <h1 class="tie-question-title">{{title}}</h1>
-                  <div class="tie-previous-instructions" ng-if="!isIframed">
+                  <div class="tie-previous-instructions" ng-if="!pageIsIframed">
                     <div ng-repeat="previousInstruction in previousInstructions track by $index">
                       <div ng-repeat="instruction in previousInstruction track by $index">
                         <p ng-if="instruction.type == 'text'">
@@ -39,7 +39,7 @@ tie.directive('learnerView', [function() {
                       <hr>
                     </div>
                   </div>
-                  <div class="tie-instructions" ng-if="!isIframed">
+                  <div class="tie-instructions" ng-if="!pageIsIframed">
                     <div ng-repeat="instruction in instructions">
                       <p ng-if="instruction.type == 'text'">
                         {{instruction.content}}
@@ -656,7 +656,7 @@ tie.directive('learnerView', [function() {
          * parent origin. If it is, the "Submit Code" button should be
          * displayed.
          */
-        $scope.isIframed = ParentPageService.isIframed();
+        $scope.pageIsIframed = ParentPageService.isIframed();
 
         /**
          * The ARIA alert message to show temporarily, as well as a random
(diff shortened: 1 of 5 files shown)


rabidbit:
From an aesthetics standpoint, LGTM.

client/question/components/LearnerViewDirective.js
```diff
@@ -651,6 +651,13 @@ tie.directive('learnerView', [function() {
          */
         $scope.printingIsSupported = PrintTerminalService.isPrintingSupported();
 
+        /**
+         * Defines whether TIE is currently being framed by the expected
+         * parent origin. If it is, the "Submit Code" button should be
+         * displayed.
+         */
+        $scope.isIframed = ParentPageService.isIframed();
```
seanlip: Here, we have things that are named `isIframed` being both a boolean and a function. This is a little confusing.

Perhaps you could say $scope.pageIsIframed instead to indicate that it's a boolean. Otherwise it feels like there should be parentheses at the end of that name.

client/question/components/LearnerViewDirective.js
```diff
@@ -894,6 +901,21 @@ tie.directive('learnerView', [function() {
...
+
+          // Only adds intro message if TIE is iframed, meaning the question
+          // is not shown, and there is nothing currently in the feedback
+          // window.
+          if (ParentPageService.isIframed() &&
+            SessionHistoryService.getBindableSessionTranscript().length === 0) {
+            var introMessageIfIframed = [
+              'Code your answer in the coding window. You can click the ' +
+              '"Get Feedback" button at any time to get feedback on your ' +
+              'code (will not be submitted for grading/credit). When you ' +
```
seanlip: Is there a missing "which" after the opening paren?

client/question/components/LearnerViewDirective.js
```diff
@@ -894,6 +901,21 @@ tie.directive('learnerView', [function() {
           EventHandlerService.createQuestionStartEvent();
           EventHandlerService.createTaskStartEvent(
             tasks[currentTaskIndex].getId());
+
+          // Only adds intro message if TIE is iframed, meaning the question
+          // is not shown, and there is nothing currently in the feedback
+          // window.
+          if (ParentPageService.isIframed() &&
+            SessionHistoryService.getBindableSessionTranscript().length === 0) {
+            var introMessageIfIframed = [
```
seanlip: Perhaps all this stuff should be part of the addIntroMessageBalloon() method (which could take no args) if it's always the same? No real reason for it to be here, and it's more testable if we put it in the domain object.

In general logic should go in the services/objects and not the controllers.

client/question/domain/SpeechBalloonObjectFactory.js
```diff
@@ -24,10 +24,14 @@ tie.factory('SpeechBalloonObjectFactory', [
     var SPEECH_BALLOON_TYPE_CODE = 'code';
     // A balloon representing feedback given by TIE.
     var SPEECH_BALLOON_TYPE_FEEDBACK = 'feedback';
+    // A balloon representing the intro message to be provided if the
+    // question is not shown (TIE is iframed).
+    var SPEECH_BALLOON_TYPE_INTRO = 'intro';
```
seanlip: What's the rationale for creating a separate type of balloon rather than using the existing feedback balloon? Probably worth documenting.

kchen05: Modified so that the public API distinguishes between intro, feedback, and code balloons, but intro balloons are actually inherently feedback balloons.

client/question/components/LearnerViewDirective.js
kchen05: Removed this and added it to SessionHistoryService to be handled there.

seanlip:
Thanks! LGTM. Just need to fix the merge conflicts.

PR #617, opened 2018-08-02
Show confirmation that code was submitted.

kchen05:
When "Submit for Grading" is clicked, the code balloon first shows up. Then after a second of loading dots, the confirmation message appears. The end result is shown below:

image


seanlip:
OK, @rabidbit's out. Can you fix the merge conflicts? I'll bring this in once that's done.

Thanks!

PR #618, opened 2018-08-02
Add responsive feedback window.

kchen05:
TIE with maximum width:
image

TIE with minimum width before collapsing:
image

TIE with collapsed windows:
image

I'll have this up on our trial site soon so that you can actually interact with the page.

### DIFF
--- a/client/question/components/LearnerViewDirective.js
+++ b/client/question/components/LearnerViewDirective.js
@@ -344,7 +344,7 @@ tie.directive('learnerView', [function() {
         }
         .tie-language-label {
           font-size: 12px;
-          padding: 4px 2px 0px 0px;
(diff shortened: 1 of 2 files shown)


client/question/components/LearnerViewDirective.js
```diff
@@ -346,12 +344,16 @@ tie.directive('learnerView', [function() {
         }
         .tie-language-label {
           font-size: 12px;
-          padding: 4px 10px 0px 4px;
+          padding: 4px 2px 0px 0px;
```
seanlip: Nit: can we change this to 0 not 0px?

kchen05: Changed.

PR #624, opened 2018-08-04
Simplify and refactor the process for giving incorrect-output feedback.

seanlip:
Still need to add some more tests (will do that on Monday), but this is ready for review otherwise.


client/question/services/LearnerStateService.js
```diff
@@ -71,27 +74,68 @@ tie.factory('LearnerStateService', [
...
-          feedbackCategory === previousFeedbackDetails.getFeedbackCategory() &&
-          taskIndex === previousFeedbackDetails.getTaskIndex() &&
-          specificTestIndex === previousFeedbackDetails.getSpecificTestIndex());
+        if (specificFailureCategory !== null) {
+          if (!previousFeedbackDetails.hasOutputRelatedFailure()) {
+            return 0;
+          }
+
+          // We would, normally, give the user all the specific feedback,
+          // followed by the incorrect-output feedback. Determine where in this
```
eyurko: Nit: "Determine where in this pipeline they are at" reads like a command. Maybe include a word like "First", or "Since we can't do that, we must first"

seanlip: Thanks, done!

PR #627, opened 2018-08-13
Pirate question

apicardgoog:
Initial check in of Pirate question. Note that I checked with Brad and he confirmed that punctuation and case are not relevant to this question.

I unfortunately got this code and the findMostCommonRepeatedCharacter code together in this branch.
Please review only the pirate stuff but tips on how to separate out the other question are appreciated.


apicardgoog:
Trying again on a clean branch

PR #628, opened 2018-08-13
Upgrades e2e tests to use async/await pattern

davyrisso-at-google:
Upgrades e2e tests to use async/await pattern.
Upgrades Node & Protractor versions.


test/e2e/protractor.conf.js
```diff
@@ -18,21 +18,21 @@
```
seanlip: One small request: could we rename this directory from "test" to "tests"? (Since there are, presumably, multiple tests.)

test/e2e/question.helpers.js
```diff
@@ -0,0 +1,28 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+
+/**
+ * @fileoverview Helpers for the Question page.
+ */
+
+/**
+ * This function is empty on the client but is used on the server for env setup.
```
seanlip: Document the definition of the function, not what's in it. Maybe something like:

"This function is called before XXX happens in order to set up the Protractor test environment."

test/e2e/question.helpers.js
```diff
@@ -0,0 +1,28 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS-IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+
+/**
+ * @fileoverview Helpers for the Question page.
```
seanlip: This is very vague. Maybe call it question.testConfig.js and say "Configuration details for the Protractor e2e tests for the Question page."

test/e2e/protractor.conf.js
```diff
@@ -18,21 +18,21 @@
...
+    browser.resetUrl = 'file://' + __dirname.replace('/test/e2e', '');
+    browser.baseUrl = 'file://' + __dirname.replace('/test/e2e', '');
   },
+  SELENIUM_PROMISE_MANAGER: false,
   params: {
-    questionUrl: '/client/question.html?qid=reverseWords',
-    questionsPage: require('./protractor_tests/questions.pageObject.js'),
-    utils: require('./protractor_tests/utils.js')
+    defaultQuestionId: 'reverseWords',
+    questionHelpers: require('./question.helpers.js'),
```
seanlip: questionPageTestConfig

seanlip:
Well, actually, not totally fine: the e2e tests on Travis fail.

scripts/run_e2e_tests.sh
```diff
@@ -16,17 +16,14 @@
...
-set -e
-
-source $(dirname $0)/setup.sh || exit 1
+source $(dirname $0)/setup.sh
 
 # Install the following node modules if they aren't already installed.
-install_node_module protractor 5.1.1
+install_node_module protractor 5.4.0
 
 # Start up a Selenium Server
```
seanlip: For future reference maybe explain what gecko=false is for and link to a reference site

test/e2e/protractor.conf.js
davyrisso-at-google: Done

PR #632, opened 2018-08-17
Organize the ConversationManagerServiceSpec tests according to the new feedback flow.

seanlip:
This PR organizes the tests in ConversationManagerServiceSpec along the following lines:

- The order of the tests follows that in the new feedback flow.
- Each test description is limited to a single line, for readability.
- The code has been refactored to be largely outside the tests, so that the same student code can be reused in multiple tests.
- Variables that are the same for all tests have been made global so that they are not passed in every function call.
- A single question dict is used for all tests in this file.
- Stdout tests have been separated out into a separate test suite.

I am hoping that this will make it easier to add new tests and spot holes in coverage.

### DIFF
--- a/client/question/services/ConversationManagerServiceSpec.js
+++ b/client/question/services/ConversationManagerServiceSpec.js
@@ -17,23 +17,26 @@
  */
 
 describe('ConversationManagerService', function() {
-  var SUPPORTED_PYTHON_LIBS;
   var ConversationManagerService;
   var CurrentQuestionService;
   var ExpectedFeedbackObjectFactory;
   var QuestionObjectFactory;
   var TaskObjectFactory;
-  var question;
-  var orderedTasks;
-  var auxiliaryCode;
-  var starterCode;
+
   var CORRECTNESS_STATE_INPUT_DISPLAYED;
   var CORRECTNESS_FEEDBACK_TEXT;
-  var TITLE = "title";
-  var STARTER_CODE = "starterCode";
-  var AUXILIARY_CODE = "auxiliaryCode";
+  var LANGUAGE_PYTHON;
+  var SUPPORTED_PYTHON_LIBS;
+
+  // For the following task list:
+  // - a piece of code which triggers buggy-output feedback for task 1 is one
+  //   that returns "I am buggy" for all inputs (unless all suites except the
+  //   third one are ignored).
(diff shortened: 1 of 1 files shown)


client/question/services/ConversationManagerServiceSpec.js
```diff
@@ -43,29 +39,61 @@ describe('ConversationManagerService', function() {
...
+      ]
+    }, {
+      buggyFunctionName: 'AuxiliaryCode.returnYesIfInputHasNoSpaces',
+      ignoredTestSuiteIds: [],
+      messages: [
+        'This buggy message should NOT trigger if code just returns Yes'
+      ]
+    }, {
+      buggyFunctionName: 'AuxiliaryCode.returnYesIfInputHasNoSpaces',
+      // TODO(sll): Remove GENERAL_CASE from here. This is due to an error in
```
seanlip: @eyurko, I want to draw your attention to this. When refactoring the tests, I discovered an error whereby suites in other tasks would also be considered when determining whether a test hits the buggy-output criterion. This is incorrect, because only suites _within a particular task_ should be considered when determining whether or not a given buggy output for the task is relevant.

We have not run into this issue before, because many of our questions, including the upcoming ones we're working on, are single-task. Fixing this issue is a fair amount of work because it requires an overhaul of the preprocessed code for detecting buggy output.

I am inclined to leave this as-is for the time being, and file an issue for it (in fact I'll just go ahead and do that anyway), since it is not an issue that will come up for a while (it only affects multi-task questions, and we haven't fully worked through the feedback flow for that in any case). However, I want to make sure you see this and are OK with it. If not, I'm happy to work on it and fix that in a separate PR before submitting this one.

seanlip: Filed #633.

eyurko: That's the right play, yeah.

eyurko: Just make sure we don't allow any multi-task questions before we fix this.

seanlip: Yep. I don't think we will do that before we sort out the feedback flow for multi-task questions anyway, at which point I'll need to do a similar audit to what I'm doing now. But this issue should be fixed well before then -- I know how to do it, it just takes a bit of time.

Thanks!

PR #634, opened 2018-08-21
Adds more e2e tests for the Question page

davyrisso-at-google:
- Adds more E2E tests for the Question Page:
  - reset feedback
  - switching theme
  - links display
  - responsive UI

- Replaces checkForConsoleError in utils.js by expectNoConsoleLogs: All log levels should be captured by the tests and no console output should be expected.

- Adds more utility functions for responsive UI testing

### DIFF
--- a/tests/e2e/question.pageObject.js
+++ b/tests/e2e/question.pageObject.js
@@ -14,11 +14,14 @@
 
 
 /**
- * @fileoverview Page object for the Question page.
+ * @fileoverview Question page object for Protractor E2E tests.
  */
+
+const DARK_THEME_WRAPPER_CLASS = 'night-mode';
 
 /**
  * An object that represents the Question page.
+ * Provides convenience methods to interact with the page.
  *
  * @constructor
  */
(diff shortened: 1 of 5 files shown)


tests/e2e/question.pageObject.js
```diff
@@ -27,76 +30,145 @@ var QuestionPage = function() {
...
+   * @type {webdriver.WebElement}
+   */
+  var feedbackLink = element(by.css('.protractor-test-feedback-link'));
 
   /**
    * Retrieves the TIE question page.
    *
-   * @param {string} questionId ID of the question to open.
+   * @param {string} questionId ID of the question to load.
+   * @param {boolean} setupPage Whether or not to setup the page after load.
```
seanlip: Hm, the name sounds like it represents a page object. Maybe rename to runSetupPageFunction or shouldSetupPageFunctionBeRun or similar?

(Basically, in this codebase, this change would cause setupPage to be used as both a function name and a boolean variable name. I think we should make the boolean variable more obviously boolean.)

seanlip: Actually, thinking about it more -- why do we even need this option? Are there cases when we shouldn't be setting up the page after load? It would be better if branching isn't needed.

tests/e2e/question.spec.js
```diff
@@ -14,51 +14,129 @@
...
+      let codingUiSize = await questionPage.getCodingUiSize();
+
+      let questionUiLocation = await questionPage.getQuestionUiLocation();
+      let questionUiSize = await questionPage.getQuestionUiSize();
+
+      // The right edge of the question UI should be less than the window width.
+      expect(questionUiLocation.x + questionUiSize.width).toBeLessThan(
+        windowSize.width);
+
+      // The right edge of the coding UI should be less than the window width.
```
seanlip: These tests are great, btw :) I didn't know before that it was possible to check this sort of thing in Protractor. Thanks!

tests/e2e/utils.js
```diff
@@ -12,56 +12,87 @@
...
-      if (errorFatal) {
-        fatalErrors.push(browserLogs[i]);
-      }
-    }
-  }
+/**
+ * Sets the dimensions of the browser window to those of a large screen.
+ */
+var setLargeScreen = async function() {
+  await setWindowSize(1280, 1024);
```
seanlip: @rabidbit could you please confirm that this (and below) are reasonable dimensions to be testing with? If students are going to be using a smaller screen then our e2e tests should cover that instead.

tests/e2e/question.pageObject.js
davyrisso-at-google: There is one specific case in the server version where we need to bypass page setup in order to test the presence of the privacy modal, changing the name of the argument to runSetupPageFunction

tests/e2e/question.spec.js
davyrisso-at-google: They are also fun to observe when running the tests locally!

tests/e2e/utils.js
davyrisso-at-google: Acknowledged, awaiting @rabidbit validation. Looking at the CSS in LearnerViewDirective.js the breakpoint is set at width <= 1058px.

davyrisso-at-google:
Fixed ambiguous param name for questionPage.get()

tests/e2e/utils.js
rabidbit: Sorry, what's the context? Is this the minimum size representing someone on a desktop with a large window (i.e., maximum width)? Is this for the stand-alone or embedded version?

rabidbit: I see. If you are just testing to see if elements are re-arranged after window re-size, then I think 1070 x 700 (wide enough to see elements in two columns) to 1050 x 700 (narrow enough to see elements in one column) should work.

davyrisso-at-google: Done. Thanks!

PR #636, opened 2018-08-21
 Added limit to how many CodeSubmitEvents could be in an event batch.

eyurko:


client/question/services/EventHandlerService.js
```diff
@@ -27,6 +27,7 @@ tie.factory('EventHandlerService', [
     var _currentSessionId = null;
     var _currentQuestionId = null;
     var _currentQuestionVersion = null;
+    var MAX_NUM_CODE_SUBMIT_EVENTS_PER_BATCH = 4;
```
seanlip: We might want to declare this constant next to where the batch timeframe is declared, since those two constants tend to be related to each other.

eyurko: fine with that

eyurko: done!

PR #639, opened 2018-08-24
Add two more buggy output tests to findMostCommonCharacter question.

apicardgoog:
One issue I ran into testing this: it would be great for the performance tests to allow for custom messages because "try making this linear" is not super-helpful for a beginner who hasn't learned O yet. I'd like to say something like: "It looks like your code is slow for large inputs. Are you going over the string multiple times? Using a dictionary might allow you to solve this problem with a single pass over the string." Or something like that.


seanlip:
Could we just remove the performance tests part of the schema? I think that would be fine, because that part of the code isn't really fleshed out yet in TIE anyway and might lead to false positives/negatives.

apicardgoog:
PTAL

seanlip:
Thanks!

PR #642, opened 2018-08-28
Bugfix: Code Window code line margin collapse when in collapsed IFrame 

michaelrobbe:
Refreshes UI in the case that window.innerWidth/.innerHeight = 0, like a collapsed IFrame.

### DIFF
--- a/client/question/components/LearnerViewDirective.js
+++ b/client/question/components/LearnerViewDirective.js
@@ -1312,15 +1312,14 @@ tie.directive('learnerView', [function() {
         }
 
         /**
-         * Refreshes UI if the window loads at 0 height or width (IFrame Firefox
+         * Refreshes UI if window loads at 0 height or width (Firefox IFrame
            Case)
          */
-        var millisecondDelay = 500;
-        $interval(function() {
+        $timeout(function() {
           if ($window.innerWidth === 0 || $window.innerHeight === 0) {
             codemirrorEditorInstance.refresh();
           }
-        }, millisecondDelay, 1);
+        });
       }
     ]
   };


client/question/components/LearnerViewDirective.js
```diff
@@ -1310,6 +1310,17 @@ tie.directive('learnerView', [function() {
...
+        /**
+         * Refreshes UI if the window loads at 0 height or width (IFrame Firefox
+           Case)
+         */
+        var millisecondDelay = 500;
+        $interval(function() {
+          if ($window.innerWidth === 0 || $window.innerHeight === 0) {
+            codemirrorEditorInstance.refresh();
+          }
+        }, millisecondDelay, 1);
```
seanlip: Is the idea to run this only once? If so, then $timeout might be better.

On the other hand, if the idea is to run this multiple times until the condition holds, then you can use $interval but then the $interval should be cancelled once the conditions are successful.

client/question/components/LearnerViewDirective.js
```diff
@@ -1310,6 +1310,17 @@ tie.directive('learnerView', [function() {
             $scope.privacyModalIsDisplayed = true;
           }
         }
+
+        /**
+         * Refreshes UI if the window loads at 0 height or width (IFrame Firefox
+           Case)
+         */
+        var millisecondDelay = 500;
```
seanlip: I think it's fine to inline this, since it's not reused anywhere and the name doesn't have an intrinsic meaning. Or alternatively rename it to something like initialIntervalDelayMsecs.

client/question/components/LearnerViewDirective.js
michaelrobbe: Done. Replaced with $timeout, equally effective.

client/question/components/LearnerViewDirective.js
michaelrobbe: Done. Removed the named variable as it wasn't necessary with $timeout.

michaelrobbe:
Replaced $interval with $timeout. Worked perfectly. Reposted.

PR #644, opened 2018-08-29
CodeResetEvents should submit the event batch to the backend.

eyurko:
#643 but without the CLA issues.


eyurko:
new session ID is already created --> LearnerViewDirective.js:1226 calls initLearnerViewDirective().

PR #649, opened 2018-09-06
Allowed regex to be tolerant of the empty string.

eyurko:


client/question/question.js
```diff
@@ -639,7 +639,7 @@ tie.constant('RUNTIME_ERROR_FEEDBACK_MESSAGES', {
       return errorString.startsWith('KeyError: ');
     },
     generateMessage: function(errorString) {
-      var keyErrorRegEx = /KeyError:\s(\w+)\s/;
+      var keyErrorRegEx = /KeyError:\s('.*'|\w+)\s/;
```
seanlip: Can we add a more deterministic indicator, like "on line" at the end?

This helps guard against cases where the key itself contains spaces.

Also I'm not sure you need the \w+ part?

seanlip: I don't think the existing test is s a valid test, actually. If you look at the message, it doesn't realy make sense, it seems like it was done that way just for expediency and seems to be several error messages smushed together.

Probably don't need to require quotes in the regex.

client/question/questionSpec.js
```diff
@@ -95,8 +95,18 @@ describe('question.js', function() {
...
+          'defined KeyError: key blah on line 47 AttributeError: \'key\' ',
+          'object has no attribute \'length\''].join(''))).toEqual('string');
+      });
+    });
+
+    it('should handle an empty key', function() {
+      var errorString = 'KeyError: key "" on line 47';
+      RUNTIME_ERROR_FEEDBACK_MESSAGES.python.forEach(function(error) {
+        if (error.checker(errorString)) {
+          expect(typeof error.generateMessage(
```
seanlip: This should be: expect(error.generateMessage(...)).toEqual(...) since the point is to test the generated string.

PR #654, opened 2018-10-02
Increases timeout for alert dialog presence testing

davyrisso-at-google:


seanlip:
LGTM, but I think 1 min feels a bit long. Would 30 seconds be fine? Just thinking about what counts as "reasonable" loading time.

FInal decision is up to you, this is just a data point. Feel free to merge once you've decided.

PR #658, opened 2018-10-10
Add support for server-side 404 page.

eyurko:
Should be fine to merge, but any servers using this would have to implement the server-side support alongside the merge otherwise they'll crash.

### DIFF
--- a/client/question/components/LearnerViewDirective.js
+++ b/client/question/components/LearnerViewDirective.js
@@ -920,6 +920,10 @@ tie.directive('learnerView', [function() {
           // Load the feedback, tasks, and stored code and initialize the
           // event services.
           var questionId = CurrentQuestionService.getCurrentQuestionId();
+          if (questionId === null) {
+            $window.location.href = '/client/404.html';
+            return;
+          }
           var question = CurrentQuestionService.getCurrentQuestion();
           var tasks = question.getTasks();
           currentTaskIndex = 0;
--- a/client/question/components/LearnerViewDirective.js
+++ b/client/question/components/LearnerViewDirective.js
@@ -795,6 +795,10 @@ tie.directive('learnerView', [function() {
 
         $scope.onVisibilityChange = function() {
           var question = CurrentQuestionService.getCurrentQuestion();
+          // If the question is null (such as if there's an error getting it)
+          // then we 404 on a server version. If that happens, this throws a
+          // mysterious console error since we call null.getTasks(), which is
+          // bad. This prevents that error.
           if (question) {
             var tasks = question.getTasks();
 
--- a/client/question/components/LearnerViewDirective.js
+++ b/client/question/components/LearnerViewDirective.js
@@ -795,20 +795,22 @@ tie.directive('learnerView', [function() {
 
         $scope.onVisibilityChange = function() {
           var question = CurrentQuestionService.getCurrentQuestion();
-          var tasks = question.getTasks();
+          if (question) {
+            var tasks = question.getTasks();
 
-          // When a user changes tabs (or comes back), add a SessionPause
-          // or SessionResumeEvent, respectively.
-          var hiddenAttributeName = (
-            $scope.determineHiddenAttributeNameForBrowser());
-          if (hiddenAttributeName !== null && tasks !== null &&
-              tasks.length > currentTaskIndex) {
-            if ($scope.isDocumentHidden(hiddenAttributeName)) {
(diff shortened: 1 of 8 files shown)


client/question/services/CurrentQuestionServiceSpec.js
```diff
@@ -90,3 +91,56 @@ describe('CurrentQuestionService', function() {
...
+    });
+  };
+
+  describe('behavior for invalid question - server', function() {
+
+    it('should return null', function(done) {
+      setServerUrl('http://katamari-dam.acy');
+      setDefaultQuestionId(null);
+
+      var PAGE_NOT_FOUND_ERROR = 404;
```
seanlip: I don't understand the point of having two separate constants here. They seem to be for post requests to the same URL. Why are they named differently?

client/question/services/CurrentQuestionServiceSpec.js
```diff
@@ -90,3 +91,56 @@ describe('CurrentQuestionService', function() {
...
+        // Set up a non-existent question ID. This must be done before
+        // CurrentQuestionService is initialized.
+        $location = $injector.get('$location');
+        $rootScope = $injector.get('$rootScope');
+        $httpBackend = $injector.get('$httpBackend');
+        $location.search('qid', 'invalidQuestion - server');
+
+        CurrentQuestionService = $injector.get('CurrentQuestionService');
+      });
+      $httpBackend.expectPOST(
```
seanlip: Might want to give some commentary about why two POSTs are expected.

Also, I'm not certain, but don't you need an $httpBackend.flush() at some point? And something like 

```
        $httpBackend.verifyNoOutstandingExpectation();
        $httpBackend.verifyNoOutstandingRequest();
```

client/question/components/LearnerViewDirective.js
```diff
@@ -795,20 +795,22 @@ tie.directive('learnerView', [function() {
...
-            $scope.determineHiddenAttributeNameForBrowser());
-          if (hiddenAttributeName !== null && tasks !== null &&
-              tasks.length > currentTaskIndex) {
-            if ($scope.isDocumentHidden(hiddenAttributeName)) {
-              EventHandlerService.createSessionPauseEvent(
-                tasks[currentTaskIndex].getId());
-            } else {
-              EventHandlerService.createSessionResumeEvent(
-                tasks[currentTaskIndex].getId());
+          if (question) {
```
seanlip: What's the plan in the "else" case?

client/question/services/CurrentQuestionServiceSpec.js
eyurko: Definitely! Thanks for the catch.

client/question/components/LearnerViewDirective.js
eyurko: Nothing should happen? That means the question isn't initialized yet and trying to put a task ID for an uninitialized question is undefined behavior.

I could set the task id to be "pre-question", if you want, but I'm not sure that would be captured properly in the analytics. Whatcha think?

eyurko: This fix is mostly for testing -- the pages manage to switch before the question is fully loaded (or it doesn't load, in some cases), which cause getTasks() to be called on a null.

client/question/services/CurrentQuestionServiceSpec.js
eyurko: Added comments to explain. :)

eyurko: that's just my bad -- sorry, thought I changed that to a 400.

client/question/services/CurrentQuestionService.js
```diff
@@ -39,6 +39,8 @@ tie.factory('CurrentQuestionService', [
             cachedQuestion = question;
             serviceIsInitialized = true;
             successCallback();
+          } else if (questionId === DEFAULT_QUESTION_ID) {
+            successCallback();
```
seanlip: One last check: should this be calling successCallback()? It seems like a clear error case.

eyurko: yeah -- it still needs to go through the motions of initializing the LearnerViewDirective, otherwise it doesn't call the redirect.

seanlip: OK. Seems misleading though; add comment in the code to explain please?

seanlip:
LGTM. Thanks!

PR #660, opened 2018-10-19
Add support for checking console errors.

eyurko:


tests/e2e/utils.js
```diff
@@ -51,6 +51,28 @@ var expectNoConsoleLogs = async function() {
   expect(consoleLogs.length).toBe(0);
 };
 
+/**
+ * Convenience function used to check the number / content of console errors.
+ *
+ * @param {number} num The number of errors to expect.
+ * @param {Array} stringsToMatch A list of strings to check logs for.
```
seanlip: Perhaps say something about this needing to be in the same order as the expected order of the console logs.

eyurko: yeah that's pretty reasonable.

PR #693, opened 2018-11-14
Fix #692: update fileoverview for FeedbackGeneratorService.

### LINKED ISSUE #692: Update file overview for FeedbackGeneratorService.js
Currently says:

/**
 * @fileoverview Service for generating feedback based on the evaluation result
 * of the code submitted by the user, preprocessed and augmented with unit
 * tests. This assumes that the base code already passes "compile-time" checks.
 */

But per offline discussion, this may no longer assume code passed compile. The description could also benefit from distinguishing its use vs FeedbackDetailsObjectFactory.js when it comes to injection of novice-friendly syntax error messages (i.e., FeedbackGeneratorService is where such transformations should occur)--right now it says "generating feedback", which can be confusing given the generation of feedback objects in FeedbackDetailsObjectFactory.

seanlip:


client/question/services/FeedbackGeneratorService.js
```diff
@@ -13,9 +13,9 @@
 // limitations under the License.
 
 /**
- * @fileoverview Service for generating feedback based on the evaluation result
- * of the code submitted by the user, preprocessed and augmented with unit
- * tests. This assumes that the base code already passes "compile-time" checks.
+ * @fileoverview Service for creating human-readable feedback messages based
```
rabidbit: I think "more informative (friendly) error messages" would be better. "human-readable" implies the original errors were not human-readable, when they are (they just aren't very informative or friendly to novices).

seanlip: Hm, I think that's not the right context for this fileoverview, though, which isn't restricted to just syntax errors. In general, what the functions in this file do is that they transform a FeedbackDetails object (which is, in general, not human-readable) to something that's human readable.

rabidbit: I see. I misunderstood our offline chat. OK.

PR #694, opened 2018-11-14
Add a question-level config option for enabling friendly syntax errors.

seanlip:


rabidbit:
Ha, I was doing almost the same thing in my upcoming PR, except I called it enableFriendlyErrors, because runtime errors are also modified, right?

PR #698, opened 2018-11-30
Add friendly syntax error messages

rabidbit:


client/question/question.js
```diff
@@ -550,101 +550,195 @@ tie.constant(
...
+      friendlyErrorCheck: function(errorString) {
+        var re = new RegExp('^NameError: global name .* is not defined$');
+        return re.test(errorString);
+      },
+      getFriendlyErrorText: function(errorString) {
+        var re = new RegExp('NameError: global name (.*) is not defined');
+        var varName = errorString.match(re)[1];
+        return [
+          '<code>', varName, '</code> appears to be a name that was not ',
+          'previously declared or defined. E.g., if <code>', varName,
```
seanlip: (Optional) This seems a little verbose IMO -- the most likely issue here is a typo of the method/function/variable/etc. name. What do you think of something shorter, like:

```
    It looks like varName wasn't previously declared or defined. Did you type it
    incorrectly, here or in the rest of the code? (Note that using consistent
    capitalization is also important.)
```

That said, please feel free to ignore this suggestion if you prefer the original.

client/question/question.js
```diff
@@ -550,101 +550,195 @@ tie.constant(
...
+      friendlyErrorCheck: function(errorString) {
+        var re = new RegExp('^SyntaxError: invalid syntax');
+        return re.test(errorString);
+      },
+      getFriendlyErrorText: function() {
+        return [
+          'This is the default, generic syntax error message in Python. Here ',
+          'are some common mistakes that can result in this error:',
+          '<ul>',
+          '<li>Mis-matching (missing or too many) braces, brackets, ',
```
seanlip: The closing li tags are all missing.

client/question/question.js
```diff
@@ -550,101 +550,195 @@ tie.constant(
...
+          'code blocks.'].join('');
+      }
+    }, {
+      friendlyErrorCheck: function(errorString) {
+        var re = new RegExp('^TypeError: can only concatenate');
+        return re.test(errorString);
+      },
+      getFriendlyErrorText: function() {
+        return [
+          'You may have tried to combine two incompatible types (you might ',
```
seanlip: Optional: Perhaps add "e.g." at the beginning of the parenthetical phrase, and strike "etc." at the end. (I think this is trying to convey "for example" rather than "and others".)

client/question/question.js
```diff
@@ -550,101 +550,195 @@ tie.constant(
...
+ * the given error
+ *
+ * @type {Object}
  **/
-tie.constant('RUNTIME_ERROR_FEEDBACK_MESSAGES', {
+tie.constant('FRIENDLY_RUNTIME_ERROR_TRANSLATIONS', {
   python: [{
     // Error when user indents lines incorrectly
-    checker: function(errorString) {
+    friendlyErrorCheck: function(errorString) {
```
seanlip: This is interesting, it appears in the syntax error checks too. Should we remove it in one of them? If not, should the error messages be consistent?

(If de-duplicating, it might be worth noting that the error messages used in this section have already undergone language review.)

client/question/question.js
```diff
@@ -550,101 +550,195 @@ tie.constant(
...
+    getFriendlyErrorText: function(errorString) {
       var nameErrorRegEx = (
-        /NameError:\s(?:global\s)?name\s'(\w+)'\sis\snot\sdefined/);
+          /NameError:\s(?:global\s)?name\s'(\w+)'\sis\snot\sdefined/);
       var found = errorString.match(nameErrorRegEx);
-      return ["It looks like " + found[1] + " isn't a declared variable. ",
-        "Did you make sure to spell it correctly? Is it correctly ",
-        "initialized?"].join('');
+      return [
+        'It looks like <span class="tie-code-text">', found[1], '</span> ',
```
seanlip: Consider using `<code>` rather than `<span class>` here.

client/question/questionSpec.js
```diff
@@ -90,28 +93,68 @@ describe('question.js', function() {
...
+  describe('FRIENDLY_SYNTAX_ERROR_TRANSLATIONS', function() {
     it('should have the correct keys and valid values', function() {
-      RUNTIME_ERROR_FEEDBACK_MESSAGES.python.forEach(function(error) {
-        expect(error.checker('test')).toBe(false);
-        expect(typeof error.generateMessage(['NameError: name \'key\' is not ',
-          'defined KeyError: blah on line 47 AttributeError: \'key\' ',
-          'object has no attribute \'length\''].join(''))).toEqual('string');
+      FRIENDLY_SYNTAX_ERROR_TRANSLATIONS.python.forEach(function(error) {
+        expect(error.friendlyErrorCheck('test')).toBe(false);
+        // If any getFriendlyErrorText method that takes in an error string
```
seanlip: The "if" clause of this comment seems to be missing one or more words.

client/question/questionSpec.js
```diff
@@ -90,28 +93,68 @@ describe('question.js', function() {
...
+            'matched: ' + errorString);
+      }
+    });
+  });
+
+  describe('FRIENDLY_RUNTIME_ERROR_TRANSLATIONS', function() {
+    it('should have the correct keys and valid values', function() {
+      FRIENDLY_RUNTIME_ERROR_TRANSLATIONS.python.forEach(function(error) {
+        expect(error.friendlyErrorCheck('test')).toBe(false);
+        // If any getFriendlyErrorText method that takes in an error string
```
seanlip: Ditto.

client/question/services/ConversationManagerServiceSpec.js
```diff
@@ -411,54 +435,54 @@ describe('ConversationManagerService', function() {
...
         code: codeThatFailsBuggyOutputForTaskOne,
         expectedFeedback: ExpectedFeedbackObjectFactory.create(
-          'Mock BuggyOutputTest Message Three for task1')
+          ['Mock BuggyOutputTest Message Three for task1'])
       }, {
         // At this point, we have run out of buggy-output test feedback.
         code: anotherCodeThatFailsBuggyOutputForTaskOne,
         expectedFeedback: ExpectedFeedbackObjectFactory.create(
-          null, null, '', false, function(submissionResult) {
+          null, '', false, function(submissionResult) {
```
seanlip: If we're going to have the first arg be all the feedback, then the first arg should be an empty list if no feedback is expected (not null).

client/question/services/ConversationManagerServiceTestUtils.js
```diff
@@ -79,38 +70,30 @@ tie.factory('ExpectedFeedbackObjectFactory', [function() {
    */
   ExpectedFeedback.prototype.verifyFeedback = function(submissionResult) {
     var errorMessages = [];
+    var observedFeedback = submissionResult.getFeedback();
+    var observedFeedbackParagraphObject = observedFeedback.getParagraphs();
```
seanlip: Variable name sounds singular, but getParagraphs() presumably returns a list. Consider renaming to observedFeedbackParagraphs.

client/question/services/ConversationManagerServiceTestUtils.js
```diff
@@ -79,38 +70,30 @@ tie.factory('ExpectedFeedbackObjectFactory', [function() {
...
-    }
-
-    if (this._secondFeedbackParagraphContent !== null) {
-      observedContent = feedback.getParagraphs()[1].getContent();
-      expectedContent = this._secondFeedbackParagraphContent;
-      if (observedContent !== expectedContent) {
-        errorMessages.push(
-          'Bad content for second feedback paragraph: expected "' +
-          expectedContent + '" but received "' + observedContent + '"');
+    for (var i in this._expectedFeedbackParagraphs) {
```
seanlip: If i is an index, use "for (var i = 0; i < this._expectedFeedbackParagraphs.length; i++)" or use forEach. In general, use "for (var i in XXX)" only for iterating over keys in a dict.

seanlip: Seems a bit odd to me that this works when this._expectedFeedbackParagraphs is null, btw (hence the comment elsewhere about making it []).

client/question/services/ConversationManagerServiceTestUtils.js
```diff
@@ -79,38 +70,30 @@ tie.factory('ExpectedFeedbackObjectFactory', [function() {
...
+      if (this._expectedFeedbackParagraphs[i] !== null) {
```
seanlip: Perhaps explain significance of a null entry in the constructor arg docstring.

client/question/services/ConversationManagerServiceTestUtils.js
```diff
@@ -129,11 +112,8 @@ tie.factory('ExpectedFeedbackObjectFactory', [function() {
   /**
    * Creates and returns an ExpectedFeedback object.
    *
-   * @param {string|null} firstFeedbackParagraphContent The expected content of
-   *   the first feedback paragraph, or null if we do not need to check this.
-   * @param {string|null} secondFeedbackParagraphContent The expected content
-   *   of the second feedback paragraph, or null if we do not need to check
-   *   this.
+   * @param {Array<string>|null} expectedFeedbackParagraphs The content of
+   *   expected feedback paragraphs.
```
seanlip: Ditto re explaining the significance of null.

client/question/services/FeedbackGeneratorService.js
```diff
@@ -157,11 +173,11 @@ tie.factory('FeedbackGeneratorService', [
      * @param {string} language
      * @returns {string | null} Text to be appended to feedback.
      */
-    var _getHumanReadableRuntimeFeedback = function(errorString, language) {
+    var _getFriendlyRuntimeFeedback = function(errorString, language) {
       var result = null;
-      RUNTIME_ERROR_FEEDBACK_MESSAGES[language].forEach(function(check) {
-        if (check.checker(errorString)) {
-          result = check.generateMessage(errorString);
+      FRIENDLY_RUNTIME_ERROR_TRANSLATIONS[language].forEach(function(check) {
```
seanlip: I think I saw some forEach loops above where the dummy variable was 'error' rather than 'check', e.g. https://github.com/google/tie/pull/698/files#diff-b5e458ce6bd4b4d8d91bb228b3182735R111 .  We might want to standardize.

(I think this was an issue in the old version of the code, too.)

client/question/services/FeedbackGeneratorService.js
```diff
@@ -402,50 +423,36 @@ tie.factory('FeedbackGeneratorService', [
...
-        var fixedErrorString = errorString.replace(
-          new RegExp('line ([0-9]+)$'), function(_, humanReadableLineNumber) {
-            var preprocessedCodeLineIndex = (
-              Number(humanReadableLineNumber) - 1);
-            if (preprocessedCodeLineIndex < 0 ||
-                preprocessedCodeLineIndex >= rawCodeLineIndexes.length) {
-              throw Error(
-                'Line number index out of range: ' + preprocessedCodeLineIndex);
-            }
-            if (rawCodeLineIndexes[preprocessedCodeLineIndex] === null) {
```
seanlip: I think you still need this logic. It helps catch cases where the error arises in the preprocessed code.

client/question/services/FeedbackGeneratorServiceSpec.js
```diff
@@ -406,19 +412,87 @@ describe('FeedbackGeneratorService', function() {
...
+      expect(feedbackString).toEqual([
+        '<code>foo</code> appears to be a name that was not previously ',
+        'declared or defined. E.g., if <code>foo</code> is a variable to ',
+        'hold a string, <code>foo = \'\'</code> would declare ',
+        '<code>foo</code> as a string variable. Similarly, ',
+        '<code>foo = []</code> declares <code>foo</code> as an array ',
+        'variable. Another possibility is that <code>foo</code> was ',
+        'misspelled or uses incorrect capitalization.'].join(''));
+    });
+    it(['should return the correct feedback string if ',
```
seanlip: Add newlines between it() blocks, similar to the rest of the file.

client/question/services/FeedbackGeneratorServiceSpec.js
```diff
@@ -567,12 +643,12 @@ describe('FeedbackGeneratorService', function() {
 
       expect(paragraphs.length).toEqual(3);
       expect(paragraphs[0].isTextParagraph()).toBe(true);
-      expect(paragraphs[0].getContent()).toEqual(
-          'It looks like your code has a syntax error. ' +
-          'Try to figure out what the error is.');
-      expect(paragraphs[1].isErrorParagraph()).toBe(true);
+      expect(paragraphs[0].getContent()).toMatch(
+          'Error detected(?: on or near line [0-9]+)*:');
```
seanlip: Probably better to do a toEqual() check, since this is a test. The more explicit the better.

Ditto elsewhere.

client/question/domain/CodeEvalResultObjectFactory.js
```diff
@@ -357,6 +357,21 @@ tie.factory('CodeEvalResultObjectFactory', [
       return this._performanceTestResults;
     };
 
+    /**
+     * Returns a number representing the line number where an error was
+     * detected in the student code (if there is one).
+     * The function should return a number if there is an error traceback in the
+     * code results. Otherwise, it should return null.
+     *
+     * @returns {number}
```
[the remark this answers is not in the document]
rabidbit: ACK Fixed in upcoming commit.

client/question/domain/CodeEvalResultObjectFactorySpec.js
```diff
@@ -607,6 +607,22 @@ describe('CodeEvalResultObjectFactory', function() {
     });
   });
 
+  describe('getErrorLineNumber5', function() {
```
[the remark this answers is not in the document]
rabidbit: Fixed in upcoming commit.

client/question/domain/ErrorTracebackObjectFactory.js
```diff
@@ -91,7 +91,7 @@ tie.factory('ErrorTracebackObjectFactory', [
         !this._tracebackCoordinates) {
         return this._errorMessage;
       }
-      return this._errorMessage + ' on line ' + this._getFirstTracebackLine();
+      return this._errorMessage;
```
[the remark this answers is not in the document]
rabidbit: Fixed in upcoming commit. Re: docstring. I'm hoping the changes elsewhere to the codebase should eliminate the expectation that a number will be returned. I wonder if calling that out here would cause some confusion.

client/question/question.js
```diff
@@ -550,101 +550,195 @@ tie.constant(
...
+    {
+      friendlyErrorCheck: function(errorString) {
+        var re = new RegExp('^NameError: global name .* is not defined$');
+        return re.test(errorString);
+      },
+      getFriendlyErrorText: function(errorString) {
+        var re = new RegExp('NameError: global name (.*) is not defined');
+        var varName = errorString.match(re)[1];
+        return [
+          '<code>', varName, '</code> appears to be a name that was not ',
```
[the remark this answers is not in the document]
rabidbit: There are other parts of the text that uses <code> as well, e.g., ' = []' and '= \'\''. I feel like it's more readable as is rather than creating multiple variables.

client/question/question.js
rabidbit: It might be verbose to someone familiar with Python and coding, yes. I think it's a good opportunity to remind the novice how to "declare" a variable.

client/question/question.js
rabidbit: Some style guides say this is preferable (shrug). I can add the closing tags if you prefer though.

client/question/question.js
rabidbit: Hmm... Yeah, I originally had "e.g.", but I wanted to emphasize that there are other examples. The concern being that the novice would focus too much on the example given.

client/question/question.js
rabidbit: I noticed that too, but I viewed it as out of scope for this (rather lengthy)PR (i.e., revisit how runtime errors work versus syntax and update translations appropriately). For now, I can at least make them consistent.

client/question/questionSpec.js
rabidbit: There were actually too many words :-P. Changing it to:

"If a getFriendlyErrorText method takes in an error string as input, add a test below."

Most getFriendlyErrorText methods do not take in any input; this is a reminder to add a test for those that do.

client/question/questionSpec.js
rabidbit: Same as above.

client/question/services/ConversationManagerServiceTestUtils.js
rabidbit: Actually... it's an array of feedback paragraph objects; so, it seems "Object" should be in the name. Would observedFeedbackParagraphsObject work?

client/question/domain/ErrorTracebackObjectFactory.js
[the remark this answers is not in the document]
seanlip: I think it's fine to update the docstring, because Python typically returns a line number with the error, so the behaviour here is a bit atypical relative to that.

client/question/question.js
seanlip: I'm happy to defer here, per offline conversation.

client/question/question.js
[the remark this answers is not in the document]
seanlip: I disagree, but only slightly. I'm happy for you to make the call.

client/question/question.js
seanlip: Hm, interesting. Yes, I would prefer having the closing tags, if that's OK.

client/question/question.js
seanlip: Hmm, I see what you mean. I'm fairly neutral on this one; your call.

client/question/question.js
seanlip: SG re consistency.

client/question/questionSpec.js
seanlip: That sounds good. Thanks!

client/question/services/ConversationManagerServiceTestUtils.js
seanlip: Typically, we omit the "object" when referring to domain objects by name, so I think observedFeedbackParagraphs is better here and more consistent with the rest of the codebase. (Otherwise you'll also need to do getParagraphObjects() etc.)

(Note also that it would be Object**s** in that case.)

client/question/domain/ErrorTracebackObjectFactory.js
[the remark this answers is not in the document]
rabidbit: ACK Fixed in upcoming commit.

client/question/services/ConversationManagerServiceTestUtils.js
rabidbit: ACK Fixed in upcoming commit.

client/question/services/FeedbackGeneratorServiceSpec.js
rabidbit: ACK Fixed in upcoming commit.

client/question/services/FeedbackGeneratorServiceSpec.js
rabidbit: toEqual() matches a specific string or value. Here, I'm matching a regex. Is the suggestion that I can use toEqual to match on a regex (which I'm not sure would work) or not use regex (but then, we would not be matching the more complete pattern)?

seanlip: Oh -- I think I got thrown off by "not matching the more complete pattern". I was just talking about a equality match to the exact text of the error string that should be shown to the user (including the line number information and everything), since that shouldn't change even if the underlying code changes.

Generally, if you want to test various cases of a pattern, try to write a range of tests that are exact-match checks to individual specific cases for that pattern. That's better than testing against the pattern itself, which is sort of "internal" and more of an implementation detail. It also catches errors where the wrong error message is given for a particular case (but it still happens, by chance, to match the regex pattern).

rabidbit: OK. Fixed for the next commit.

client/question/question.js
rabidbit: OK. Fixed in next commit.

client/question/question.js
rabidbit: Done in next commit.

client/question/services/ConversationManagerServiceTestUtils.js
rabidbit: Fixed in upcoming commit.

rabidbit: This is because <code>for (var i in null)</code> is valid but <code>for (var i = 0; i < null.length; i++)</code> is not

client/question/services/ConversationManagerServiceTestUtils.js
rabidbit: Done in next commit.

client/question/services/ConversationManagerServiceTestUtils.js
rabidbit: Done in next commit.

rabidbit:
@seanlip PTAL

client/question/services/FeedbackGeneratorService.js
seanlip: @rabidbit FYI, this comment hasn't been replied to.

client/question/services/FeedbackGeneratorService.js
```diff
@@ -323,12 +345,10 @@ tie.factory('FeedbackGeneratorService', [
           throw new Error(['Unrecognized prereq check failure type ',
             'in getPrereqFailureFeedback().'].join());
         }
-
```
seanlip: Just a note -- I want to caution about deleting all these lines, here and elsewhere in this file (see the diff in "Files Changed" on github). They were there in this file to separate blocks of code from each other so that it's easier to read / more consumable.

I'm not strongly against their deletion (it doesn't really matter either way, code-wise), but if it's unintentional, please consider reinstating.

client/question/services/FeedbackGeneratorService.js
seanlip: FYI @rabidbit -- this comment was not responded to.

Consider looking at the comment history or "Files Changed" in the future, in order to make sure no existing comments are outstanding, before requesting review.

client/question/services/FeedbackGeneratorService.js
```diff
@@ -402,50 +422,36 @@ tie.factory('FeedbackGeneratorService', [
...
+        var friendlyRuntimeFeedbackString = _getFriendlyRuntimeFeedback(
+            errorString, language);
+        if (friendlyRuntimeFeedbackString) {
+          feedback.appendTextParagraph(friendlyRuntimeFeedbackString);
         } else {
           feedback.appendTextParagraph(
-            'Looks like your code had a runtime error when evaluating the ' +
-            'input ' + _jsToHumanReadable(errorInput) + '.');
-          feedback.appendErrorParagraph(fixedErrorString);
+              'Looks like your code had a runtime error when evaluating the ' +
```
seanlip: Here and elsewhere -- I understand that indentation by 4 is idiomatic in some styles, but most of the codebase goes with 2-space indentation for everything. I think we should either fix indentation in the whole codebase or keep the existing convention throughout, rather than end up with a half-and-half.

client/question/services/FeedbackGeneratorServiceSpec.js
```diff
@@ -721,11 +776,15 @@ describe('FeedbackGeneratorService', function() {
...
       expect(paragraphs[1].isErrorParagraph()).toBe(true);
       expect(paragraphs[1].getContent()).toBe(
-        'ZeroDivisionError: integer division or modulo by zero on line 5');
+        'ZeroDivisionError: integer division or modulo by zero');
       expect(paragraphs[2].isTextParagraph()).toBe(true);
       expect(paragraphs[2].getContent()).toBe(
         FeedbackGeneratorService._getUnfamiliarLanguageFeedback(
           LANGUAGE_PYTHON));
+      // Make sure an unknown language does not return any language
+      // unfamiliarity feedback
```
seanlip: nit: add period at end.

client/question/question.js
```diff
@@ -550,103 +550,168 @@ tie.constant(
...
+          'assignments don\'t try to retrieve from indexes below 0 or above ',
+          'the length of the string.'].join('');
+      }
+    }, {
+      // Error when user tries to access a key that isn't defined in a dict
+      friendlyErrorCheck: function(errorString) {
+        return errorString.startsWith('KeyError: ');
+      },
+      getFriendlyErrorText: function(errorString) {
+        var keyErrorRegEx = /KeyError:\s(.*)/;
```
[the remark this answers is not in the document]
rabidbit: I didn't I introduce these variable names. nameErrorRegEx and attributeErrorRegEx also seem to already be in the codebase. Change these to [blah]Regex?

client/question/services/FeedbackGeneratorService.js
rabidbit: I did look at "Files Changed" (I actually did a search on 'seanlip' to make sure that I didn't miss any comments). But thanks for pointing these out, in case some of your comments are not displayed for whatever reason.

client/question/services/FeedbackGeneratorService.js
rabidbit: Will change `WRONG_LANGUAGE_ERRORS.python.forEach(function(error)` to `WRONG_LANGUAGE_ERRORS.python.forEach(function(check)`

client/question/services/FeedbackGeneratorService.js
rabidbit: I think these were flagged by code test coverage for some reason; so, I just deleted them. I can add them back in if you'd like and hope the code test coverage doesn't flag it again (I guess we could approve/push the PR anyway if it came down to it).

client/question/question.js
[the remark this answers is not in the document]
seanlip: Yes please. Thanks!

client/question/services/FeedbackGeneratorService.js
seanlip: Oh, that's odd. They should not be flagged by coverage because there is nothing there to test. But I think having the line breaks is better if that's possible.

client/question/domain/CodeEvalResultObjectFactorySpec.js
```diff
@@ -607,6 +607,18 @@ describe('CodeEvalResultObjectFactory', function() {
     });
   });
 
+  describe('getErrorLineNumber', function() {
+    it('should return correct error line number', function() {
+      var someErrorTraceback = ErrorTracebackObjectFactory.create(
+        'Error on line 5', [TracebackCoordinatesObjectFactory.create(5, 1)]);
```
seanlip: Typically the error message does not have a line number. If that's the case we should not include one in the first argument of the call to ErrorTracebackObjectFactory.create(). 

Ditto below.

rabidbit:
Thanks for the review! I'll see what I can do to improve the test coverage.

client/question/domain/FeedbackDetailsObjectFactory.js
```diff
@@ -30,6 +30,7 @@ tie.factory('FeedbackDetailsObjectFactory', [
      *
      * @param {string} feedbackCategory The category of the feedback. Must be
      *    a valid entry in FEEDBACK_CATEGORIES.
+     * @param {number|null} errorLineNumber The error line number.
```
eyurko:
```suggestion
     * @param {number|null} errorLineNumber The line number of the error.
```

client/question/domain/FeedbackDetailsObjectFactory.js
```diff
@@ -275,6 +289,22 @@ tie.factory('FeedbackDetailsObjectFactory', [
...
+     * This function should return a number representing the line number
+     * where an error in the student code was detected.
+     *
+     * @returns {string}
+     */
+    FeedbackDetails.prototype.getErrorLineNumber = function() {
+      if (
+          this._feedbackCategory !== FEEDBACK_CATEGORIES.SYNTAX_ERROR &&
+          this._feedbackCategory !== FEEDBACK_CATEGORIES.RUNTIME_ERROR) {
+        throw Error('Non-syntax or runtime errors have no error string.');
```
eyurko:
```suggestion
        throw Error('Only syntax and runtime errors have an error string.');
```

eyurko: Also add a test for this case -- it seems like it was added and I don't see a corresponding test.

client/question/question.js
```diff
@@ -550,103 +550,168 @@ tie.constant(
...
+      friendlyErrorCheck: function(errorString) {
+        var re = new RegExp('SyntaxError: invalid syntax');
+        return re.test(errorString);
+      },
+      getFriendlyErrorText: function() {
+        return [
+          'This is the default, generic syntax error message in Python. Here ',
+          'are some common mistakes that can result in this error:',
+          '<ul>',
+          '<li>Mis-matching (missing or too many) braces, brackets, ',
```
eyurko:
```suggestion
          '<li>Mismatching (missing or too many) braces, brackets, ',
```

client/question/services/FeedbackGeneratorServiceSpec.js
```diff
@@ -406,35 +412,68 @@ describe('FeedbackGeneratorService', function() {
...
+      'a general syntax error is generated'].join(''), function() {
+      var errorString = "SyntaxError: invalid syntax";
+      var feedbackString =
+          FeedbackGeneratorService._getFriendlySyntaxFeedback(
+            errorString, LANGUAGE_PYTHON);
+      expect(feedbackString).toEqual([
+        'This is the default, generic syntax error message in Python. Here ',
+        'are some common mistakes that can result in this error:',
+        '<ul>',
+        '<li>Mis-matching (missing or too many) braces, brackets, ',
```
eyurko:
```suggestion
        '<li>Mismatching (missing or too many) braces, brackets, ',
```

client/question/domain/FeedbackDetailsObjectFactory.js
rabidbit: Done

client/question/domain/FeedbackDetailsObjectFactory.js
rabidbit: Changed

rabidbit: Done

client/question/question.js
rabidbit: Done

client/question/services/FeedbackGeneratorServiceSpec.js
rabidbit: Done

rabidbit:
Increased test coverage (except I didn't add test files for the test files ;-) ... which is where the coverage is red).

Also addressed @eyurko comments. Will merge by EOD unless I get more comments from @eyurko.

PR #699, opened 2018-12-11
Un-reverse order in vertical layout

rabidbit:
This PR changes order of the feedback and coding windows in vertical/portrait mode. This is for a few reasons: (a) currently, the pop-up/modal covers the feedback window in horizontal layout and the coding window in vertical layout--this change lets it cover the feedback window in both modes with little to no changes to the pop-up/modal (b) users might not notice the feedback window if they do not scroll down enough--whereas with the proposed layout, it's fine because we can point the users to the coding window anyway and (c) the existing behavior is counter to normal wrapping behavior.

### DIFF
--- a/tests/e2e/question.spec.js
+++ b/tests/e2e/question.spec.js
@@ -90,7 +90,7 @@ describe('Question Page', function() {
        // Coding and Question UI should be vertically aligned.
        expect(codingUiLocation.x).toEqual(questionUiLocation.x);
 
-       // Question UI should be below Coding UI.
+       // Question UI should be above Coding UI.
        expect(codingUiLocation.y).toBeGreaterThan(
          questionUiLocation.y + questionUiSize.height);
      });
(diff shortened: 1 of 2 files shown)


tests/e2e/question.spec.js
```diff
@@ -84,15 +84,15 @@ describe('Question Page', function() {
...
 
        let questionUiLocation = await questionPage.getQuestionUiLocation();
+       let questionUiSize = await questionPage.getQuestionUiSize();
        let codingUiLocation = await questionPage.getCodingUiLocation();
-       let codingUiSize = await questionPage.getCodingUiSize();
 
        // Coding and Question UI should be vertically aligned.
        expect(codingUiLocation.x).toEqual(questionUiLocation.x);
 
        // Question UI should be below Coding UI.
```
eyurko:
```suggestion
       // Question UI should be above Coding UI.
```

rabidbit: Thanks for the catch!

Issue #540, opened 2018-03-16
#532 broke modal dismiss animation


rabidbit:
#532 broke css animation used when dismissing modal (used to slide back up, now it just disappears).

PR #702, opened 2018-12-14
Fix modal dismissal

rabidbit:
The dismiss button on the modal does not dismiss the modal in a smooth manner. Hitting the "Get Feedback" button also dismisses the modal abruptly (it just disappears).

### DIFF
--- a/client/question/components/LearnerViewDirective.js
+++ b/client/question/components/LearnerViewDirective.js
@@ -610,7 +610,8 @@ tie.directive('learnerView', [function() {
       'FEEDBACK_CATEGORIES', 'DEFAULT_EVENT_BATCH_PERIOD_SECONDS',
       'DELAY_STYLE_CHANGES', 'THEME_NAME_LIGHT', 'THEME_NAME_DARK',
       'CODE_RESET_CONFIRMATION_MESSAGE', 'PRIVACY_URL', 'ABOUT_TIE_URL',
-      'ABOUT_TIE_LABEL', 'TERMS_OF_USE_URL',
+      'ABOUT_TIE_LABEL', 'TERMS_OF_USE_URL', 'FEEDBACK_MODAL_HEIGHT_OFFSET',
+      'FEEDBACK_MODAL_HIDE_HEIGHT_OFFSET',
       function(
           $scope, $interval, $timeout, $location, $window,
           ConversationManagerService, QuestionDataService, LANGUAGE_PYTHON,
@@ -626,7 +627,8 @@ tie.directive('learnerView', [function() {
           FEEDBACK_CATEGORIES, DEFAULT_EVENT_BATCH_PERIOD_SECONDS,
           DELAY_STYLE_CHANGES, THEME_NAME_LIGHT, THEME_NAME_DARK,
           CODE_RESET_CONFIRMATION_MESSAGE, PRIVACY_URL, ABOUT_TIE_URL,
-          ABOUT_TIE_LABEL, TERMS_OF_USE_URL) {
+          ABOUT_TIE_LABEL, TERMS_OF_USE_URL, FEEDBACK_MODAL_HEIGHT_OFFSET,
+          FEEDBACK_MODAL_HIDE_HEIGHT_OFFSET) {
         $scope.PRIVACY_URL = PRIVACY_URL;
         $scope.ABOUT_TIE_URL = ABOUT_TIE_URL;
         $scope.ABOUT_TIE_LABEL = ABOUT_TIE_LABEL;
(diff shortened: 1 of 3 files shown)


rabidbit:
I added seanlip for the overall review and davyrisso-at-google for the E2E test review.

I used browser.sleep() because I could not get browser.wait() to work properly. It works as is, but I suspect using browser.wait() might be more robust? Love to get suggestions on how to make the tests better.

davyrisso-at-google:
> I used browser.sleep() because I could not get browser.wait() to work properly. It works as is, but I suspect using browser.wait() might be more robust? Love to get suggestions on how to make the tests better.

I recommend using await browser.wait() with ExpectedConditions.presenceOf()/visibilityOf()/invisibilityOf(). 
(See https://www.protractortest.org/#/api?view=ProtractorExpectedConditions for more details)

E.g in pageObjects openModal() would become:

this.openModal = async function() {
    // We wait for the modal link to be present.
    await browser.wait(
      ExpectedConditions.presenceOf(modalLink),
      MODAL_LINK_PRESENCE_TIMEOUT);
    await modalLink.click();
    // We wait for the modal to be visible.
    await browser.wait(
      ExpectedConditions.invisibilityOf(this.monospaceModalContainerElement),
      MODAL_DISMISSAL_TIMEOUT);
  };

and dismissModal(): 

this.dismissModal = async function() {
    // We wait for the modal dismissal link to be visible.
    await browser.wait(
      ExpectedConditions.visibilityOf(dismissModalButton),
      MODAL_PRESENCE_TIMEOUT);
    await dismissModalButton.click();
    // We wait for the modal to be invisible.
    await browser.wait(
      ExpectedConditions.invisibilityOf(this.monospaceModalContainerElement),
      MODAL_DISMISSAL_TIMEOUT);
  };

Note that in the spec it is necessary to use await when calling both these methods, i.e:
await questionPage.openModal();
// Test...
await questionPage.dismissModal();

I am not sure if the current tests in specs are all necessary (e.g testing the presence/absence of the ng-hide class should be taken care of with ExpectedCondition.visibilityOf/invisibilityOf, maybe we should rather test the presence and content of the text in the modal?

Let me know if this works!

client/question/components/LearnerViewDirective.js
```diff
@@ -1180,13 +1182,46 @@ tie.directive('learnerView', [function() {
             tasks[currentTaskIndex].getId());
         };
 
+        /**
+         * Event handler to hide modal after transition animation ends.
+         */
+        $scope.hideModalAfterAnimation = function(event) {
```
seanlip: I'm not sure this needs to be defined here; it duplicates code in MonospaceDisplayModalService.

client/question/components/LearnerViewDirective.js
```diff
@@ -1180,13 +1182,46 @@ tie.directive('learnerView', [function() {
...
+          $timeout(function() {
+            MonospaceDisplayModalService.hideModal();
+          }, 0);
+          event.target.removeEventListener("transitionend",
+              $scope.hideModalAfterAnimation, false);
+        };
+
+        /**
+         *  Close modal by calling MonospaceDisplayModalService.closeModal.
```
seanlip: The function doesn't seem to be doing what this docstring says.

client/question/components/LearnerViewDirective.js
```diff
@@ -1180,13 +1182,46 @@ tie.directive('learnerView', [function() {
...
+         */
+        $scope.closeModal = function() {
```
seanlip: The code here duplicates code in MonospaceDisplayModalDirective. In general try to strictly avoid duplicating code since that makes things harder to maintain and introduces the possibility of skew.

client/question/components/MonospaceDisplayModalDirective.js
```diff
@@ -171,26 +171,34 @@ tie.directive('monospaceDisplayModal', [function() {
         });
 
         /**
-         * Close the modal.
+         * Event handler to hide modal after transition animation ends.
+         */
+        $scope.hideModalAfterAnimation = function(event) {
```
seanlip: Consider using `evt` instead of `event` throughout (which I think we do already).

(`event` is a global variable in some versions of IE, so typically I avoid using it.)

client/question/components/MonospaceDisplayModalDirective.js
```diff
@@ -171,26 +171,34 @@ tie.directive('monospaceDisplayModal', [function() {
...
+          $timeout(function() {
+            MonospaceDisplayModalService.hideModal();
+          }, 0);
+          event.target.removeEventListener("transitionend",
```
seanlip: nit: break after '(' if contents of (...) span more than one line

Here and elsewhere, use single quotes

rabidbit:
@seanlip Re: UI/UX, I think it looks janky when it slides down on display and disappears on dismiss. Also, we have https://github.com/google/tie/issues/540, which this fixes.

seanlip:
OK. I'm happy to defer on UX.

